PR #1, opened 2019-07-19
DON-15 simple campaign page

tbg-jasonfung:
- DON-36 Create a Service that can load a campaign's detail by ID
- DON-37 Build a page that renders data from the Service

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -8,12 +8,6 @@
         <a href="https://www.google.com/chrome/" target="_blank">Chrome</a>,
         but any of these will work with the Big Give donations and keep you safer everywhere on the internet.</p>
     </div>
-
-    <!-- Temporary hard-coded link to donate to a Full sandbox campaign -->
-    <p><a routerLink="/donate/a051X000000sbmuQAA">&#62; Try a donation!</a></p>
-
-    <!-- Temporary hard-coded link to view a campaign on Full sandbox campaign -->
-    <p><a routerLink="/campaign/a051r00001EywjpAAB">&#62; View a campaign!</a></p>
   </div>
 
   <router-outlet *ngIf="browserSupported"></router-outlet>
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,7 +1,6 @@
 <div>
   <h1>
-    <img src="/assets/logo.png" alt="The Big Give" width="200">
-    Donations
+    <a href="/"><img src="/assets/logo.png" alt="The Big Give" width="200"></a>
   </h1>
 
   <div *ngIf="!browserSupported" class="error">
@@ -14,6 +13,9 @@
 
   <!-- Temporary hard-coded link to donate to a Full sandbox campaign -->
   <p><a routerLink="/donate/a051X000000sbmuQAA">Try a donation!</a></p>
+
+  <!-- Temporary hard-coded link to view a campaign on Full sandbox campaign -->
+  <p><a routerLink="/campaign/a051r00001EywjpAAB">View a campaign!</a></p>
 </div>
 
 <router-outlet *ngIf="browserSupported"></router-outlet>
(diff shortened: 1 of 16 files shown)


src/app/campaign.model.ts
```diff
@@ -0,0 +1,17 @@
+export interface Campaign {
```
NoelLH: Interfaces are used only at TypeScript compile time. It seems like Angular encourages classes so it's probably best to use these for all models for consistency https://stackoverflow.com/questions/37652801/when-to-use-interface-and-model-in-typescript-angular2

I think you will also find when writing unit tests for the new parts that you can't make the objects you need without implementing or replacing the interface

src/app/campaign.service.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+          description: 'x',
+          amount: 2000.00,
+        },
+      ],
+      bannerUri: 'https://thebiggive--Full--c.cs101.content.force.com/sfc/dist/version/download/?oid=00D1X0000008ahG&ids=0681X0000007w5l&d=%2Fa%2F1X0000008UpU%2FvdeP6lXwciVq6CMSA8BhwNQ6hNTyZHu7VmcugZgN2SY&asPdf=false',
+      amountRaised: 200.00,
+      additionalImageUris: [],
+    }];
+
+    service.getCampaign('a051X000000tMVZQA2').subscribe(campaign => {
```
NoelLH: How do we know these assertions actually ran? Observables in tests should have the error case caught and fail the test, if they're expected to succeed. See `donation.service.spec.ts` for one approach to ensuring this.

src/app/campaign.service.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    imports: [ HttpClientTestingModule ],
+    providers: [ CampaignService ],
+  }));
+
+  it('should retrieve campaign details from API', () => {
+    const service: CampaignService = TestBed.get(CampaignService);
+    const httpMock: HttpTestingController = TestBed.get(HttpTestingController);
+
+    const dummyCampaign: Campaign[] = [{
+      video: [],
```
NoelLH: For the service test, I think it would make sense to populate all these empty properties even if the app doesn't show them yet, so we can see that serialisation is working in the service and it's ready to use with those properties.

src/app/campaign.service.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { Campaign } from './campaign.model';
+import { environment } from '../environments/environment';
+
+
+@Injectable({
+  providedIn: 'root',
+})
+export class CampaignService {
+  private apiPath = '/campaigns/services/apexrest/v1.0/campaigns/';
```
NoelLH: I would move the `/` out of this so we can use the same var across the service

src/app/campaign-details/campaign-details.component.ts
```diff
@@ -0,0 +1,27 @@
+import { Component, OnInit } from '@angular/core';
+import { CampaignService } from '../campaign.service';
+import { ActivatedRoute, Params } from '@angular/router';
```
NoelLH: `Params` is not being used

src/app/campaign-details/campaign-details.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="container">
+    <img class="banner" src={{campaign.bannerUri}} alt="Campaign Banner">
```
NoelLH: I think binding `src` using the expression context is a bit more readable in this case https://angular.io/guide/template-syntax#expression-context

src/app/app.component.html
```diff
@@ -1,9 +1,5 @@
-<div>
-  <h1>
-    <img src="/assets/logo.png" alt="The Big Give" width="200">
-    Donations
-  </h1>
-
+<div class="header">
+  <h1><a href="/"><img src="/assets/logo.png" alt="The Big Give" width="200"></a>2.1</h1>
```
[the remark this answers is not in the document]
tbg-jasonfung: @NoelLH No reason, thought it just looked better, i'll swap this back to 'donations'

tbg-jasonfung:
Please see https://github.com/thebiggive/donate-frontend/pull/4

PR #5, opened 2019-07-24
DON-15 - fix type errors blocking demo build + improve tests

NoelLH:
@tbg-jasonfung I missed some of the type issues in review, sorry. Please keep an eye out for what should and should be an array (multiple) type.

You can also see on the CLI server if there are runtime type errors like the ones that blocked the previous S3 demo build from transpiling. I don't think these will be very common but if this happens again we can look into how we can have these caught during the CI `build` job too so this type of error doesn't get into `master`.

* Fix several places incorrectly looking for an _array of_ charities or an _array of_ campaigns when retrieving a single campaign or linked charity from a campaign (many-to-one relation)
* Update `CampaignService` spec's dummy data so it passes now that types are correct
* Add some actual test data and some more substantial assertions to `CampaignDetailsComponent` spec so it can be rendered in the test now that types are correct

### DIFF
--- a/src/app/campaign-details/campaign-details.component.html
+++ b/src/app/campaign-details/campaign-details.component.html
@@ -1,7 +1,7 @@
 <img class="banner" [src]="campaign.bannerUri" alt="Campaign Banner">
 <div class="details">
 	<div>
-		<h2>{{campaign.title}} <em *ngIf="campaign.championName"> | Championed by: {{campaign.championName}}</em></h2>
+		<h2>{{campaign.title}}<em *ngIf="campaign.championName"> | Championed by: {{campaign.championName}}</em></h2>
 		<h3 *ngIf="campaign.isMatched">Matched Campaign!</h3>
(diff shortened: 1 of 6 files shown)


tbg-jasonfung:
> @tbg-jasonfung I missed some of the type issues in review, sorry. Please keep an eye out for what should and should be an array (multiple) type.
> 
> You can also see on the CLI server if there are runtime type errors like the ones that blocked the previous S3 demo build from transpiling. I don't think these will be very common but if this happens again we can look into how we can have these caught during the CI `build` job too so this type of error doesn't get into `master`.
> 
> * Fix several places incorrectly looking for an _array of_ charities or an _array of_ campaigns when retrieving a single campaign or linked charity from a campaign (many-to-one relation)
> * Update `CampaignService` spec's dummy data so it passes now that types are correct
> * Add some actual test data and some more substantial assertions to `CampaignDetailsComponent` spec so it can be rendered in the test now that types are correct

@NoelLH No problem thanks, I should have followed the existing examples better - cheers for sorting!

PR #10, opened 2019-08-28
DON-27, DON-28 - add meta-campaign subtitles

NoelLH:
Show time left to the meta-campaign's start date if in Preview, or its total amount raised otherwise.

### DIFF
--- a/src/app/meta-campaign/meta-campaign.component.html
+++ b/src/app/meta-campaign/meta-campaign.component.html
@@ -2,10 +2,12 @@
   <mat-card-header>
     <div mat-card-avatar class="childhood-trust-icon"></div>
     <mat-card-title>{{ campaign.title }}</mat-card-title>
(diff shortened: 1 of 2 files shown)


src/app/meta-campaign/meta-campaign.component.html
```diff
@@ -2,10 +2,12 @@
   <mat-card-header>
     <div mat-card-avatar class="childhood-trust-icon"></div>
     <mat-card-title>{{ campaign.title }}</mat-card-title>
+    <mat-card-subtitle *ngIf="campaign.status == 'Preview'">Opens in {{ campaign.startDate | timeLeft }}</mat-card-subtitle>
```
tbg-jasonfung: @NoelLH This looks good. Can we add logic to say `IF campaign.startDate == Today` then show `Opens today!` as this looks a bit weird:

image

NoelLH: @tbg-jasonfung as noted on the Jira task I think we should leave refinements + exact presentation of any countdown until we have both more time and some UX input

PR #12, opened 2019-09-09
DON-38 include new alternative fund use field in campaign details view

tbg-jasonfung:
- DON-38 include new alternative fund use field in campaign details view

### DIFF
--- a/src/app/campaign-details/campaign-details.component.html
+++ b/src/app/campaign-details/campaign-details.component.html
@@ -3,33 +3,34 @@
 <h2>Banner Image here!</h2>
 
 <div class="details" *ngIf="campaign">
-	<div>
-		<h2>{{campaign.title}}</h2>
+  <div>
+    <h2>{{campaign.title}}</h2>
     <h3 *ngIf="campaign.championName">Championed by {{campaign.championName}}</h3>
-		<h4 *ngIf="campaign.isMatched">Matched Campaign!</h4>
-		<h4 *ngIf="!campaign.isMatched">Regular campaign</h4>
-	</div>
+    <h4 *ngIf="campaign.isMatched">Matched Campaign!</h4>
+    <h4 *ngIf="!campaign.isMatched">Regular campaign</h4>
+  </div>
 
-	<div>
-		<h3>Campaign Summary</h3>
+  <div>
+    <h3>Campaign Summary</h3>
     <p>{{campaign.summary}}</p>
     <p>Status: <strong>{{ campaign.status }}</strong></p>
     <p *ngIf="campaign.status == 'Active'">{{ campaign.endDate | timeLeft }} left</p>
-	</div>
+  </div>
 
-	<div>
-		<h3>Campaign Targets</h3>
-		<p *ngIf="campaign.target > 0">Campaign Target: {{ campaign.target | currency : '£' : 'symbol' : '1.0-0' }}</p>
-    <p *ngIf="campaign.amountRaised > 0">Amount Raised: {{ campaign.amountRaised | currency : '£' : 'symbol' : '1.0-0' }}</p>
+  <div>
+    <h3>Campaign Targets</h3>
+    <p *ngIf="campaign.target > 0">Campaign Target: {{ campaign.target | currency : '£' : 'symbol' : '1.0-0' }}</p>
+    <p *ngIf="campaign.amountRaised > 0">Amount Raised:
+      {{ campaign.amountRaised | currency : '£' : 'symbol' : '1.0-0' }}</p>
     <p>Donations: {{ campaign.donationCount }}</p>
-	</div>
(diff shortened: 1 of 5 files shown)


src/app/campaign-details/campaign-details.component.html
```diff
@@ -5,23 +5,28 @@ <h2>Banner Image here!</h2>
...
 	<div>
 		<h3>Campaign Targets</h3>
 		<p *ngIf="campaign.target > 0">Campaign Target: {{ campaign.target | currency : '£' : 'symbol' : '1.0-0' }}</p>
-    <p *ngIf="campaign.amountRaised > 0">Amount Raised: {{ campaign.amountRaised | currency : '£' : 'symbol' : '1.0-0' }}</p>
-    <p>Donations: {{ campaign.donationCount }}</p>
+		<p *ngIf="campaign.amountRaised > 0">Amount Raised: {{ campaign.amountRaised | currency : '£' : 'symbol' : '1.0-0' }}</p>
+		<p>Donations: {{ campaign.donationCount }}</p>
+	</div>
+
+	<div *ngIf="campaign.alternativeFundUse != null">
```
NoelLH: @tbg-jasonfung with the current proposed model change this is never actually `null` (the type is not nullable), so would probably make more sense to test for either `campaign.alternativeFundUse` (is it 'truth-y' in general?) or `campaign.alternativeFundUse.length > 0`

src/app/campaign-details/campaign-details.component.html
```diff
@@ -5,23 +5,28 @@ <h2>Banner Image here!</h2>
 <div class="details" *ngIf="campaign">
 	<div>
 		<h2>{{campaign.title}}</h2>
-    <h3 *ngIf="campaign.championName">Championed by {{campaign.championName}}</h3>
+    	<h3 *ngIf="campaign.championName">Championed by {{campaign.championName}}</h3>
```
NoelLH: This indent looks weird - we should avoid literal tab characters as it makes this inconsistent across editors.

This seems to have changed for whitespace in general. Please make sure your editor is configured to respect the project's `.editorconfig` for HTML too @tbg-jasonfung

src/app/campaign-details/campaign-details.component.html
tbg-jasonfung: @NoelLH Interestingly, I swapped to use `.length` check but its throwing this error:

image

NoelLH: @tbg-jasonfung in that case the API and the client model are not consistent. We should probably make the property nullable in the model (similar to optional fields on `Donation`) if the API is returning a null value or omitting the property.

The truth-i-ness check option will work either way I think, but we should make our types accurate in order to benefit from the checks built into TypeScript.

src/app/meta-campaign/meta-campaign.component.spec.ts
```diff
@@ -55,6 +55,7 @@ describe('MetaCampaignComponent', () => {
       'Test Master Campaign!',
       [],
       [],
+      '',
```
NoelLH: @tbg-jasonfung surely the optional end param should be omitted if we're trying to simulate what happens for meta-campaigns where it's not set? This is why the param is now optional!

src/app/campaign-details/campaign-details.component.html
```diff
@@ -5,23 +5,28 @@ <h2>Banner Image here!</h2>
 <div class="details" *ngIf="campaign">
 	<div>
 		<h2>{{campaign.title}}</h2>
-    <h3 *ngIf="campaign.championName">Championed by {{campaign.championName}}</h3>
+		<h3 *ngIf="campaign.championName">Championed by {{campaign.championName}}</h3>
```
NoelLH: @tbg-jasonfung this is still changing various spaces to tabs, please check your editor config and the diff for changes already applied

src/app/meta-campaign/meta-campaign.component.spec.ts
tbg-jasonfung: @NoelLH Yup you're right, I went a bit update trigger happy

src/app/campaign-details/campaign-details.component.html
```diff
@@ -23,6 +23,11 @@ <h3>Campaign Targets</h3>
     <p *ngIf="campaign.amountRaised > 0">Amount Raised: {{ campaign.amountRaised | currency : '£' : 'symbol' : '1.0-0' }}</p>
     <p>Donations: {{ campaign.donationCount }}</p>
 	</div>
+
```
NoelLH: @tbg-jasonfung  this was still using tabs, I guess because the editorconfig was not respected when the file was first made and VS Code follows the file's existing convention. I'll update and merge.

PR #13, opened 2019-09-16
DON-19 + DON-20 show quotes and budget details

tbg-jasonfung:
- DON-19 + DON-20 show quotes and budget details

### DIFF
--- a/src/app/campaign-details/campaign-details.component.html
+++ b/src/app/campaign-details/campaign-details.component.html
@@ -25,20 +25,20 @@
     <p>Donations: {{ campaign.donationCount }}</p>
   </div>
 
-  <div *ngIf="campaign.budgetDetails.length != 0">
-      <h3>Budget Details</h3>
-      <div *ngFor="let detail of campaign.budgetDetails">
-        <p>Amount: {{ detail.amount | currency : '£' : 'symbol' : '1.0-0' }}</p>
-        <p>Description: {{ detail.description }}</p>
-      </div>
+  <div *ngIf="campaign.budgetDetails.length > 0">
+    <h3>Budget Details</h3>
+    <ul *ngFor="let detail of campaign.budgetDetails">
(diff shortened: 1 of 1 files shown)


src/app/campaign-details/campaign-details.component.html
```diff
@@ -25,6 +25,22 @@ <h3>Campaign Targets</h3>
     <p>Donations: {{ campaign.donationCount }}</p>
   </div>
 
+  <div *ngIf="campaign.budgetDetails.length != 0">
```
NoelLH: @tbg-jasonfung I think these would be a little more readable and less likely to ever have weirdness with type coercion if they checked specifically for `> 0`

src/app/campaign-details/campaign-details.component.html
```diff
@@ -25,6 +25,22 @@ <h3>Campaign Targets</h3>
...
+      <h3>Budget Details</h3>
+      <div *ngFor="let detail of campaign.budgetDetails">
```
NoelLH: @tbg-jasonfung for both things iterated over in this PR, they are semantically a list of similar un-ordered elements. Why don't we use the HTML list elements so the semantics are more accurate for all devices / bots and more likely to work with screen readers etc.?

src/app/campaign-details/campaign-details.component.html
```diff
@@ -25,6 +25,22 @@ <h3>Campaign Targets</h3>
...
 
+  <div *ngIf="campaign.budgetDetails.length > 0">
+    <h3>Budget Details</h3>
+    <ul *ngFor="let detail of campaign.budgetDetails">
+      <li>Amount: {{ detail.amount | currency : '£' : 'symbol' : '1.0-0' }}</li>
+      <li>Description: {{ detail.description }}</li>
+    </ul>
+  </div>
+  
+  <div *ngIf="campaign.quotes.length != 0">
```
NoelLH: @tbg-jasonfung  same as above here

PR #14, opened 2019-09-17
DON-17-DON-26 Display campaign additional images on carousel

tbg-jasonfung:
- DON-17-DON-26 Display campaign additional images on carousel


tbg-jasonfung:
Example campaign with images uploaded to aws bucket: http://localhost:4200/campaign/a051q000001svQ5AAI

PR #15, opened 2019-09-18
DON-56 - add Docker support + related

NoelLH:
* DON-56 - add a production-ready `Dockerfile`
* DON-53 - work around a font issue with Server-Side Rendering
* DON-34 - warn Internet Explorer 9 & 10 users and exclude those versions from build support
* Add a custom CircleCI-based Docker container which CI can use in future for faster builds


tbg-jasonfung:
Looks good to me, also tested in IE9+10 after building prod copy, working as expected

PR #44, opened 2019-10-22
DON-79 client side donate retry

tbg-jasonfung:


tbg-jasonfung:
@NoelLH Hey, this is what I've got so far, can you point me in the right direction on how to use the EventEmitter correctly?

package.json
```diff
@@ -41,7 +41,9 @@
     "helmet": "^3.21.1",
     "morgan": "^1.9.1",
     "ngx-webstorage-service": "^4.1.0",
-    "rxjs": "~6.5.0",
+    "puppeteer-core": "^1.20.0",
+    "rxjs": "^6.5.3",
+    "rxjs-compat": "^6.5.3",
```
NoelLH: @tbg-jasonfung We shouldn't be adding in `rxjs-compat` to a new project - this lib is for compatibility with the old API, so the right approach is to find the new versions of the operators.

package.json
```diff
@@ -41,7 +41,8 @@
     "helmet": "^3.21.1",
     "morgan": "^1.9.1",
     "ngx-webstorage-service": "^4.1.0",
-    "rxjs": "~6.5.0",
+    "puppeteer-core": "^1.20.0",
```
NoelLH: @tbg-jasonfung  I don't think we need to add this? Other branch was passing without it.

tbg-jasonfung: @NoelLH Went a bit import trigger happy, have removed now

PR #55, opened 2019-11-04
More realistic meta-campaign search

NoelLH:
* Add general purpose in-metacampaign search
* Remove SG19-specific search demo toggle

For now, this keeps global search logic using linkable URIs but
keeps search within a meta-campaign stateful. This will mean
anybody copy/pasting from meta-campaign page would share the
base (e.g.) CC19 URI rather than a filtered view

### DIFF
--- a/src/app/meta-campaign/meta-campaign.component.ts
+++ b/src/app/meta-campaign/meta-campaign.component.ts
@@ -75,10 +75,6 @@ export class MetaCampaignComponent implements OnInit {
    */
   setQueryProperty(property, event) {
     this.query[property] = event.value;
-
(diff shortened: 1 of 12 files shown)


src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -75,6 +75,10 @@ export class MetaCampaignComponent implements OnInit {
    */
   setQueryProperty(property, event) {
     this.query[property] = event.value;
+
+    console.log('new query:');
+    console.log(this.query);
```
tbg-jasonfung: @NoelLH Are these intentional?

PR #122, opened 2019-11-22
DON-141 - add ticker

james-daughter:


src/app/hero/hero.component.ts
```diff
@@ -28,6 +28,15 @@ export class HeroComponent implements OnInit {
     if (isPlatformBrowser(this.platformId)) {
       Modernizr.on('webp', browserSupportsWebp => this.webp = browserSupportsWebp);
     }
+
+    this.cloneTicker();
+  }
+
+  cloneTicker() {
+    var ticker = document.querySelector('.c-ticker');
```
NoelLH: As discussed on Slack, this is probably not server rendering-proof. Would be good to get Angular to build 2 straight up.

PR #125, opened 2019-11-22
Align 1- and 2-line titles when showing 2+ cols

NoelLH:


NoelLH:
Discussed on Slack and have applied @james-daughter's suggested media constraint, so will merge this straight in.

PR #127, opened 2019-11-24
DON-119 - transfer server state to the client

NoelLH:
On the three main pages which can benefit from this, finish following the data handover approach detailed in https://medium.com/@evertonrobertoauler/angular-5-universal-with-transfer-state-using-angular-cli-19fe1e1d352c

This makes page loading smoother and eliminates unnecessary extra client-side HTTP calls

### DIFF
--- a/src/app/meta-campaign/meta-campaign.component.ts
+++ b/src/app/meta-campaign/meta-campaign.component.ts
@@ -1,4 +1,5 @@
 import { Component, Inject, OnInit, PLATFORM_ID } from '@angular/core';
+import { makeStateKey, TransferState } from '@angular/platform-browser';
 import { ActivatedRoute } from '@angular/router';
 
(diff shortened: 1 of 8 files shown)


src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -8,6 +9,9 @@ import { Fund } from '../fund.model';
 import { FundService } from '../fund.service';
 import { PageMetaService } from '../page-meta.service';
 
+const FUND_KEY = makeStateKey('fund');
+const METACAMPAIGN_KEY = makeStateKey('metacampaign');
```
tbg-jasonfung: This is new to me and pretty useful!

tbg-jasonfung:
@NoelLH Conflict to resolve, otherwise thumbs up from me

PR #128, opened 2019-11-25
Minor UX tweaks

NoelLH:
Quick changes discussed with Jonathan this morning:

* Meta-campaign: Tweak filter copy & move UK to the top
* Campaign detail: Move budget detail to same line as amount
* Donation start: Take out campaign summary so donors reach the form earlier


NoelLH:
These are very low risk changes and I don't think really need individual review or QA so merging to `develop` for now

PR #130, opened 2019-11-25
DON-119 - fix state confusion

NoelLH:


src/app/campaign-details/campaign-details.component.ts
```diff
@@ -32,13 +30,13 @@ export class CampaignDetailsComponent implements OnInit {
   }
 
   ngOnInit() {
-    this.campaign = this.state.get(CAMPAIGN_KEY, undefined);
```
tbg-jasonfung: @NoelLH  could we not have `campaign-${this.campaignId}` as a shared var? Similarly for `fund-for-${this.campaignId}` ?

NoelLH: Done and added typing to keys @tbg-jasonfung

PR #135, opened 2019-11-26
DON-142 - add polyfills in hopes of fixing IE11

NoelLH:
The basics worked with fewer Material components in play, so hopefully the animation polyfill will get it back to being usable

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11580,6 +11580,11 @@
         "minimalistic-assert": "^1.0.0"
       }
     },
+    "web-animations-js": {
+      "version": "2.3.2",
+      "resolved": "https://registry.npmjs.org/web-animations-js/-/web-animations-js-2.3.2.tgz",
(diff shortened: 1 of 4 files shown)


tbg-jasonfung:
@NoelLH It's not loading anything now on IE11

image

The equivalent but on staging before this polyfill change:

image

NoelLH:
@tbg-jasonfung 
* what do the logs say?
* did you `npm i`?

tbg-jasonfung:
@NoelLH Yep, have the latest dependencies, you mean IE console log?

image

not alot

NoelLH:
@tbg-jasonfung what about the rendered markup on inspect, loaded HTTP requests, ...?

tbg-jasonfung:
Markup:
image

Network:
image

NoelLH:
@tbg-jasonfung the network panel looks odd. If everything is coming from the browser cache, how do we know it's what the current code is doing?

And if on a full load it still fails, could you try with just the new animations library please? Even if it works it would be good to check whether we need any of the `/es/` polyfills or not so we're not bloating the polyfill bundle unnecessarily.

tbg-jasonfung:
Hmm @NoelLH I think theres something odd with my IE11, even changing to `develop` branch the page isn't loading - i'm going to merge this in and have Barbora double check on her machine - we can alter this if needed.

PR #169, opened 2019-11-29
DON-153 Include charity regulator + number if available

tbg-jasonfung:
- DON-153 Include charity regulator + number if available


NoelLH:
@tbg-jasonfung I think we should update the data model so it's clear from code what's expected to be received / deserialised from API responses

src/app/campaign.model.ts
```diff
@@ -7,7 +7,16 @@ export class Campaign {
     public bannerUri: string,
     public budgetDetails: Array<{amount: number, description: string}>,
     public championName: string,
-    public charity: {id: string, name: string},
+    public charity: {id: string,
+                      name: string,
+                      donateLinkId: string,
+                      facebook?: string,
```
NoelLH: Good call, not sure how I missed these! :s

PR #173, opened 2019-12-02
DON-167 - fix IE11 not accepting donation amounts

NoelLH:


NoelLH:
From debugging locally, it looks like this is a discrepancy/bug with how IE handles the contents of the [expansion panel](https://material.angular.io/components/expansion/overview) containing the donate amount box when it has disabled=”true”. It seems to kill all user input inside the same box when that attribute is set.

Removing the attribute means the user can collapse the donation amount box, in any browser. This probably isn’t ideal but I think we should go with it for now as it’s the smallest change that fixes the bug without causing another major regression. It is at least consistent with the other collapsing boxes on the page and there is an arrow to indicate you can reopen it. It is possible to hide the close/open arrow without disabling the box, but you can still tap the title to collapse the box, and I think this will be much more confusing if we keep the functionality but don’t have an arrow to indicate it can be re-expanded.

Longer term it is probably better for the containing element to not be an expansion panel at all but just have a similar border without being collapsible.

PR #178, opened 2019-12-02
[DO NOT MERGE] Production config change: disable all suggested amounts

NoelLH:
This PR is ready as a contingency plan if we find from Analytics data that the new suggested donation amounts feature is having an adverse effect on donations.

We should merge it only after allowing enough time and analysing data to a point where we are convinced this is the case.


NoelLH:
I'm going to close this for now as in the last day it's very unlikely we'll choose to change / end the current multivariant experiment. I've been looking at the GA events occasionally and there certainly isn't a clear pattern that all suggestions pull amounts down, though the performance of each set has been changing across the campaign. Right now the highest suggestion set has a higher average than no suggestions and the other 3 sets are lower.

PR #196, opened 2019-12-08
Add a line to indicate what's going on outside campaign dates

NoelLH:
This should only be seen if a campaign's status and dates do not agree, as we redirect Donate to the Campaign page when it's expected to be closed based on the status


tbg-jasonfung:
@NoelLH I think we had some dynamic copy here which switched to "the campaign was open during... thank you for your support" or words to that effect?

NoelLH:
@tbg-jasonfung when statuses are self-consistent this page isn't supposed to be visible so any messaging would belong on the campaign details view, not here.

PR #197, opened 2019-12-08
DON-178 - improve donation amount validation

NoelLH:
* Accept '.00' after main amount
* Show a validation message next to the Donate button if it's greyed out


tbg-jasonfung:
Holding off merge until we confirm in stand up this morning

tbg-jasonfung:
@NoelLH small conflict on this branch - otherwise merge when you've got confirmation from stand up

PR #213, opened 2020-01-21
DON-117 dynamic hero image

tbg-jasonfung:
- DON-117 dynamic hero image


src/app/hero/hero.component.html
```diff
@@ -1,4 +1,4 @@
-<div [class]="'hero ' + (webp ? 'webp-bg' : 'nowebp-bg')">
+<div [class]="'hero ' + (webp ? 'webp-bg' : 'nowebp-bg')" [ngStyle]="{'background-image': 'url(' + campaign.bannerUri + ')'}">
```
NoelLH: @tbg-jasonfung we should stop setting the CSS classes now they don't do anything.

For now I think we can keep the logic which sets up the `webp` var in TypeScript as I'm hoping we can bring in requests that get automatic webp conversion via our Lambda / API Gateway once the image access update work is done, so we will hopefully use the same var in a slightly different way in a few days.

src/app/hero/hero.component.html
```diff
@@ -1,4 +1,4 @@
-<div [class]="'hero ' + (webp ? 'webp-bg' : 'nowebp-bg')">
+<div [ngStyle]="{'background-image': 'url(' + campaign.bannerUri + ')'}">
```
NoelLH: @tbg-jasonfung  don't we still need `hero` as a fixed/static class?

tbg-jasonfung: Ahh oops!

PR #214, opened 2020-01-21
DON-118 Include new surplus info copy in popup modal

tbg-jasonfung:
- DON-118 Include new surplus info copy in popup modal

### DIFF
--- a/src/app/campaign.model.ts
+++ b/src/app/campaign.model.ts
@@ -38,7 +38,7 @@ export class Campaign {
     public campaignCount?: number,
     public championRef?: string,
     public parentRef?: string,
-    public video?: {provider: string, key: string},
     public surplusDonationInfo?: string,
+    public video?: {provider: string, key: string},
   ) {}
 }
--- a/src/app/campaign.model.ts
+++ b/src/app/campaign.model.ts
@@ -39,5 +39,6 @@ export class Campaign {
     public championRef?: string,
     public parentRef?: string,
     public video?: {provider: string, key: string},
+    public surplusDonationInfo?: string,
   ) {}
 }
--- a/src/app/donation-start/donation-start-match-confirm-dialog.html
+++ b/src/app/donation-start/donation-start-match-confirm-dialog.html
@@ -6,9 +6,9 @@
   <p>{{ data.statusDetail }}</p>
 
   <p>Donations received once the charity has hit its target may be used for the charity's general funds.</p>
-  <!-- Temporarily fixed to CC19 copy. This will be pulled from the API soon. -->
-  <p>If the charity has hit their target and is participating in the Christmas Challenge, unmatched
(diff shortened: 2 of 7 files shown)


src/app/campaign.model.ts
```diff
@@ -39,5 +39,6 @@ export class Campaign {
     public championRef?: string,
     public parentRef?: string,
     public video?: {provider: string, key: string},
+    public surplusDonationInfo?: string,
```
NoelLH: Better to keep optional params alphabetical for readability / maintainability. Can't remember if this one has been reworked yet but we should try to avoid direct constructors that depend on the param order so new param additions anywhere besides the end aren't too painful to make going forward.

src/app/donation-start/donation-start-match-confirm-dialog.html
```diff
@@ -6,9 +6,9 @@ <h2 mat-dialog-title>{{ data.title }}</h2>
   <p>{{ data.statusDetail }}</p>
 
   <p>Donations received once the charity has hit its target may be used for the charity's general funds.</p>
-  <!-- Temporarily fixed to CC19 copy. This will be pulled from the API soon. -->
-  <p>If the charity has hit their target and is participating in the Christmas Challenge, unmatched
-    donations can help them to win a Christmas Challenge Award.</p>
+
+  <p *ngIf="data.surplusDonationInfo">{{ data.surplusDonationInfo }}</p>
```
NoelLH: I've changed 'prepended' to 'appended' in the API field description to reflect what is actually happening here. Based on the overall copy we landed on for CC19 I think this way round makes sense.

src/app/donation-start/donation-start.component.ts
```diff
@@ -337,21 +337,22 @@ export class DonationStartComponent implements OnInit {
     this.charityCheckoutService.startDonation(donation);
   }
 
-  private promptToContinueWithNoMatchingLeft(donation: Donation) {
+  private promptToContinueWithNoMatchingLeft(donation: Donation, surplusDonationInfo?: string) {
```
NoelLH: This method is already non-static and the campaign never changes. This shouldn't need a new input param? This should also make all changes in the above `.subscribe(...)` unnecessary.

src/app/donation-start/donation-start.component.ts
```diff
@@ -345,6 +348,7 @@ export class DonationStartComponent implements OnInit {
       'Remember every penny helps & you can continue to make an unmatched donation to the charity!',
       'Cancel',
       donation,
+      this.surplusDonationInfo,
```
NoelLH: @tbg-jasonfung  I still don't understand why we're passing this around, or adding an extra class var? We already have `campaign` as a class var, let's just use that and avoid extraneous inter-method params.

PR #240, opened 2020-03-20
DON-19 temporary campaign target logic

tbg-jasonfung:
- DON-19 temporary campaign target logic

### DIFF
--- a/src/app/campaign-details-card/campaign-details-card.component.html
+++ b/src/app/campaign-details-card/campaign-details-card.component.html
@@ -16,10 +16,10 @@
       <div *ngIf="showTemporaryTarget; else showRealTarget">
         <p class="c-campaign-target__value b-bold">{{ 1000000 | currency : '£' : 'symbol' : '1.0-0' }}</p>
       </div>
-      <ng-template #showRealTarget>
+      <div #showRealTarget>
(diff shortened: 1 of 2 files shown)


src/app/campaign-details-card/campaign-details-card.component.html
```diff
@@ -12,8 +12,14 @@
   <div class="c-card-body">
     <div class="c-campaign-target" *ngIf="campaign.target > 0">
       <h3 class="c-campaign-target__description b-grey b-light">Campaign target</h3>
-      <p class="c-campaign-target__value b-bold">{{ campaign.target | currency : '£' : 'symbol' : '1.0-0' }}</p>
-      <mat-progress-bar [value]="percentRaised" *ngIf="campaign.amountRaised > 0"></mat-progress-bar>
+      <!-- TODO remove hard-coded campaign target after COVID-19 Emergency Appeal -->
+      <div *ngIf="showTemporaryTarget; else showRealTarget">
+        <p class="c-campaign-target__value b-bold">{{ 1000000 | currency : '£' : 'symbol' : '1.0-0' }}</p>
+      </div>
+      <ng-template #showRealTarget>
```
NoelLH: Doesn't seem like a good time to be trying out a new element in a hurry @tbg-jasonfung. Why not avoid the local template var and use `*ngIf=!showTemporaryTarget` for readability, and so we're confident it will behave like others?

PR #247, opened 2020-03-23
Always show total funds raised on the right hand side of card

tbg-jasonfung:
- Always show total funds raised on the right hand side of card


tbg-jasonfung:
Decision to leave this on the left hand side

PR #249, opened 2020-03-30
DON-204 facebook pixel

tbg-jasonfung:


src/app/facebook-pixel-service/facebook-pixel-service.ts
```diff
@@ -0,0 +1,49 @@
...
+  providedIn: 'root',
+})
+export class FacebookPixelService {
+  private loadOk = false;
+
+  constructor() {}
+
+  load() {
+    if (!this.loadOk) {
+      (function(f: any, b, e, v, n, t, s) {
```
tbg-jasonfung: @NoelLH Any ideas how I can resolve this warning `on-arrow functions are forbidden` think this is causing the rest of the code to not compile correctly, for example line 17 is throwing `Type '() => void' is not assignable to type 'undefined'`

tbg-jasonfung: Okay think i've managed to convert this to an arrow function

src/app/facebook-pixel-service/facebook-pixel-service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  constructor() {}
+
+  load() {
+    if (!this.loadOk) {
+      ((f: any, b, e, v, n, t, s) => {
+        if (f.fbq) {
+          return;
+        }
+        n = f.fbq = () => {
```
tbg-jasonfung: @NoelLH Any ideas on this error here: `TS2322: Type '() => void' is not assignable to type 'undefined'` - also getting `TS2532: Object is possibly 'undefined'` for every param variable assignment

NoelLH: @tbg-jasonfung if you look at where the error is first highlighted it's the assignment to `n` which the FB snippet uses extensively. So you need to tell TypeScript a compatible type for `n`.

FB's JS is not in our control and we don't know or care what types it uses (and they are liable to change over time), so the only appropriate type is `any`.

Put that in and you'll find there is a new error because the anon fn they tell you to declare takes up to 7 args but we only pass in 4 on the initial call. So we need to tell TypeScript it's an _optional_ param of type `any`: `n?: any`

Repeat for the other params.

src/app/facebook-pixel-service/facebook-pixel-provider.ts
```diff
@@ -0,0 +1,29 @@
+import { Provider, APP_BOOTSTRAP_LISTENER, ComponentRef } from '@angular/core';
+import { Router, NavigationEnd } from '@angular/router';
+import { FacebookPixelService } from './facebook-pixel-service';
+
+export const FacebookPixelProvider: Provider = {
```
NoelLH: @tbg-jasonfung this seems like an odd and over-complicated way to init the service.

We also haven't accounted for the fact that server-side there is no `window`.

Have a look at where we call these in `AppComponent` – we need to do something analogous for the FB code.

```
      this.analyticsService.init();
      this.getSiteControlService.init();
```

src/app/facebook-pixel-service/facebook-pixel-service.ts
```diff
@@ -0,0 +1,49 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class FacebookPixelService {
```
NoelLH: How about `FacebookService` to reflect that we're loading the SDK, and the main tracking mechanism is JS rather than a pixel?

src/app/facebook-pixel-service/facebook-pixel-service.ts
```diff
@@ -0,0 +1,49 @@
...
+  private loadOk = false;
```
NoelLH: I don't think we should need this var

src/app/facebook-pixel-service/facebook-pixel-service.ts
```diff
@@ -0,0 +1,49 @@
...
+        document,
+        'script',
+        'https://connect.facebook.net/en_US/fbevents.js',
+      );
+      (window as any).fbq('init', '272508403759002');
+      (window as any).fbq('track', 'PageView');
+      this.loadOk = true;
+      console.log('Facebook pixel init run!');
+    } else {
+      (window as any).fbq('track', 'PageView');
```
NoelLH: Don't think we need to track every page view for our use case. If we do, we should do it like GA, not here.

src/app/facebook-pixel-service/facebook-pixel-service.ts
tbg-jasonfung: @NoelLH Thank you, that's really helpful - its now compiling okay

tbg-jasonfung:
@NoelLH This seems to be compiling okay now. I'm now getting `[Facebook Pixel] - You must provide an argument to fbq().` within chrome console and using `Facebook Pixel Helper` chrome extension, it doesn't detect the Facebook pixel. Googling the warning doesn't lead me anywhere, can you pull down the changes and try on your chrome?

NoelLH:
> @NoelLH This seems to be compiling okay now. I'm now getting `[Facebook Pixel] - You must provide an argument to fbq().` within chrome console and using `Facebook Pixel Helper` chrome extension, it doesn't detect the Facebook pixel. Googling the warning doesn't lead me anywhere, can you pull down the changes and try on your chrome?

I get the same @tbg-jasonfung. The pixel not being detected as loading by the extension might be specific to local use or the way we load it. However I think we can be pretty confident that as the lib itself gives an error, the approach will not work.

I've got what I think is a safer way to load the library initially on my local. However if you think about what we're doing here it still doesn't make sense in terms of tracking conversions. Nobody has documented how we expect the lib to be used, but as I understand it there are three possibilities, none of which could work with this branch / our GTM in their current state:

1. We send every `PageView`, like with GTM/GA, even though we're only really meant to be focusing on conversions. Facebook would be configured with a [Custom Conversion](https://developers.facebook.com/docs/facebook-pixel/implementation/conversion-tracking#custom-conversions) to treat viewing `/thanks/{...}` as a conversion. This would require subscribing to a router event as with GA. It feels a bit brittle for conversion tracking since links could be shared around and this doesn't verify that the page loaded successfully / the donor had a token. And it means conversion tracking can't include donation value.
2. We [send a `Purchase` event](https://developers.facebook.com/docs/facebook-pixel/implementation/conversion-tracking#standard-events) when loading the thank you page with a valid token. This feels like the best option to me but we need to clarify with whoever is managing the Facebook account that it's what they expect and whether we can use this _instead_ of sending `PageView` events.
3. We [load the Facebook pixel inside GTM](https://www.facebook.com/business/help/1021909254506499) and use GTM events to trigger Facebook events. This sounds appealing initially but would probably be very hard to make work properly with a single-page app and super brittle because CCh event changes (they trigger GA ecommerce events currently) would necessitate GTM changes rather than each piece being independent.

tbg-jasonfung:
> > @NoelLH This seems to be compiling okay now. I'm now getting `[Facebook Pixel] - You must provide an argument to fbq().` within chrome console and using `Facebook Pixel Helper` chrome extension, it doesn't detect the Facebook pixel. Googling the warning doesn't lead me anywhere, can you pull down the changes and try on your chrome?
> 
> I get the same @tbg-jasonfung. The pixel not being detected as loading by the extension might be specific to local use or the way we load it. However I think we can be pretty confident that as the lib itself gives an error, the approach will not work.
> 
> I've got what I think is a safer way to load the library initially on my local. However if you think about what we're doing here it still doesn't make sense in terms of tracking conversions. Nobody has documented how we expect the lib to be used, but as I understand it there are three possibilities, none of which could work with this branch / our GTM in their current state:
> 
> 1. We send every `PageView`, like with GTM/GA, even though we're only really meant to be focusing on conversions. Facebook would be configured with a [Custom Conversion](https://developers.facebook.com/docs/facebook-pixel/implementation/conversion-tracking#custom-conversions) to treat viewing `/thanks/{...}` as a conversion. This would require subscribing to a router event as with GA. It feels a bit brittle for conversion tracking since links could be shared around and this doesn't verify that the page loaded successfully / the donor had a token. And it means conversion tracking can't include donation value.
> 2. We [send a `Purchase` event](https://developers.facebook.com/docs/facebook-pixel/implementation/conversion-tracking#standard-events) when loading the thank you page with a valid token. This feels like the best option to me but we need to clarify with whoever is managing the Facebook account that it's what they expect and whether we can use this _instead_ of sending `PageView` events.
> 3. We [load the Facebook pixel inside GTM](https://www.facebook.com/business/help/1021909254506499) and use GTM events to trigger Facebook events. This sounds appealing initially but would probably be very hard to make work properly with a single-page app and super brittle because CCh event changes (they trigger GA ecommerce events currently) would necessitate GTM changes rather than each piece being independent.

Okay thanks @NoelLH - have you got the changes for safely loading the library you could push to this branch?

Also I found a similar solution here which loads the service through app.component.ts: https://stackoverflow.com/a/57182368/6620085 and in the same post, another solution is to load the JS through the header (like we do with modernizr) and declare a const `fbq`: https://stackoverflow.com/a/49521128/6620085

NoelLH:
@tbg-jasonfung I'll carry on tweaking and push my changes in a bit. I think basically we can treat it a lot more like Google Analytics than I realised before.

We already init the service from `AppComponent` so I don't understand what you're suggesting there? My change doesn't impact this, it just build the script tag more like we do for GA/GTM.

My WIP change doesn't load the lib like we do Modernizr per se (that won't work because we need to be able to get any change FB make), but it does add an ambient type declaration like we do for GA. Mine is better IMO because it's more specific so better guards against code errors we might make (though I have just changed it to `const` having looked at that post):

```
declare const fbq: (action: string, eventOrId: string, params?: object) => void;
```

This is to do with whether our app knows what the global `fbq` is. It's got nothing to do with staging having an ID set or not. I don't see why we can't set up a staging pixel, but regardless the service needs to

- know what `fbq` is, and
- be able to skip doing anything seamlessly if there is no ID set

PR #250, opened 2020-03-31
WIP: Alternative way to load facebook js

tbg-jasonfung:
image


tbg-jasonfung:
@NoelLH As suggested in my other PR, an alternative way to load the JS through the classic convention - but just demonstrating the pixel helper does detect this in local

tbg-jasonfung:
Demonstration PR, no action required.

PR #252, opened 2020-04-03
DON-205 remove covid19 donation limit

tbg-jasonfung:
- DON-205 remove covid19 donation limit + leave in abstract control for dynamic max amount validation

### DIFF
--- a/src/app/donation-start/donation-start.component.html
+++ b/src/app/donation-start/donation-start.component.html
@@ -47,7 +47,7 @@
 
             <div class="error" *ngIf="f.donationAmount.invalid && f.donationAmount.errors && (f.donationAmount.dirty || f.donationAmount.touched)">
               <div *ngIf="f.donationAmount.errors.min">Sorry, the minimum donation is currently £5.</div>
-              <div *ngIf="f.donationAmount.errors.max">Your donation must be £25,000 or less to proceed. You can make multiple matched donations of £25,000 if match funds are available.</div>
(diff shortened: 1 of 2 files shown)


src/app/donation-start/donation-start.component.html
```diff
@@ -47,7 +47,7 @@ <h3 class="b-rh-2">Your donation</h3>
 
             <div class="error" *ngIf="f.donationAmount.invalid && f.donationAmount.errors && (f.donationAmount.dirty || f.donationAmount.touched)">
               <div *ngIf="f.donationAmount.errors.min">Sorry, the minimum donation is currently £5.</div>
-              <div *ngIf="f.donationAmount.errors.max">Your donation must be {{ maximumDonationAmount | currency : '£' : 'symbol' : '1.0-0' }} or less to proceed. You can make multiple matched donations of {{ maximumDonationAmount | currency : '£' : 'symbol' : '1.0-0' }} if match funds are available.</div>
+              <div *ngIf="f.donationAmount.errors.max">Your donation must be £25,000 or less to proceed. You can make multiple matched donations of £25,000 if match funds are available.</div>
```
NoelLH, line 50: @tbg-jasonfung sorry, I forgot this part was also changed with the constraint commit. I suggest we keep this part dynamic and put the public property back (just setting it from the env var directly), as the overhead should be minimal and one-shot (whereas I suspect dynamic validator fns which keep re-evaluating are heavier). It's good to have the config in only one place.

tbg-jasonfung: No probs, that's update now

PR #255, opened 2020-04-08
DON-206 Revert don201 campaign target

tbg-jasonfung:
Revert "DON-201 temporary campaign target logic"

This reverts commit 480dbd1479cf061ed37d5728379b29ae0cf00e79.


tbg-jasonfung:
Wrong commit message, should be `DON-201` instead of `DON-19`

PR #295, opened 2020-07-28
DON-248, DON-232, DON-234 + more

NoelLH:
* split up Create & Update, reserving match funds earlier in the form
* prompt for various additional fields
* rework form structure & corresponding tests
* rework UI to use a Stepper
* various hint, form design & copy enhancements
* global styling updates:
  * flip `primary` and `accent` to better reflect the real brand colour usage
  * globally have radio buttons use the primary colour
* enable Stripe dynamically based on each charity's setup status
* improve and complete form validation
* improve Stripe error handling
* reduce minimum donation to £1


src/environments/environment.regression.ts
```diff
@@ -18,7 +18,7 @@ export const environment = {
       initUri: 'https://fundraise.charitycheckouttest.co.uk/api/checkout/init',
     },
     stripe: {
-      enabled: false,
+      enabled: true,
```
tbg-jasonfung, line 21: @NoelLH Isn't this going to clash with Enthuse if both are `true` ?

src/environments/environment.staging.ts
```diff
@@ -16,7 +16,7 @@ export const environment = {
       initUri: 'https://fundraise.charitycheckouttest.co.uk/api/checkout/init',
     },
     stripe: {
-      enabled: false,
+      enabled: true,
```
tbg-jasonfung, line 19: @NoelLH Same as above

src/environments/environment.regression.ts
NoelLH: @tbg-jasonfung no, see https://thebiggive.atlassian.net/browse/DON-254 – We need both to be true. Have a look at `DonationStartComponent.handleCampaign()` for how this is used.

PR #311, opened 2020-08-19
DON-268 cancelling partially matched donation

tbg-jasonfung:
- DON-268 when cancelling partially matched donation, remove donation + force stepper back a step + silently reload component

### DIFF
--- a/src/app/donation-start/donation-start.component.ts
+++ b/src/app/donation-start/donation-start.component.ts
@@ -821,6 +821,9 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
           () => {
             this.analyticsService.logEvent('cancel', `Donor cancelled donation ${donation.donationId} to campaign ${this.campaignId}`),
             this.donationService.removeLocalDonation(donation);
+            delete this.donation;
+            this.stepper.previous();
+            this.ngOnInit();
           },
           response => {
             this.analyticsService.logError(


src/app/donation-start/donation-start.component.ts
```diff
@@ -821,6 +821,14 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
...
             this.analyticsService.logEvent('cancel', `Donor cancelled donation ${donation.donationId} to campaign ${this.campaignId}`),
             this.donationService.removeLocalDonation(donation);
+
+            // Removes match funds reserved timer if present
+            delete this.donation;
+
+            // Go back to previous step to encourage donor to try again
+            this.stepper.previous();
+
+            this.ngOnInit();
```
NoelLH, line 831: @tbg-jasonfung It doesn't seem like a great idea to manually call a [lifecycle hook](https://v2.angular.io/docs/ts/latest/guide/lifecycle-hooks.html) event. By doing this manually inside our logic we could end up with hook-related logic executing in a different, untested order in edge cases.

This also feels like overkill – can't we just remove the donation object and leave the form data as it was? I certainly don't think we need to do things like re-render the form component itself as if the whole page was new.

src/app/donation-start/donation-start.component.ts
```diff
@@ -821,6 +821,14 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
           () => {
             this.analyticsService.logEvent('cancel', `Donor cancelled donation ${donation.donationId} to campaign ${this.campaignId}`),
             this.donationService.removeLocalDonation(donation);
+
+            // Removes match funds reserved timer if present
+            delete this.donation;
+
+            // Go back to previous step to encourage donor to try again
+            this.stepper.previous();
```
NoelLH, line 829: We don't know what step the donor is on when this asynchronous process completes. What about `reset()`?

NoelLH (changes requested):

PR #314, opened 2020-08-24
DON-268 double check amount is greater than 0 before creating

tbg-jasonfung:
- DON-268 double check amount is greater than 0 before creating

### DIFF
--- a/src/app/donation-start/donation-start.component.ts
+++ b/src/app/donation-start/donation-start.component.ts
@@ -274,7 +274,10 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
         activeStepLabel.scrollIntoView({ behavior: 'smooth', block: 'start' });
     }
 
-    if (event.previouslySelectedStep.label === 'Your donation' && (this.previousDonation === undefined || this.previousDonation.status === 'Cancelled')) {
+    // Only create donation if step index is greater than 0, and previous donation is invalid.
+    // There are situations when `stepper.reset()` is called and relying on previous event labels
+    // would trigger `createDonation()` with an empty Donation object.
+    if (event.selectedIndex > 0 && (this.previousDonation === undefined || this.previousDonation.status === 'Cancelled')) {
       this.createDonation();
 
       if (this.psp === 'stripe') {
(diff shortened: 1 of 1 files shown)


tbg-jasonfung:
This PR should address incorrect error message showing when cancelling a donation, whether it's part of the re-use dialog box or partially matched dialog box

src/app/donation-start/donation-start.component.ts
```diff
@@ -582,11 +584,15 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
       tipAmount: this.sanitiseCurrency(this.amountsGroup.value.tipAmount),
     };
 
-    // No re-tries for create() where donors have only entered amounts. If the
-    // server is having problem it's probably more helpful to fail immediately than
+    // Only create if amount is greater than 0, we check here because
```
NoelLH, line 587: @tbg-jasonfung this doesn't really make sense – `createDonation()` is called based on a step change and conditionally, but not at an unpredictable time after an AJAX call. I don't see why we wouldn't just not call it in this case?

I'm also not convinced this fixes the problem although I'm struggling to now see the exact situation I ended up in, where the form values were non-zero (meaning this check would have passed and tried to make the donation) but the donation object had zeroes.

Looking at the XHR requests, we also have a more fundamental issue which needs fixing here, in that simply going back to step 1 and then forward again creates new donations. We need the condition to check for a 'live' donation instead of just the previous one before deciding it needs to `createDonation()` afresh.

NoelLH (changes requested):

src/app/donation-start/donation-start.component.ts
tbg-jasonfung: @NoelLH I've updated the stepper check and replaced with an index check instead, this will help identify donations that are live and new

NoelLH: @tbg-jasonfung this actually expands the cases in which we create a donation. If you look at the network traffic you will see it still creates extra donations where we should be aiming for reuse. We need to look at the other donation var or set `previousDonation` in more cases for either to work.

This will actually also introduce a new, more severe bug whereby going from e.g. step 2 to 3 will try to create a new donation.

NoelLH (approved):

PR #316, opened 2020-08-25
DON-274 prevent multiple donations from being created

tbg-jasonfung:
- DON-274 fix edge case bug where returning to 1st step and clicking next would create a new donation despite amount being the same

### DIFF
--- a/src/app/donation-start/donation-start.component.ts
+++ b/src/app/donation-start/donation-start.component.ts
@@ -304,8 +304,6 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
 
     const invalidPreviousDonation = (this.previousDonation === undefined || this.previousDonation.status === 'Cancelled');
 
-    console.log('activelySelectedNext: ' + activelySelectedNext + ' invalidDonation: ' + invalidDonation + ' invalidPreviousDonation: ' + invalidPreviousDonation);
-
     // Create a donation if user actively clicks 'next' from first step and
     // the current and previous donations are invalid.
     if (activelySelectedNext && invalidDonation && invalidPreviousDonation) {
--- a/src/app/donation-start/donation-start.component.ts
+++ b/src/app/donation-start/donation-start.component.ts
@@ -242,7 +242,7 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
 
     // If the original donation amount was updated, cancel that donation,
     // we'll create a new one later on for this updated amount.
-    if (this.donation !== undefined && this.donationAmount !== this.donation.donationAmount) {
+    if (this.donation !== undefined && this.donationAmount > 0 && this.donationAmount !== this.donation.donationAmount) {
       this.analyticsService.logEvent(
         'cancel_auto',
         `Donation cancelled because amount was updated ${this.donation.donationId} to campaign ${this.campaignId}`,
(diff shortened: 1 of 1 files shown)


src/app/donation-start/donation-start.component.ts
```diff
@@ -239,6 +239,17 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
...
   async stepChanged(event: StepperSelectionEvent) {
+
+    // If the original donation amount was updated, cancel that donation,
+    // we'll create a new one later on for this updated amount.
+    if (this.donation !== undefined && this.donationAmount !== this.donation.donationAmount) {
+      this.analyticsService.logEvent(
+        'cancel_auto',
+        `Donation cancelled because amount was updated ${this.donation.donationId} to campaign ${this.campaignId}`,
+      );
+      this.donationService.cancel(this.donation).subscribe(() => this.donationService.removeLocalDonation(this.donation));
```
NoelLH, line 250: What about `this.donation`?

src/app/donation-start/donation-start.component.ts
```diff
@@ -284,10 +295,14 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
...
 
-
-    // Only create donation if user actively clicks 'next' from first step, and previous donation is invalid.
+    // Proceed if user actively clicks 'next' from first step, and previous donation is invalid.
     if (activelySelectedNext && (this.previousDonation === undefined || this.previousDonation.status === 'Cancelled')) {
-      this.createDonation();
+
+      // Create donation if one hasn't been defined yet or
+      // if the original donation amount was updated.
+      if (this.donation === undefined || (this.donation && this.donationAmount !== this.donation.donationAmount)) {
```
NoelLH, line 303: If the cancellation works we should be able to avoid having a separate check. Having 2 overlapping async things seems like asking for trouble with timing issues – we ought to be able to do the checks in 1 place.

NoelLH (changes requested):

src/app/donation-start/donation-start.component.ts
tbg-jasonfung: @NoelLH I don't think I follow, the if check ensures the form amount is different to the object amount, and it's then cancelling the existing donation

tbg-jasonfung: @NoelLH Doing a delete on `this.donation` when the amount changes causes the offer existing donation dialog box to not appear on refresh, I've tested the current approach and it works for all scenarios

NoelLH: @tbg-jasonfung yes, as per your edit, they can't reuse a cancelled donation. So shouldn't we put the local var deletion in?

tbg-jasonfung: @NoelLH Once the amount has changed and the existing donation is cancelled and a new donation is created, even then on refresh the existing donation dialog box doesn't appear :/

NoelLH: @tbg-jasonfung that can't be to do with the removal of `this.donation`, which is only in local state, since that doesn't impact the new page on refresh. So I don't understand why we're not removing it? It sounds like there may also be a separate issue either way in the case you describe.

tbg-jasonfung: @NoelLH Think i figured it out, it was deleting the local donation before `removeLocalDonation` was being triggered, so on refresh it found the donation that was `Cancelled` but there's a method that says, don't offer this as a previous donation if the status is not `Pending` or  `Reserved`

NoelLH (approved):

PR #317, opened 2020-08-27
DON-275 start sending charity logo to enthuse + update models and tests

tbg-jasonfung:
- DON-275 start sending charity logo to enthuse + update models and tests

### DIFF
--- a/src/app/donation-start/donation-start.component.ts
+++ b/src/app/donation-start/donation-start.component.ts
@@ -715,7 +715,7 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
     });
   }
 
-  private redirectToEnthuse(donation: Donation, logoUri: any) {
+  private redirectToEnthuse(donation: Donation, logoUri?: string) {
     this.analyticsService.logAmountChosen(donation.donationAmount, this.campaignId, this.suggestedAmounts);
     this.analyticsService.logEvent('payment_redirect_click', `Donating to campaign ${this.campaignId}`);
     this.charityCheckoutService.startDonation(donation, logoUri);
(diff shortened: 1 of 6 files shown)


src/app/donation-start/donation-start.component.ts
```diff
@@ -373,6 +373,11 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
       )
       .subscribe(async (donation: Donation) => {
         if (donation.psp === 'enthuse') {
+
+          // Because we're not storing the charity logo in the DB
```
NoelLH, line 377: Given this is the case and it's not needed for the other mode, I think it would be better to add it as a second param for the Enthuse-specific method calls rather than adding it to the main `Donation` model.

NoelLH (changes requested):

src/app/donation-start/donation-start.component.ts
tbg-jasonfung: @NoelLH Okay, makes sense, i've passed in the uri as a 2nd param now and updated tests

src/app/donation-start/donation-start.component.ts
```diff
@@ -715,10 +715,10 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
     });
   }
 
-  private redirectToEnthuse(donation: Donation) {
+  private redirectToEnthuse(donation: Donation, logoUri: any) {
```
NoelLH, line 718: Type can be `?string`?

tbg-jasonfung: Hmm doesn't seem to like it @NoelLH, getting this `Argument of type 'string | undefined' is not assignable to parameter of type 'string'. Type 'undefined' is not assignable to type 'string'.`

NoelLH: @tbg-jasonfung it should be fine as long as the optional bit (`?` / `undefined`) is consistent across methods

tbg-jasonfung: Ahh cool, got it!

NoelLH (approved):

PR #318, opened 2020-09-02
DON-274 on submit, get the latest tip amount value

tbg-jasonfung:
- DON-274 on submit, get the latest tip amount value


NoelLH (changes requested):
Why not update this in `stepChanged()` where we do most properties, so the local copy is in sync sooner?

NoelLH (approved):

PR #328, opened 2020-09-23
DON-284 use api values for opt in statement

tbg-jasonfung:
- DON-284 use api values for opt in statement

---

image


src/app/donation-start/donation-start.component.html
```diff
@@ -179,7 +179,7 @@ <h1 class="b-rh-1 b-bold">{{ campaign.title }}</h1>
             <p class="b-rt-1 b-bold" id="optInCharityEmail-label">Would you be happy to receive emails from {{ campaign.charity.name }}?</p>
             <mat-radio-group tabindex="0" aria-labelledby="optInCharityEmail-label" formControlName="optInCharityEmail">
               <!-- TODO update this hint to vary based on Gift Aid selection and be clearer -->
-              <mat-hint>This may include thanks for your donation, updates on how it is being used, newsletters and future campaigns. If you select no, we will pass on your details as these may be required for Gift Aid purposes, but the charity will be informed that you do not wish to receive communications.</mat-hint>
+              <mat-hint>{{ campaign.charity.charityOptInStatement }}</mat-hint>
```
NoelLH, line 182: We should have a default, at least until this is fully rolled out on both apps so we don't have to release in a specific order

src/app/donation-start/donation-start.component.ts
```diff
@@ -871,4 +875,12 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
         );
     };
   }
+
+  private setchampionOptInValidity() {
```
NoelLH, line 879: Odd casing

src/app/donation-start/donation-start.component.ts
```diff
@@ -871,4 +875,12 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
...
+    if (this.campaign?.championOptInStatement !== null) {
```
NoelLH, line 880: This conditional should be identical to the template so we can't have glitches with e.g. blank strings where only one of the two matches. May be simpler to set a public boolean var and use that in both?

NoelLH (commented):
Approach looks good. A few small things inline + suggested API contract tweak on other PR

NoelLH (approved):

PR #329, opened 2020-10-14
DON-291 refine meta campaign page

tbg-jasonfung:
- DON-175 implement query params
- DON-291 styling and UI refinement


NoelLH (changes requested):
I'm really not sure about using local storage for this at all, and especially having a whole separate service distinct from existing local storage. Why not use URL params as suggested on DON-175 which you linked?

tbg-jasonfung:
Leaving in draft to keep develop branch clean before Full Sandbox refresh.

src/app/meta-campaign/meta-campaign.component.scss
```diff
@@ -15,3 +15,58 @@
...
+.align-center {
+  display: block;
+  margin: auto;
+}
+
+.padding-top-s {
+  padding-top: 2rem;
+}
+
+.padding-top-m {
```
NoelLH, line 28: I'm not sure about adding lots more utility classes, especially ones that aren't used and especially when the element already has a more descriptive class applied to it. This all adds to the bundle size.

src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -97,7 +103,11 @@ export class MetaCampaignComponent implements OnInit {
     }
   }
 
-  setDefaultFilters() {
+  /**
+   * @method  setDefaultFilters
+   * @param   setQueryParams determine whether to override query with params in the URL
+   */
+  setDefaultFilters(setQueryParams: boolean) {
```
NoelLH, line 110: Why wouldn't we just always set the query params?

src/app/meta-campaign/meta-campaign.component.html
```diff
@@ -1,48 +1,94 @@
...
+          mat-raised-button
+          color="primary"
+          type="submit"
+          (click)="loadMore()"
+        >
+          Load More
+        </button>
+      </div>
+
+      <div *ngIf="!hasMore" class="t-align-center padding-top-s" >
```
NoelLH, line 61: Why do we need this? And why are we removing infinite scroll for the meta-campaign page?

src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -199,5 +229,48 @@ export class MetaCampaignComponent implements OnInit {
...
+    } else if (campaign.video && campaign.video.provider === 'vimeo') {
+      this.videoEmbedUrl = this.sanitizer.bypassSecurityTrustResourceUrl(`https://player.vimeo.com/video/${campaign.video.key}`);
+    }
+  }
+
+  /**
+   * Set query params to this.query object, if any are available.
+   */
+  private setQueryParams() {
+    if (window.location.search) {
```
NoelLH, line 244: This is not SSR safe (possibly why the boolean param workaround was needed). We should get params from Angular-injected params instead. This will be more durable e.g. when we move to loading data earlier such that presence of `campaign` is not an indicator of whether we are server-rendering or not.

src/app/campaign.service.ts
```diff
@@ -11,6 +11,7 @@ import { environment } from '../environments/environment';
 })
 export class CampaignService {
   private apiPath = '/campaigns/services/apexrest/v1.0';
+  private readonly storageKey = `${environment.donateUriPrefix}/campaignFilters/v2`; // Key is per-domain/env
```
NoelLH, line 14: Redundant?

NoelLH (changes requested):
Query param approach looks good so far. A few things inline, I think the main one being that I don't think we need / should add user controlled pagination in preference to infinite scrolling

src/app/meta-campaign/meta-campaign.component.ts
tbg-jasonfung: @NoelLH When `onClearFiltersApplied()` is triggered I've reused `setDefaultFilters()` to clear all sort / filters - and within `setDefaultFilters()` if `setQueryParams()` is triggered every time, the subscribed `route.queryParams` is overriding the clear method and essentially nothing is emptied.

src/app/explore/explore.component.html
```diff
@@ -23,6 +23,7 @@ <h1 class="b-rh-1 b-bold c-home-hero__heading">The leading online match funding
   <div class="b-container" *ngIf="campaigns">
     <app-filters
       [hasTerm]="hasTerm"
+      [isHomepage]="true"
```
NoelLH, line 26: I've just realised this is saying that Explore is the home page. What I was anticipating when we discussed splitting out the pages and I split them on Jira was a new Home component which should be the only one working in `isHomepage` mode. So we have 3 grids:

* Home, which has a limited number and no filters and no infinite scroll, but directs you to Explore if you want to see more campaigns
* Explore which has filters and infinite scroll
* Meta-campaign which has filters and infinite scroll and is scoped to one parent campaign

src/app/meta-campaign/meta-campaign.component.ts
NoelLH: This sounds like something that should be fixed elsewhere then @tbg-jasonfung. Stuff you've done on the page should take precedence if these two things diverge. Moreover, they should not ever diverge because updating the form should immediately update the route, just like other navigation events.

NoelLH (changes requested):

src/app/explore/explore.component.html
tbg-jasonfung: I've renamed the var to match it's intention so it's less confusing when we start creating a new home page component

src/app/meta-campaign/meta-campaign.component.ts
tbg-jasonfung: I've added a new method called `setDefaultFiltersOrParams()` which helps identify exactly its purpose when query params exist. This gets rid of the Boolean flag and also resolves the filters not clearing.

src/app/explore/explore.component.html
NoelLH: Ah, I think I was just confused about the order we're doing things @tbg-jasonfung. So what I think should continue to be Explore will have this set to `true` too. But for today, Explore continues to be the home page and so it's `false`. Let me know if that's wrong, as I think then we might still need to clarify?

src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -154,12 +154,16 @@ export class MetaCampaignComponent implements OnInit {
   }
 
   onClearFiltersApplied() {
+    // Remove any query params from URL
```
NoelLH, line 157: Nice

NoelLH (approved):

NoelLH:
Changes look good, nice one. Ready to go I think if there's nothing keeping it in draft

src/app/explore/explore.component.html
tbg-jasonfung: @NoelLH that's right, once we've split out the home component and explore component, then we can set that flag to `true` since we want clear filters button to show on the explore page.

PR #330, opened 2020-10-16
DON-298 update comms pref sections

tbg-jasonfung:
- DON-298 dynamically define opt in copy from campaign API + accommodate for backwards compatibility
- Update unit test campaign model structure


tbg-jasonfung:
Merging in favour of Full sandbox refresh.

src/app/analytics.service.spec.ts
```diff
@@ -47,6 +47,17 @@ describe('AnalyticsService', () => {
         regulatorNumber: '123456',
         regulatorRegion: 'Scotland',
       },
+      {
+        charityOptIn: 'Charity opt in',
```
NoelLH, line 51: @tbg-jasonfung what's going on with all of these?!

tbg-jasonfung: @NoelLH it's custom labels I mentioned [here](https://thebiggive.slack.com/archives/DCJLDGN2J/p1603095146002400)

NoelLH: @tbg-jasonfung these are still not documented per-value anywhere I can see, and it doesn't seem like the examples are realistic so the tests don't help at all in understanding the purposes.

tbg-jasonfung: @NoelLH this might help paint a better picture: https://github.com/thebiggive/donate-frontend/pull/332/files

PR #332, opened 2020-10-19
DON-298 Update unit tests with realistic test data

tbg-jasonfung:
- Update unit tests with realistic test data


NoelLH (commented):
We'll likely replace this with a simpler approach as discussed today

tbg-jasonfung:
Not required anymore

PR #340, opened 2020-10-29
DON-279 DON-280 featured campaigns and remove filters

tbg-jasonfung:
- DON-279 new home page component defaulted to 6 cards + default query attributes
- Update explore component styling classes to match component name
- Update explore campaign route links
- DON-279 add /explore as new route path


src/app/home-page/home-page.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { CampaignService, SearchQuery } from '../campaign.service';
+import { CampaignSummary } from '../campaign-summary.model';
+
+@Component({
+  selector: 'app-home-page',
+  templateUrl: './home-page.component.html',
+  styleUrls: ['./home-page.component.scss'],
+})
+export class HomePageComponent implements OnInit {
```
NoelLH, line 12: I think this should be `HomeComponent` (and file names aligned) at least for now, for consistency with existing pages. If we think it's confusing we should rename everything en masse in a separate PR

src/app/home-page/home-page.component.ts
```diff
@@ -0,0 +1,55 @@
...
+      this.campaigns = campaignSummaries; // Success
+      this.loading = false;
+    }, () => {
+        this.loading = false;
+      },
+    );
+  }
+
+  search(term: string) {
+    // TODO DON-303 redirect to /explore page with query param
```
NoelLH, line 53: I don't think we can keep the button in the template without this working properly, so we need to take it out for now for this to be safe to merge.

NoelLH (changes requested):
Mostly looks good – couple of things inline

NoelLH (approved):

PR #341, opened 2020-10-30
DON-281 - about tbg section

tbg-jasonfung:
- DON-281 Implement tbg about section following figma designs


tbg-jasonfung:
This PR branches off [340](https://github.com/thebiggive/donate-frontend/pull/340) - therefore the diffs have added noise until we merge this branch in - keeping in draft for now.

src/app/home-page/home-page.component.ts
```diff
@@ -13,18 +14,22 @@ export class HomePageComponent implements OnInit {
...
   constructor(
     private campaignService: CampaignService,
+    private sanitizer: DomSanitizer,
   ) {
   }
 
   ngOnInit() {
     this.setDefaults();
     this.run();
+    this.videoEmbedUrl = this.sanitizer.bypassSecurityTrustResourceUrl(`https://www.youtube.com/embed/${this.youtubeId}`);
```
NoelLH, line 32: As this one's not dynamic I wonder if we should just hardcode it in the template to avoid the weight of all this?

NoelLH (approved):

PR #345, opened 2020-11-03
DON-312 – filters & search

NoelLH:
* improve filters' UX
* refactor, reducing boilerplate, simplifying and ensuring behaviour doesn't depend on whether campaigns were cached in state or whether the component already loaded this session


src/app/campaign.service.ts
```diff
@@ -44,6 +47,23 @@ export class CampaignService {
     return 100 * campaign.amountRaised / campaign.target;
   }
 
+  buildQuery(selected: SelectedType, offset: number): {[key: string]: any} {
+    const perPage = 6;
+    const query: SearchQuery = {
```
tbg-jasonfung, line 52: @NoelLH the meta campaign pages are including child campaigns that aren't associated to the master, I think because `parentCampaignId` here isn't set when loading the meta pages.

tbg-jasonfung:
@NoelLH loading the filters from the query params is now broken

NoelLH:
> @NoelLH loading the filters from the query params is now broken

In what way on what page @tbg-jasonfung? I tested a lot of permutations so more direction would be helpful!

tbg-jasonfung:
This looks good now, @NoelLH i've added a quick [change](https://github.com/thebiggive/donate-frontend/pull/345/commits/67f461d0e6ac807668866bb21bf232850207c1d4) to this PR - it fixes the conditional clear filters button logic

PR #346, opened 2020-11-04
DON-307 reduce scope of global classes to just donation input fields

tbg-jasonfung:
- BG2-307 reduce scope of global classes to just donation input fields

### DIFF
--- a/src/assets/scss/_overwrite-mat.scss
+++ b/src/assets/scss/_overwrite-mat.scss
@@ -81,7 +81,7 @@ app-campaign-card {
   padding: 0.3em 0 0.7em 0;
 }
 
-.mat-form-field-appearance-legacy .mat-form-field-infix {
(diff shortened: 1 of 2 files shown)


tbg-jasonfung:
Just to keep things moving and allow for production deploy, merging this now

src/assets/scss/_overwrite-mat.scss
```diff
@@ -81,7 +81,7 @@ app-campaign-card {
   padding: 0.3em 0 0.7em 0;
 }
 
-.mat-form-field-appearance-legacy .mat-form-field-infix {
+.donation-input .mat-form-field-appearance-legacy .mat-form-field-infix {
```
NoelLH, line 84: @tbg-jasonfung can these not just apply to the specific fields and live on the component instead of in the global file?

PR #350, opened 2020-11-05
DON-307 make tip input to tbg smaller and less instrusive

tbg-jasonfung:
- DON-307 make tip input to tbg smaller and less instrusive

### DIFF
--- a/src/app/donation-start/donation-start.component.scss
+++ b/src/app/donation-start/donation-start.component.scss
@@ -100,6 +100,18 @@
   font-size: 22px;
 }
 
+.c-your-donation__form-field-to-tbg {
+  max-width: 300px;
+  margin: 0 auto;
+  display: block;
+  font-weight: 700;
+  font-size: 15px;
+
+  @media only screen and (max-width: 650px) {
+    max-width: 200px;
+  }
+}
+
 .c-your-donation__placeholder {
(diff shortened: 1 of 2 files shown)


src/app/donation-start/donation-start.component.scss
```diff
@@ -100,6 +100,18 @@
   font-size: 22px;
 }
 
+.c-your-donation__form-field-to-tbg {
```
NoelLH, line 103: @tbg-jasonfung the QA feedback was around making these consistent with each other, so couldn't they just share a class?

tbg-jasonfung: @NoelLH think the change is to make the tip input field smaller as outlined by Babsy

tbg-jasonfung: @NoelLH sorry I just realised I tagged the wrong ticket, keep doing that recently...see this ticket: https://thebiggive.atlassian.net/browse/DON-307

NoelLH: Hm, OK. This seems to be the opposite of the suggestion on DON-308 but I agree it makes sense for it to be smaller.

NoelLH (approved):

PR #355, opened 2020-11-08
DON-311 - remove Explore links for now

NoelLH:


tbg-jasonfung:
@NoelLH I think we need to use the dates set on the env vars for this one - we've re-opened DEC coronavirus appeal [campaign ](https://donate.thebiggive.org.uk/campaign/a051r00001fOoWJAA0) which still has 15 days duration left, they'll want a way for people to search for their campaign.

tbg-jasonfung (approved):

PR #356, opened 2020-11-08
DON-310 - make suggested amounts 'sticky'

NoelLH:
* add logic to remember suggestions per device and move it to DonationService
* remove logic to pass amounts from server to browser state
* update the point at which we tell GA about chosen amount, so it works with Stripe mode too


tbg-jasonfung:
[Commit](https://github.com/thebiggive/donate-frontend/pull/356/commits/4b7f1d30fda21c92631cfab0dd816dd791d39daa) fixes conflicts on `package.json`

tbg-jasonfung (approved):

PR #357, opened 2020-11-09
Update dockerfile with fontawesome config instructions

tbg-jasonfung:


tbg-jasonfung:
@NoelLH not sure if this is the right change to address the latest build failures?

NoelLH:
> @NoelLH not sure if this is the right change to address the latest build failures?

I've pushed a suggested tweak @tbg-jasonfung. I think hopefully this will work (my push to `develop` directly didn't because of the arg setup order) without repeating the same config we have in `.npmrc`.

NoelLH (approved):

PR #360, opened 2020-11-10
redirect root path to CC20

tbg-jasonfung:
- redirect root path to CC20

N.B revert this PR after Staging user testing, then do the same change for this [ticket](https://thebiggive.atlassian.net/browse/DON-319)

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -52,7 +52,7 @@ const routes: Routes = [
   },
   {
     path: '',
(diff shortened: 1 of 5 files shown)


NoelLH (commented):
@tbg-jasonfung I think we should probably put this behind a feature flag so we can keep changes moving through to Prod, especially so close to the live campaign

tbg-jasonfung:
Merging now to prepare for E2E testing later this afternoon

PR #364, opened 2020-11-12
DON-336 campaign card ordering and tweaks

tbg-jasonfung:
- Add filter description headers + left align last child + rename css class for reusability
- DON-336 campaign card ordering and tweaks

### DIFF
--- a/src/app/campaign-details-card/campaign-details-card.component.scss
+++ b/src/app/campaign-details-card/campaign-details-card.component.scss
@@ -16,6 +16,9 @@
 
 .c-icon-value {
   padding-left: 0.2rem;
+  .ng-fa-icon {
+    color: $colour-grey;
+  }
 }
 
 .c-banner__image {
@@ -65,6 +68,6 @@
   }
 }
 
-.c-country{
+.c-country {
   margin-top: 1.5rem;
 }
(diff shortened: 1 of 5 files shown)


src/app/campaign-details-card/campaign-details-card.component.scss
```diff
@@ -4,8 +4,18 @@
   @include make-card($box-shadow-hard);
 }
 
-.c-icon-lists {
-  @include card-icons;
+.c-icon-list {
+  list-style-type: none;
```
NoelLH, line 8: We should probably either try to align and keep using the mixin, or get rid of it if it is now only used in 1 place

src/app/campaign-details-card/campaign-details-card.component.scss
```diff
@@ -44,13 +57,14 @@
...
   color: $colour-teal;
+  margin-top: 0;
   a {
     text-decoration: none;
   }
 }
 
-.c-country {
-  margin-top: 0;
+.c-country{
```
NoelLH, line 68: Spacing inconsistent

NoelLH (changes requested):

NoelLH (approved):

PR #369, opened 2020-11-16
DON-338 - fix conditions for Stripe card mount & donation create

NoelLH:
Also repopulate on resume, and don't explicitly validate, tipPercentage, to make form validation less brittle

### DIFF
--- a/src/app/donation-start/donation-start.component.ts
+++ b/src/app/donation-start/donation-start.component.ts
@@ -315,7 +315,10 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
     // Create a donation if coming from first step and not offering to resume
     // an existing donation.
     if (event.previouslySelectedStep.label === 'Your donation') {
-      if (this.previousDonation === undefined || this.previousDonation.status === 'Cancelled') {
+      if (
+        (this.previousDonation === undefined || this.previousDonation.status === 'Cancelled') &&
+        event.selectedStep.label !== 'Your donation' // Resets fire a 0 -> 0 index event.
+      ) {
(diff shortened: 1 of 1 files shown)


src/app/donation-start/donation-start.component.ts
```diff
@@ -312,21 +312,12 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
...
-
-    const invalidPreviousDonation = (this.previousDonation === undefined || this.previousDonation.status === 'Cancelled');
-
-    // Create a donation if user actively clicks 'next' from first step and
-    // the current and previous donations are invalid.
-    if (activelySelectedNext && invalidDonation && invalidPreviousDonation) {
-      this.createDonation();
+    // Create a donation if coming from first step and not offering to resume
+    // an existing donation.
+    if (event.previouslySelectedStep.label === 'Your donation') {
```
tbg-jasonfung, line 317: @NoelLH removing the `event.previouslySelectedStep.interacted === true` check here has caused donations that are cancelled through this modal:

image

to show this error:

image

because in `donation-start.component.getDialogResponseFn()` we reset the stepper and its causing it to create an invalid donation

NoelLH: OK – it sounds like this is a reason to validate the state of the local vars, not the specific UI interactions which is far too brittle @tbg-jasonfung 

I'm not clear on the order of steps for this scenario, could you list them please?

NoelLH: Don't worry, got it

tbg-jasonfung (approved):

PR #371, opened 2020-11-16
DON-324 - fix search navigation + refactor moving Search to a service

NoelLH:


tbg-jasonfung:
Changes to address navigation history is good, however @NoelLH when conducting multiple search terms and then hitting back, the `Find a charity` input field doesn't update with the previous search term, to replicate,

1.) search for dogs
2.) search for person
3.) hit back on browser
4.) observe input field will still show `person`

tbg-jasonfung (approved):

PR #372, opened 2020-11-16
Increase donation input height

tbg-jasonfung:
- Adjust tipping input fields width


NoelLH (commented):
I thought we were going to make the main field taller instead of making this easier to miss?

tbg-jasonfung:
> I thought we were going to make the main field taller instead of making this easier to miss?

Oh yep we did @NoelLH

NoelLH (approved):

PR #383, opened 2020-11-24
DON-352 replace routelink with href to kill the state of the page

tbg-jasonfung:
- DON-352 replace routelink with href to kill the state of the page


NoelLH (changes requested):
This seems like a bit of an extreme change @tbg-jasonfung, do we really need to make the whole app load again?

We use `onSameUrlNavigation: 'reload'` in the routing file which should in theory mean we pick up route changes and can respond accordingly. So I think what's really wrong is maybe to do with getting the search service to respond in the way we expect at the same time such navigation happens. I think we could listen for certain routing events to handle this properly.

NoelLH (approved):

PR #391, opened 2020-12-08
INFRA-16 - avoid sending blank slug Campaign API requests

NoelLH:
Observed a handful of these last night. Suspect a rare loading race condition.


NoelLH:
Good to get through while CC20 open if possible @tbg-jasonfung

tbg-jasonfung (approved):

PR #393, opened 2020-12-08
DON-360 add max length attributes which correlate to those defined in SF

tbg-jasonfung:
- DON-360 add max length attributes which correlate to those defined in SF


NoelLH (commented):
Any reason not to use Angular validators @tbg-jasonfung? I think it's clearer what's happening if the data rules are in one place.

NoelLH (approved):

PR #395, opened 2020-12-08
DON-360 add validators in correct position + add validation messaging

tbg-jasonfung:
- DON-360 add validators in correct position + add validation messaging


src/app/donation-start/donation-start.component.ts
```diff
@@ -505,6 +505,30 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
     return this.donationForm.controls.amounts.get('tipAmount');
   }
 
+  get firstNameField() {
```
NoelLH, line 508: These seems like quite a lot of JS weight – could we not use the groups directly in the templates as with other similar validation messages?

NoelLH (approved):

PR #396, opened 2020-12-10
DON-362 handle 'preview' campaign status

tbg-jasonfung:
- DON-362 handle 'preview' campaign status

### DIFF
--- a/src/app/campaign-details/campaign-details.component.html
+++ b/src/app/campaign-details/campaign-details.component.html
@@ -34,12 +34,17 @@
       </button>
 
       <mat-card *ngIf="!donateEnabled" class="c-not-open">
-        <mat-card-title *ngIf="campaignInFuture">
+        <mat-card-title *ngIf="inPreview">
+          <mat-icon aria-hidden="false" aria-label="Upcoming campaign">update</mat-icon>
+          Campaign preview
+        </mat-card-title>
+
+        <mat-card-title *ngIf="campaignInFuture && !inPreview">
           <mat-icon aria-hidden="false" aria-label="Future campaign">update</mat-icon>
           Campaign opens soon!
         </mat-card-title>
 
-        <mat-card-title *ngIf="!campaignInFuture">
+        <mat-card-title *ngIf="!campaignInFuture && !inPreview">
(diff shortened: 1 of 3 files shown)


src/app/campaign.service.spec.ts
```diff
@@ -110,17 +110,17 @@ describe('CampaignService', () => {
     expect(CampaignService.isInFuture(campaign)).toBe(false);
   });
 
-  it ('should allow donation attempts to any campaign in active date range, even if Status gets stuck in Preview', () => {
+  it ('should not allow donation attempts to campaigns in Preview status', () => {
```
NoelLH, line 113: @tbg-jasonfung this seems problematic because the whole point of this test & logic was to make sure we weren't relying solely on the Status, which had a delay updating. By reverting that we will reinstate the bug we had before.

NoelLH (changes requested):

src/app/campaign.service.spec.ts
tbg-jasonfung: @NoelLH Yeah I thought that too, but I think that was when we were using `Campaign Status` rather than the refactored `Status` which is a much more reactive formula field - i've been testing flicking the status between `Preview` and `Active` and frontend is reacting immediately

tbg-jasonfung: Alternatively, we could handle regular campaigns that have not had the appropriate approvals with a separate status like `Pending Approval`

NoelLH: @tbg-jasonfung yeah – I woke up thinking about this! :/ And I don't think this can ever be reliable as things stand because we have to cache campaign data for up to a minute. So there is no way this can reliably flip for everyone at the right second. I definitely think a status that accurately captures these 2 different states is necessary, and that will probably mean no front end logic change except maybe listening for a different status. What we are currently calling _Pending_ is probably less confusing as something like _Scheduled_. So we would then have the understanding that Scheduled campaigns are about to launch if nothing changes, whereas Preview ones need some additional work, whether that's approval by us or by the charity or completion of required account info.

If it's simpler we could repurpose _Draft_, but I think the distinction between the reasons it's pending is probably useful and relied upon within Salesforce code too.

NoelLH (approved):

PR #397, opened 2020-12-14
Stop homepage redirect + reinstate explore in nav menu

tbg-jasonfung:
- Stop homepage redirect + reinstate explore in nav menu


src/app/main-menu/main-menu.component.html
```diff
@@ -1,11 +1,10 @@
 <mat-nav-list [class]="listClass">
-  <!-- TODO – reinstate after CC20 -->
-  <!-- <mat-list-item>
+  <mat-list-item>
```
NoelLH, line 2: Should be doing footer counterpart too

NoelLH (changes requested):

PR #409, opened 2021-01-12
Update top level libs + `npm audit fix`

NoelLH:


NoelLH:
It looks like the Karma / `test-ci` hang is related to https://github.com/angular/angular-cli/issues/19773

I've tried temporarily pinning the dev dependency mentioned to the previous version, with and without a CLI version downgrade, but it still seems not to work so I think there is another compatibility issue interacting with this.

The best option for now is probably to wait until tomorrow's release and update everything without pinning any old versions.

tbg-jasonfung (approved):

PR #441, opened 2021-04-23
DON-388 - support any donor country

NoelLH:
* expose a required billing country field, defaulting to GB, for Stripe
* use this in place of fixed 'GB' for data passed to Donate API & Stripe
* use a simplified, looser pattern requirement for billing postcode


alihejazi (approved):
Looks good.

PR #462, opened 2021-05-26
DON-386 - first pass multicurrency landing page

NoelLH:
No real attention to layout yet, but this supports
* defaulting to the CDN-detected country
* letting donors choose this or another
* loading the potential child campaigns invisibly
* jumping to DonationStartComponent for the appropriate campaign


src/app/multicurrency-landing/multicurrency-landing.component.html
```diff
@@ -0,0 +1,64 @@
+<main>
+  <!-- <app-ticker [campaign]="campaign" [fund]="fund"></app-ticker> -->
+  <div class="hero">
+
+  </div>
+
+  <section class="cta">
+    <div class="b-container">
+      <h1>Vaccines for everyone, everywhere</h1>
+      <div class="multicurrency-location-pick-wrapper" style="padding: 0 0 150px 0;">
```
NoelLH, line 10: I would avoid inline `style`, especially when mixed with SCSS with a selector matching the same object.

src/app/multicurrency-location-pick/multicurrency-location-pick.component.scss
```diff
@@ -0,0 +1,26 @@
+$colour-multicurrency-brand: #ed6743;
+$colour-multicurrency-brand-darker: #b14324;
```
NoelLH, line 2: I think we should probably avoid adding new colours linked to the brand unless there's a reference using these @alihejazi? 

https://gogiveone.org/ does have a slightly different hover colour but it uses #f24d2d

src/assets/svgs/facebook-square-brands.svg
```diff
@@ -0,0 +1 @@
+<svg aria-hidden="true" focusable="false" data-prefix="fab" data-icon="facebook-square" class="svg-inline--fa fa-facebook-square fa-w-14" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><path fill="#ed6743" d="M400 32H48A48 48 0 0 0 0 80v352a48 48 0 0 0 48 48h137.25V327.69h-63V256h63v-54.64c0-62.15 37-96.48 93.67-96.48 27.14 0 55.52 4.84 55.52 4.84v61h-31.27c-30.81 0-40.42 19.12-40.42 38.73V256h68.78l-11 71.69h-57.78V480H400a48 48 0 0 0 48-48V80a48 48 0 0 0-48-48z"></path></svg>
```
NoelLH, line 1: @alihejazi  see `campaign-groups.service.ts` for somewhere we import icons without having to check in our own copies.

src/app/multicurrency-landing/multicurrency-landing.component.html
alihejazi: I've added the inline css here to override the default styling for the `multicurrency-location-pick-wrapper.` Because it appears in two places, I have the default padding style defined in scss, but for the first occurrence of `multicurrency-location-pick-wrapper` it needs some extra padding, so I've overridden like this instead of making a second, conflicting class name.

src/app/multicurrency-location-pick/multicurrency-location-pick.component.scss
alihejazi: Cool, I've removed the darker colour.

src/assets/svgs/facebook-square-brands.svg
alihejazi: Done, now updated. Thanks for the pointer; that was helpful.

alihejazi (approved):

src/app/multicurrency-landing/multicurrency-landing.component.html
NoelLH: @alihejazi this sounds like it should just be two different classes

PR #503, opened 2021-08-13
DON-418 Fixed ticker alignment on tablets

alihejazi:


src/app/ticker/ticker.component.scss
```diff
@@ -34,10 +37,6 @@
   width: 100%;
   position: relative;
   overflow: hidden;
-  top: 1px;
```
NoelLH, line 37: @alihejazi removing `top` completely on a `relative` element seems like a risk for regressions in some browsers. I think the original design also had the smaller ticker pieces vertically aligned in the middle of the primary value.

PR #510, opened 2021-09-20
BG2-1855 hide donate button for draft campaigns

alihejazi:


src/app/campaign-details/campaign-details.component.html
```diff
@@ -37,17 +37,22 @@ <h1 class="c-header__campaign-title b-rh-1 b-bold">{{ campaign.title }}</h1>
           Campaign preview
         </mat-card-title>
 
-        <mat-card-title *ngIf="campaignInFuture && !isPending">
+        <mat-card-title *ngIf="isDraft">
+          <mat-icon aria-hidden="false" aria-label="Draft campaign">preview</mat-icon>
+          Campaign draft
```
NoelLH, line 42: I don't think we need to distinguish the scenarios for donors – if they've got to the page we can just provisionally say it 'opens soon'

src/app/campaign.model.ts
```diff
@@ -37,7 +37,7 @@ export class Campaign {
     public quotes: Array<{person: string, quote: string}>,
     public solution: string,
     public startDate: Date,
-    public status: 'Active' | 'Expired' | 'Preview' | 'Pending',
+    public status: 'Active' | 'Draft' | 'Expired' | 'Preview' | 'Pending',
```
NoelLH, line 40: I think we can/should avoid adding a whole new logic branch separate from 'Preview' – will add more on the other PR

NoelLH (changes requested):

PR #529, opened 2021-10-12
DON-444 - update suggested tip for < £100 donations

NoelLH:


src/app/donation-start/donation-start.component.ts
```diff
@@ -1100,6 +1100,8 @@ export class DonationStartComponent implements AfterContentChecked, OnDestroy, O
             newDefault = 7.5;
           } else if (donationAmount >= 300) {
             newDefault = 10;
+          } else if (donationAmount >= 100) {
```
alihejazi, line 1103: @NoelLH did you mean `<=` instead of `>=`?

PR #561, opened 2021-11-02
Fix GG1 asset paths breaking SSR loads

NoelLH:
@alihejazi I noticed Staging GG1 glitching on server load (big blank space) and looked at the server logs. This PR updates the asset paths refs to use the same approach as the rest of the app, which should make asset path refs work in both the Express server / dynamic ECS version of the app, and the client-side bundle.


NoelLH:
Screenshot 2021-11-02 at 17 21 07
Screenshot 2021-11-02 at 17 22 27

alihejazi (approved):

PR #563, opened 2021-11-05
DON-442 footer clean-up, tiny tweak to DON-441 menu

alihejazi:


src/app/main-menu/main-menu.component.scss
```diff
@@ -10,7 +10,7 @@
 .mat-list-item {
   // Works whether rendering as `block` or `inline-block`. Must override Material base.
   width: auto !important;
-  padding: 0 1rem !important;
```
alihejazi, line 13: @NoelLH I added this tweak to the menu because between screen widths of 960 and 1000, the menu was slightly too wide for the page, giving a vertical gap from the top to the bottom of the page. If you want to see, you can use the dev tools to view the current staging site as it is in a width between 960 and 1000 and you'll see what I mean.

alihejazi: It's been updated to `padding: 0 0.75rem !important; `

NoelLH (approved):

PR #589, opened 2022-01-08
DON-480 - WIP hCaptcha demo

NoelLH:
Using "invisible [until something is to be verified] mode" and free
publisher service


NoelLH:
See https://github.com/thebiggive/donate-frontend/pull/591

PR #590, opened 2022-01-10
DON-474 add Experian logo, make social share icons work

alihejazi:


src/app/multicurrency-landing/multicurrency-landing.component.html
```diff
@@ -55,9 +49,9 @@ <h2>A global solution to a local problem</h2>
     <div class="b-container">
       <h2>Share this</h2>
       <div class="socials-wrapper">
-        <fa-icon [icon]="faFacebookSquare" class="social-icon"></fa-icon>
-        <fa-icon [icon]="faTwitter" class="social-icon"></fa-icon>
-        <fa-icon [icon]="faInstagram" class="social-icon"></fa-icon>
+        <a href="http://www.facebook.com/share.php?u=donate.thebiggive.org.uk/gogiveone" target="blank"><fa-icon [icon]="faFacebookSquare" class="social-icon"></fa-icon></a>
```
NoelLH, line 52: @alihejazi  did you check the counterparts in `DonationCompleteComponent`? I think they are better in a few ways.

* What about protocols in the URL?
* Do we want to define share copy for platforms supporting it?
* Why are we hard-coding this when we could make it work on `thebiggive.com` (which this campaign will use!) where appropriate, as well as facilitating more realistic QA?

In this case I suggest we also check in with the team about `?utm` parameters – have we discussed with WHO how they will track traffic coming in to the page, and would it be appropriate to have a campaign parameter for these shares in line with their conventions?

NoelLH (changes requested):
@alihejazi Comment inline on the share links.

We also said we were going to load the sponsor logo(s) dynamically as on the Jira issue. Although having thought about this more I suggest we use a Salesforce file with a public link so that it's easy for anyone in the team to publish a new version without needing AWS access. Either way the asset shouldn't be hard-coded in this repo.

src/environments/environment.production.ts
```diff
@@ -11,6 +11,9 @@ export const environment: Environment = {
   getSiteControlId: '97792',
   googleAnalyticsId: 'UA-2979952-1',
   googleOptimizeId: 'OPT-W78W6BT',
+  gg1FunderLogos: [
+    'https://thebiggive--c.visualforce.com/resource/1641902289000/imageGG1FunderExperian',
```
NoelLH, line 15: @alihejazi I was anticipating that this would just be one composite image so we could avoid new deploys. So I'm not sure we really need the array config or that the file name makes sense.

NoelLH (commented):
Looking good mostly @alihejazi. I'm not sure about the logo iterator/ config but I think otherwise this should be good to go.

src/environments/environment.production.ts
alihejazi: As discussed with Noel via DMs, we canceled the approach of using public URLs because:
1. Static resources aren't available from outside of Salesforce.
2. Public links of image files can't be used in image tags; they don't work because the link runs a form first which then displays a preview of the image. If you refresh the page after loading the public link, you'll see the browser says 'confirm form resubmission?', highlighting what was loaded is a form. The alternative is to right-click on the image that shows after opening that link, and get that url instead, but that url constrains the images to a specific content version, which is no good.

NoelLH (approved):

PR #595, opened 2022-01-17
DON-474 GG1 tweaks

alihejazi:


src/app/multicurrency-landing/multicurrency-landing.component.ts
```diff
@@ -15,7 +15,8 @@ export class MulticurrencyLandingComponent implements OnInit {
   faTwitter = faTwitter;
   faLinkedin = faLinkedin;
   shareUrl = environment.donateGlobalUriPrefix + '/gogiveone';
-  prefilledText = 'Go Give One!';
+  // The '%23' is an encoded '#'. Using '#' or '\#' did not work. Ticket DON-474.
+  prefilledText = 'Donate to the %23GoGiveOne campaign to help COVID-19 vaccines reach the rest of the world, prioritizing those who need them the most in countries that cannot afford them! Click below to donate today!';
```
NoelLH, line 19: @alihejazi I'd suggest just using a string literal and `encodeURIComponent()` as in the donation complete counterpart. This saves needing the manual encoding and also ensures that everything else in the string (e.g. spaces) is not unnecessarily brittle and at risk of breaking on certain browsers + social platforms.

NoelLH (changes requested):
All looks good apart from the string encoding point @alihejazi

NoelLH (approved):

PR #596, opened 2022-01-18
DON-480 - prepare captcha code consistently

NoelLH:


src/app/donation-start/donation-start.component.ts
```diff
@@ -688,7 +688,7 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
 
   captchaReturn(captchaResponse: string) {
     this.captchaCode = captchaResponse;
```
alihejazi, line 690: Might be helpful to sanity check that `captchaResponse` is not falsey in case a dev makes a change in the future which breaks this functionality without realising.

alihejazi, line 690: @NoelLH Might be helpful to sanity check that `captchaResponse` is not falsey in case a dev makes a change in the future which breaks this functionality without realising.

alihejazi (approved):

src/app/donation-start/donation-start.component.ts
NoelLH: @alihejazi Surely we would notice that in testing as it would break all donation creates? If it's not given a string it will also break at type level before the method is even entered, so this check would never be reached except specifically in the case of a blank string invocation, which would be a pretty surprising error.

PR #600, opened 2022-01-24
Update Puppeteer + Chromium + docs

NoelLH:
* Latest `puppeteer` major version includes fixing an upstream library
security issue
* Supported Chromium is 98 – revision details taken from https://github.com/puppeteer/puppeteer/commit/a55c86fac504b5e89ba23735fb3a1b1d54a4e1e5
* Readme now better explains the status of the Chromium version
pinning strategy / workarounds


NoelLH:
cc @alihejazi

PR #619, opened 2022-03-10
DON-493 post code retries allowed

alihejazi:


src/app/donation-start/donation-start.component.ts
```diff
@@ -750,6 +757,19 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
...
+  onBillingPostCodeChanged(event: Event) {
+    if (this.stripeResponseErrorCode === 'incorrect_zip') {
+      this.stripeError = undefined;
+      this.stripeResponseErrorCode = undefined;
+    }
+
+    if (this.donation) {
+      // Update the donation billingPostalAddress to ensure re-attempt of Stripe payment has the new post code
+      // See ticket DON-493 for more information.
+      this.donation.billingPostalAddress = (<HTMLInputElement>event.target).value;
```
NoelLH, line 769: This is a bit odd @alihejazi.

Firstly we already have this data bound to nice Angular-known form group properties as used throughout this file – we shouldn't have to resort to a new way of reading it that bypasses Angular Forms.

Secondly, why are we having to do this again just for this field? I think we should go back and work out why `submit()` is not behaving here when it seemingly works for all other data. In particular, I suggest having a look at the condition `... !this.donation?.billingPostalAddress && ...`

This part of the checks was first added on https://github.com/thebiggive/donate-frontend/commit/9db9dbc7a6a4ebf65075d72ab8492c56560259c7 but looks overzealous now – it means we don't react to new billing info at all if we had a billing postal address already. This is tied to avoiding regression in Apple Pay so we should be careful and think about whether in PRB use cases we need a new check to replace this one, but I think it's safe to say we should rationalise the current one.

We then won't need a separate, additional way that we patch `this.donation`. We _might_ not even need the `(change)` event – we can maybe just clear `stripeResponseErrorCode` at the start of `submit()`?

NoelLH (changes requested):
Investigation and approach looks good so far @alihejazi but I've highlighted a couple of things inline. And ideally I think a Validator would still help surface what the error is faster and more accessibly when billing postcodes are wrong.

src/app/donation-start/donation-start.component.ts
```diff
@@ -203,7 +205,7 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
           Validators.pattern(emailRegex),
         ]],
         billingCountry: [this.defaultCountryCode], // See setConditionalValidators().
-        billingPostcode: [null],  // See setConditionalValidators().
+        billingPostcode: [null, [Validators.required, ValidateBillingPostCode(this.stripeResponseErrorCode)]],  // See setConditionalValidators().
```
NoelLH, line 208: @alihejazi I think the fundamental problem here is that you've made a validator-creator, a "thunk", but only the function it returns is relevant after this point. Angular needs a simple validator fn and that's what it's got, which means no further evaluation is ever going to happen. Only the value of `this.stripeResponseErrorCode` _when the form was first created_ can ever be used, so the approach can't work.

It also feels like there is too much Stripe detail inside the validator here anyway – I reckon all we really need is a dumb validator which _always_ says something's wrong when attached.

So with this in mind I would suggest we:
1. simplify the validator to be a tiny fn that always returns an error; and
2. keep the logic to attach or remove it in `DonationStartComponent`, following the logic of or expanding `setConditionalValidators()` to make the validator apply only when we know from Stripe that there is a postcode matching error from the postcode in combination with the given payment card.

src/app/donation-start/donation-start.component.ts
```diff
@@ -608,13 +608,8 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
         this.stripeError = `Payment processing failed: ${result.error.message}`;
         this.stripeResponseErrorCode = result.error.code;
         if (this.isBillingPostCodeInvalid()) {
-          console.log(this.paymentGroup);
+          this.paymentGroup.controls.billingPostcode.setValidators([ValidateBillingPostCode]);
```
NoelLH, line 611: This is going to remove the required and pattern validators which are added when in Stripe credit card mode as it sets the full array to just one.

Note that these must remain dynamic because the field is not applicable when using Payment Request Buttons so the validators are sometimes set back to `[]`.

Similarly, we must have logic which removes this validator again when the postcode or card number is changed, otherwise it will continue to block submitting the form even after a correction.

src/app/donation-start/donation-start.component.html
```diff
@@ -264,8 +264,9 @@ <h1 class="b-rh-2 b-bold">{{ campaign.title }}</h1>
 
               <mat-form-field>
                 <mat-label for="billingPostcode">Billing postcode</mat-label>
-                <input formControlName="billingPostcode" id="billingPostcode" matInput>
+                <input formControlName="billingPostcode" id="billingPostcode" matInput (change)="onBillingPostCodeChanged($event)">
               </mat-form-field>
+              <div class="error" *ngIf="paymentGroup.get('billingPostcode')?.errors?.invalidBillingPostCode">Post code doesn't match the one used for your card.</div>
```
NoelLH, line 269: I still don't think we need a distinct message here. The Stripe error copy is already surfaced and has been used with donors without issue, whereas I think this wording and having 2x errors shown could both be confusing.

NoelLH (changes requested):

src/app/donation-start/donation-start.component.ts
```diff
@@ -750,6 +770,32 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
...
+    // updates their post code, pressing the 'Next' button will keep them
+    // where they are and not proceed them to the next step, because the
+    // next() method calls goToFirstVisibleError().
+    if (this.isBillingPostCodeInvalid()) {
+      this.stripeError = undefined;
+      this.stripeResponseErrorCode = undefined;
+
+      // Remove the ValidateBillingPostCode custom validator so billing postcode
+      // doesn't show as invalid after a change
+      this.paymentGroup.controls.billingPostcode.setValidators([
```
NoelLH, line 787: @alihejazi  could call `addStripeCardBillingValidators()` here so we're not maintaining it in more places than needed? I believe we're already relying on this to clear your extra validator on card # changes so this also makes the implementation self-consistent for both types of correction.

NoelLH (approved):

PR #669, opened 2022-09-13
DON-271 – set up Identity Person records during donation; offer to set a password afterwards

NoelLH:
When feature flag is enabled.

### DIFF
--- a/src/app/donation-complete/donation-complete.component.ts
+++ b/src/app/donation-complete/donation-complete.component.ts
@@ -1,4 +1,5 @@
 import { Component } from '@angular/core';
+import { MatDialog } from '@angular/material/dialog';
 import { ActivatedRoute } from '@angular/router';
 import { environment } from '../../environments/environment';
 
@@ -6,8 +7,11 @@ import { AnalyticsService } from '../analytics.service';
 import { Campaign } from '../campaign.model';
 import { CampaignService } from '../campaign.service';
 import { Donation } from '../donation.model';
+import { DonationCompleteSetPasswordDialogComponent } from './donation-complete-set-password-dialog.component';
 import { DonationService } from '../donation.service';
+import { IdentityService } from '../identity.service';
 import { PageMetaService } from '../page-meta.service';
+import { Person } from '../person.model';
 
 @Component({
   selector: 'app-donation-complete',
@@ -21,6 +25,7 @@ export class DonationCompleteComponent {
   public donation: Donation;
   public giftAidAmount: number;
   public noAccess = false;
+  offerToSetPassword = false;
   public prefilledText: string;
   public timedOut = false;
   public totalValue: number;
@@ -28,13 +33,16 @@ export class DonationCompleteComponent {
 
   private donationId: string;
   private maxTries = 5;
+  private person?: Person;
   private retryInterval = 2; // In seconds
   private tries = 0;
 
   constructor(
     private analyticsService: AnalyticsService,
     private campaignService: CampaignService,
+    public dialog: MatDialog,
     private donationService: DonationService,
+    private identityService: IdentityService,
     private pageMeta: PageMetaService,
     private route: ActivatedRoute,
   ) {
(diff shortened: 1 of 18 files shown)


src/app/donation-complete/donation-complete.component.ts
```diff
@@ -21,20 +25,24 @@ export class DonationCompleteComponent {
   public donation: Donation;
   public giftAidAmount: number;
   public noAccess = false;
+  offerToSetPassword = false;
```
alihejazi, line 28: @NoelLH Is there a reason you've omitted the access modifier here?

src/app/person.model.ts
```diff
@@ -0,0 +1,44 @@
...
+     * These 3 expected if donor's claiming Gift Aid.
+     */
+    home_address_line_1?: string;
+    home_postcode?: string;
+    home_country_code?: string;
+
+    /**
+     * Set on update if the donor opts to make a resuable account.
+     */
+    raw_password?: string;
```
alihejazi, line 29: @NoelLH Isn't the password hashed when stored in the ID DB? If so, I would advise renaming this variable to `hashed_password`.

src/app/identity.service.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const TBG_DONATE_ID_STORAGE = new InjectionToken<StorageService>('TBG_DONATE_ID_STORAGE');
+
+@Injectable({
+  providedIn: 'root',
+})
+export class IdentityService {
+  private readonly apiPath = '/people';
+  // Key is per-domain/env. For now we simply store a single JWT (or none).
+  private readonly storageKey = `${environment.identityApiPrefix}/v1/jwt`;
```
alihejazi, line 19: @NoelLH The `/v1` here is extra, I think. You've already included `v1` as part of the environment variable definition. See the environment files.

src/app/identity.service.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  create(person: Person): Observable<Person> {
+    return this.http.post<Person>(
+      `${environment.identityApiPrefix}${this.apiPath}`,
+      person);
+  }
+
+  update(person: Person): Observable<Person> {
+    return this.http.put<Person>(
+      `${environment.donationsApiPrefix}${this.apiPath}/${person.id}`,
```
alihejazi, line 36: @NoelLH You should be calling out to the Identity API here, not MatchBot

src/app/donation-complete/donation-complete.component.ts
NoelLH: @alihejazi Yes, `public` is the default and I have worked with TypeScript (and I think Angular) style guides before that explicitly suggest it is good practice not to include it for readability. Angular's docs' own examples also leave it out when demonstrating properties to be shared with templates. So I tend these days to phase it out as I go. I don't think it's worth refactoring the whole app but open to suggestions.

src/app/person.model.ts
NoelLH: @alihejazi This wouldn't be correct because the password sent from the client is only ever the raw password. Additionally, the API will never returned the hashed password: it is used only for server-side validation of login attempts.

src/app/identity.service.ts
NoelLH: @alihejazi This key is not a URL, but an identifier so we can tell with no risk of clashes which local data works with which version of the frontend. It's desirable to have "v1" to designate that this is the first way of using local storage because we should be able to change the client implementation without publishing a new API: the two are independent.

I think this syntax is how Donate already works which the other storage example, so I think that making a different structure (e.g. dot separators) at this point is probably more confusing than just sticking with this.

src/app/identity.service.ts
NoelLH: Good catch!

src/app/identity.service.ts
```diff
@@ -0,0 +1,75 @@
...
+    return this.http.put<Person>(
+      `${environment.identityApiPrefix}${this.apiPath}/${person.id}`,
+      person,
+      this.getAuthHttpOptions(person),
+    );
+  }
+
+  getIdAndJWT(): { id: string, jwt: string } | undefined {
+    if (!this.storage.has(this.storageKey)) {
+      return undefined;
```
alihejazi, line 43: @NoelLH this looks redundant. See next comment.

src/app/identity.service.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    return this.storage.get(this.storageKey) || undefined;
+  }
+
+  getJWT(): string | undefined {
+    return this.getIdAndJWT()?.jwt || undefined;
```
alihejazi, line 50: `|| undefined` here seems like unreachable code if `this.getIdAndJWT()?.jwt` can already return `undefined`

PR #672, opened 2022-09-27
DON-529 – auto-expire old ID JWTs

NoelLH:
Expiry is part of `getIdAndJWT()` which existing token/person checks already generally depend upon, so donors with old tokens should be treated like those who never had them transparently, as well as the old tokens being automatically deleted from `localStorage`.


src/app/identity.service.ts
```diff
@@ -62,29 +62,44 @@ export class IdentityService {
...
   getIdAndJWT(): { id: string, jwt: string } | undefined {
-    return this.storage.get(this.storageKey) || undefined;
+    const idAndJwt = this.storage.get(this.storageKey) || undefined;
+
+    if (idAndJwt === undefined) {
+      return undefined;
+    }
+
+    const data = this.getTokenPayload(idAndJwt?.jwt);
+    if (data.exp as number < Math.floor(Date.now() / 1000)) {
```
alihejazi, line 72: Shouldn't this be `>`? As in tokens expire _after_ their expiry date is > now?

src/app/identity.service.ts
```diff
@@ -62,29 +62,44 @@ export class IdentityService {
   }
 
   getIdAndJWT(): { id: string, jwt: string } | undefined {
-    return this.storage.get(this.storageKey) || undefined;
+    const idAndJwt = this.storage.get(this.storageKey) || undefined;
```
alihejazi, line 65: Is the `undefined` needed here? I presume it wouldn't if `this.storage.get(this.storageKey)` can return `null`.

src/app/identity.service.ts
alihejazi: @NoelLH actually never mind, I misunderstood this!

src/app/identity.service.ts
NoelLH: I think maybe not, let's try without

alihejazi (approved):

PR #683, opened 2022-10-11
Revert "DON-516 – drop temporary promoted [meta]campaign bits"

alihejazi:
This reverts commit 13ab2ae8cd77487e7b5fb795e51351de5aa1e9f2.

I also manually fixed some merge conflicts in app.module.ts.


NoelLH (changes requested):
This is rolling back stuff I don't think it should be (video autoplay back on?) and tests are failing

PR #684, opened 2022-10-11
DON-531 Revert "DON-516 – drop temporary promoted [meta]campaign bits"

alihejazi:


NoelLH (changes requested):
@alihejazi Please check the tests – they're still red. I think we need Staging data in place before we can release this without taking down Staging.

NoelLH (approved):

PR #689, opened 2022-10-13
DON-535 – captcha expiry fix tweaks

NoelLH:
* Stop doing stuff sooner if our captcha callback needs no action
* Protect against use of expired captcha codes

### DIFF
--- a/src/app/donation-complete/donation-complete.component.ts
+++ b/src/app/donation-complete/donation-complete.component.ts
@@ -98,6 +98,13 @@ export class DonationCompleteComponent {
   }
 
   loginCaptchaReturn(captchaResponse: string) {
+    if (captchaResponse === null) {
+      // This is expected after ~1 min when the code expires. At this point we should
+      // never be executing the login again because if the captcha was set up at all then
+      // we auto-logged-in with the password the donor just chose.
+      return;
+    }
+
     const credentials: Credentials = {
       email_address: this.donation.emailAddress as string,
       raw_password: this.person?.raw_password as string,
--- a/src/app/donation-start/donation-start-login-dialog.component.ts
+++ b/src/app/donation-start/donation-start-login-dialog.component.ts
@@ -40,6 +40,13 @@ export class DonationStartLoginDialogComponent implements OnInit {
   }
 
   captchaReturn(captchaResponse: string) {
+    if (captchaResponse === null) {
+      // We had a code but now don't, e.g. after expiry at 1 minute. In this case
+      // the trigger wasn't a login click so do nothing. A repeat login attempt will
+      // re-execute the captcha in `login()`.
+      return;
+    }
+
     this.loggingIn = true;
 
     const credentials: Credentials = {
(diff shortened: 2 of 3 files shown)


src/app/donation-start/donation-start.component.ts
```diff
@@ -729,13 +729,32 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
...
+      // the catcha to get a new one as needed instead.
+      this.captchaCode = undefined;
+      return;
+    }
+
     this.captchaCode = captchaResponse;
-    this.createDonationAndMaybePerson();
+
+    if (!this.donation) {
+      this.createDonationAndMaybePerson();
```
alihejazi, line 742: `this.createDonationAndMaybePerson()` is called from 4 different places in the same file, but you've only added the 'falsey' check for the two captchaReturn functions. Shouldn't the two other calls of said function also have the `if` check. And if so, shouldn't we just put the `if` check as the first thing inside the `createDonationAndMaybePerson()` function definition?

src/app/donation-complete/donation-complete.component.ts
```diff
@@ -98,6 +98,13 @@ export class DonationCompleteComponent {
   }
 
   loginCaptchaReturn(captchaResponse: string) {
+    if (captchaResponse === null) {
+      // This is expected after ~1 min when the code expires. At this point we should
+      // never be executing the login again because if the captcha was set up at all then
```
alihejazi, line 103: Same as https://github.com/thebiggive/donate-frontend/pull/689/files#r994633304

src/app/donation-start/donation-start-login-dialog.component.ts
```diff
@@ -40,6 +40,13 @@ export class DonationStartLoginDialogComponent implements OnInit {
   }
 
   captchaReturn(captchaResponse: string) {
+    if (captchaResponse === null) {
+      // We had a code but now don't, e.g. after expiry at 1 minute. In this case
+      // the trigger wasn't a login click so do nothing. A repeat login attempt will
```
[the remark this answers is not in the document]
NoelLH: @alihejazi No – if you look at the next line you'll see we use the result directly. There is no such property in this component as it's much simpler.

src/app/donation-start/donation-start.component.ts
NoelLH: @alihejazi I did think about this but as we have identified the specific cause of issues as the captcha resolve callback, I would rather change only that fn's behaviour for now. We don't know without a much more time consuming audit if there could be edge cases that tacitly rely on an existing `this.donation` being updated in-place.

src/app/donation-complete/donation-complete.component.ts
NoelLH: Same explanation: there is no local var and also no case where the captcha should execute a 2nd time in this component.

alihejazi (approved):

PR #693, opened 2022-10-17
DON-532 Introduce Buy Credits Page

alihejazi:


src/app/app-routing.module.ts
```diff
@@ -15,6 +16,10 @@ import { HomeComponent } from './home/home.component';
 import { MetaCampaignComponent } from './meta-campaign/meta-campaign.component';
 
 const routes: Routes = [
+  {
+    path: 'credits',
```
NoelLH, line 20: I think `buy-credits` would be more clear + future proof?

src/app/validators/credit-max.ts
```diff
@@ -0,0 +1,17 @@
+import { AbstractControl } from '@angular/forms';
+
+import { environment } from '../../environments/environment';
+
+export function ValidateCreditMax(control: AbstractControl) {
```
NoelLH, line 5: As both validators are near identical to existing code, it would probably make sense to DRY this up. You could introduce a simple wrapper "thunk" that returns the fn, perhaps also explicitly setting up `ValidatorFn` return values as covered in part of https://blog.angular-university.io/angular-custom-validators/

src/app/identity.service.ts
```diff
@@ -96,6 +96,15 @@ export class IdentityService {
     this.storage.set(this.storageKey, { id, jwt });
   }
 
+  getFundingInstructions(id: string, jwt: string): Observable<any> {
```
NoelLH, line 99: You can use the model instead of `any` now

src/app/buy-credits/buy-credits.component.scss
```diff
@@ -0,0 +1,61 @@
+@use '@angular/material' as mat;
+@import '../../abstract';
+
+#loggedInHeader {
+    background-color: green;
```
NoelLH, line 5: I don't think we should be introducing brand new colours! Can we just use existing visual elements so this is consistent with the rest of the app while we test ahead of the new design?

src/app/buy-credits/buy-credits.component.scss
```diff
@@ -0,0 +1,61 @@
...
+    margin-bottom: 0;
+}
+
+#giftAidOptIn {
+    margin-bottom: 20px;
+}
+
+.error {
+    color: mat.get-color-from-palette($donate-warn);
+    font-size: 1rem; // slightly smaller text
```
NoelLH, line 41: Why?

src/app/buy-credits/buy-credits.component.html
```diff
@@ -0,0 +1,150 @@
...
+                    <div>
+                        <button mat-button (click)="buyCredits()">Purchase Credits</button>
+                    </div>
+                </mat-step>
+            </mat-stepper>
+
+        </form>
+
+        <div *ngIf="isLoading">
+            <div id="loading-text">One moment...</div>
```
NoelLH, line 122: Not sure we need this _and_ a spinner? It's inconsistent with rest of the app

src/app/buy-credits/buy-credits.component.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  isLoggedIn: boolean = false;
+  isLoading: boolean = false;
+  isPurchaseComplete = false;
+  isOptedIntoGiftAid = false;
+  userFullName: string;
+  creditForm: FormGroup;
+  amountsGroup: FormGroup;
+  giftAidGroup: FormGroup;
+  isLinear = true;
```
NoelLH, line 32: Why's this variable?

src/app/buy-credits/buy-credits.component.ts
```diff
@@ -0,0 +1,245 @@
...
+    if (e.value === "yes") {
+      this.isOptedIntoGiftAid = true;
+    }
+    else {
+      this.isOptedIntoGiftAid = false;
+    }
+  }
+
+  showLoginDialog() {
+    const loginDialog = this.dialog.open(DonationStartLoginDialogComponent);
```
NoelLH, line 131: Good that we can reuse this, but I guess if doing so we should rename it + move it up in the app structure to avoid confusion

src/app/buy-credits/buy-credits.component.html
```diff
@@ -0,0 +1,150 @@
...
+                    <p>I confirm that I am a UK tax payer and I understand that if I pay less income tax and/or capital gains tax in the current tax year than the amount of
+                        Gift Aid claimed on all my donations it is my responsibility to pay the difference.</p>
+
+                    <p><strong>Note:</strong>This Gift Aid section is only for claiming Gift Aid on the tip to Big Give. If you wish to claim Gift Aid on the credit amount used for making donations, you may do so when donating
+                    to a charity's campaign.</p>
+
+                    <p><a href="">Find out more about Gift Aid </a></p>
+
+                    <mat-radio-group aria-label="Select an option" (change)="giftAidChoiceSelected($event)">
+                        <mat-radio-button value="yes" [checked]="false">Yes, I want to claim <strong>{{ giftAidAmount() | exactCurrency:'£' }}</strong> Gift Aid on my tip to Big Give, and any additional charitable donations via Big Give from the past 4 years and in the future.</mat-radio-button>
```
NoelLH, line 96: Is there a reason we're not binding directly to true/false as in the donate form?

I think that binding in TS with e.g. `this.giftAidGroup.get('giftAid')?.valueChanges.subscribe(...)` is also a bit safer, because it keeps things agnostic to the cause of a change: any future programmatic updates to the form data would be the same as clicks for example.

src/app/buy-credits/buy-credits.component.html
```diff
@@ -0,0 +1,150 @@
...
+        <div *ngIf="isLoggedIn && isPurchaseComplete">
+            <h2>Purchase complete 🎉🎊</h2>
+            <p>Please bank transfer us the credit amount of {{ creditAmountField?.value | exactCurrency: '£' }} using the following details (keep them safe!):</p>
+            <p><strong>Recipient</strong> {{ accountHolderName }}</p>
+            <p><strong>Sort Code:</strong> {{ sortCode }}</p>
+            <p><strong>Account Number:</strong> {{ accountNumber }}</p>
+
+            <hr>
+
+            <!-- should we keep this copy? -->
```
NoelLH, line 144: I would say not for MVP

src/app/buy-credits/buy-credits.component.html
```diff
@@ -0,0 +1,150 @@
...
+            <p>You must be logged in to purchase donation credits.</p>
+            <button
+                mat-raised-button
+                color="primary"
+                (click)="showLoginDialog()"
+            >Log in</button>
+        </div>
+
+        <div *ngIf="isLoggedIn && isPurchaseComplete">
+            <h2>Purchase complete 🎉🎊</h2>
```
NoelLH, line 136: I'm not sure about this heading – it implies we already have the money somehow.

src/app/buy-credits/buy-credits.component.html
```diff
@@ -0,0 +1,150 @@
...
+                    </div>
+                </mat-step>
+                <mat-step>
+                    <ng-template matStepLabel>Submission</ng-template>
+
+                    <p>By clicking on the 'Purchase Credits' button below, you legally agree to transfer the credit funds via bank transfer to the bank account and sort code
+                        displayed on the next screen. See our <a href="https://www.thebiggive.org.uk/s/terms-and-conditions">Terms and Conditions</a> for more information.
+                    </p>
+
+                    <p>We use <a href="https://stripe.com">Stripe</a> as our payments processor.</p>
```
NoelLH, line 111: New tab?

src/app/buy-credits/buy-credits.component.html
```diff
@@ -0,0 +1,150 @@
...
+                    <p *ngIf="
+                        creditAmountField?.value &&
+                        !creditAmountField?.errors &&
+                        !creditAmountField?.invalid &&
+                        !(customTip() && !customTipAmountField?.value) &&
+                        !customTipAmountField?.errors &&
+                        !customTipAmountField?.invalid
+                    "><strong>{{ calculatedTipAmount() | exactCurrency:'£' }}</strong> donation (tip) to support Big Give
+                        <span *ngIf="calculatedTipAmount() > 0">💗</span> <!-- display heart emoji if tip > £0 -->
+                        <span *ngIf="calculatedTipAmount() === 0 && customTip() && customTipAmountField?.value">😢</span> <!-- display sad emoji if custom tip amount set to £0 -->
```
NoelLH, line 60: I'm not sure it's a good idea to be hitting big corporate donors with a sad face emoji!

src/app/buy-credits/buy-credits.component.html
```diff
@@ -0,0 +1,150 @@
...
+        <form *ngIf="isLoggedIn && !isLoading && !isPurchaseComplete" [formGroup]="creditForm">
+            <mat-stepper orientation="vertical" [linear]="isLinear" #stepper>
+                <mat-step
+                    [completed]="false"
+                    formGroupName="amounts"
+                    [stepControl]="amountsGroup"
+                >
+                    <ng-template matStepLabel>Your Donation Credits</ng-template>
+                    <mat-form-field  class="form-field b-center size-lg" hideRequiredMarker>
+                        <mat-label>Credit amount (£)</mat-label>
```
NoelLH, line 17: Any reason not to make this a public class var? We pretty much eliminated currency hard-coding in templates across the whole app, at some length, so it seems a shame to bring it back.

src/app/buy-credits/buy-credits.component.ts
```diff
@@ -0,0 +1,245 @@
...
+    if (!unsanitisedCreditAmount) {
+      return 0;
+    }
+
+    if (this.customTip()) {
+      const unsanitisedCustomTipAmount = this.amountsGroup.value.customTipAmount;
+      if (unsanitisedCustomTipAmount) {
+        return this.sanitiseCurrency(unsanitisedCustomTipAmount)
+      }
+      else {
```
NoelLH, line 196: I think we can make this more readable by losing this "else", and the next one. `return` handles it and allows for less nesting + cognitive load.

NoelLH (commented):
Coming along well!

I've left some comments inline. We'll also need the test for the new component mounting to pass.

src/app/buy-credits/buy-credits.component.html
alihejazi: Wdym?

src/app/buy-credits/buy-credits.component.scss
alihejazi: So get rid of the logged in header altogether?

src/app/buy-credits/buy-credits.component.scss
alihejazi: It looks nicer imo, otherwise it's the same size as the normal text.

src/app/buy-credits/buy-credits.component.html
alihejazi: Oh nvm, I get you. Added target _blank

src/app/validators/credit-max.ts
alihejazi: Good you reminded me I actually don't need the `ValidateCreditMin` validator because I directly use `Validators.min(0)` instead. So I'll delete `ValidateCreditMin` and keep `ValidateCreditMax` as is.

src/app/buy-credits/buy-credits.component.scss
alihejazi: I think it's fine to keep it, looks good imo.

src/app/validators/credit-max.ts
alihejazi: Nice suggestion, done.

src/app/buy-credits/buy-credits.component.scss
NoelLH: Nice – yep I think giving it an element (until we have a new header component) makes sense – I just didn't want to introduce a mix and match of colour palettes that could muddy the rebrand launch.

PR #694, opened 2022-10-17
DON-535 reset captcha before re-executing

alihejazi:


NoelLH (commented):
@alihejazi Nice, this looks promising! I think the one remaining place we should probably be considering the same improvement for edge case handling is `DonationStartLoginDialogComponent.login()`, since this too is a user-invoked fn which could easily happen > 1 minute apart twice on a page load.

I believe that in `donation-complete` where the password set login success and autologin should only be able to occur once per pageload, it is safe not to add this.

NoelLH (approved):

PR #701, opened 2022-10-28
DON-533 – credit use QA fixes

NoelLH:
* Resolve JS crash from missing Payment Request Button element under some conditions
* Ensure amount is always loaded as an int, so that `===` checks work as intended
* Don't require billing address or country when using credit as no new

### DIFF
--- a/src/app/donation-start/donation-start.component.ts
+++ b/src/app/donation-start/donation-start.component.ts
@@ -385,6 +385,7 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
     this.personId = undefined;
     this.personIsLoginReady = false;
     this.stripeFirstSavedMethod = undefined;
+    this.stripePaymentMethodReady = false;
     this.donationForm.reset();
     this.identityService.clearJWT();
     this.idCaptcha.reset();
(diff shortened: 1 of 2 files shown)


src/app/donation-start/donation-start.component.ts
```diff
@@ -911,7 +912,11 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
       this.personIsLoginReady = true;
 
       if (environment.creditDonationsEnabled && person.cash_balance && person.cash_balance[this.campaign.currencyCode.toLowerCase()] > 0) {
-        this.creditPenceToUse = person.cash_balance[this.campaign.currencyCode.toLowerCase()];
+        this.creditPenceToUse = parseInt(
+          person.cash_balance[this.campaign.currencyCode.toLowerCase()].toString() as string,
```
alihejazi, line 916: @NoelLH why are you casting a number to a string and then back to a number?

NoelLH: @alihejazi `parseInt(...)` gives a type error given anything but a string.

alihejazi (approved):

PR #706, opened 2022-11-01
DON-590 Homepage Re-skinning (MVP)

alihejazi:


e2e/src/app.e2e-spec.ts
```diff
@@ -8,11 +8,6 @@ describe('Donate Frontend', () => {
     page = new AppPage();
   });
 
-  it('should display TBG logo', () => {
```
NoelLH, line 11: @alihejazi It seems like a mistake to remove the _only_ e2e check we have that a page renders. Could we introduce a similarly simple but different test?

PR #715, opened 2022-11-04
Bring back a simple e2e contents test

NoelLH:


NoelLH:
Tried numerous workarounds for shadow DOM access without getting one to work yet.

Dropping this in favour of migrating to Cypress and its `shadow()` reasonably soon, as I can't justify spending more time on tests we will replace v soon anyway.

PR #725, opened 2022-11-09
DON-587 Meta Totalizer / Ticker

alihejazi:


src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -173,6 +224,17 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
     return (childCampaign.amountRaised / childCampaign.target) * 100;
   }
 
+  formatAmount(amount: number) {
```
NoelLH, line 227: @alihejazi Is there a reason not to use one of the currency pipes we have everywhere else for consistency? I see we're already using pipes in TS elsewhere in this same update.

NoelLH: @alihejazi  I have tried to do this without us repeating core Angular stuff in https://github.com/thebiggive/donate-frontend/pull/733 – this is also necessary to avoid breaking non-GBP campaigns

PR #728, opened 2022-11-09
Switch remaining fonts to Euclid, including in Stripe.js elements

NoelLH:
And take old Maven Pro bits out of the build config


NoelLH:
@alihejazi I think we can consider merging this unless you can think of any problems?

I'm cautiously optimistic we may not need code to control font load waiting, especially as Material icons are being phased out, which should simplify render readiness logic and perhaps speed up first paint.

Stripe.js access to the font didn't work on local for me but I think it probably will on Staging where the hash output is used. I think we can live with this limitation for now, and I don't think the has should change unless the font file itself does.

PR #739, opened 2022-11-11
DON-607 – get latest ticker component lib version

NoelLH:


NoelLH:
Will be superseded by @alihejazi's imminent update

PR #740, opened 2022-11-13
DON-633 – fix context-sensitive totalizer items + main message

NoelLH:


src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -123,63 +115,18 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
     if (this.fundSlug) {
       fundKey = makeStateKey<Fund>(`fund-${this.fundSlug}`);
       this.fund = this.state.get<Fund | undefined>(fundKey, undefined);
+      this.setFundSpecificTickerParams();
     }
 
     if (!this.fund && this.fundSlug) {
```
alihejazi, line 121: @NoelLH this `if` could be simplified with the one above since they have a common condition. Only caveat is we'd need an internal `if` to check `if (!this.fund)`

src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -371,4 +318,75 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
...
+          figure: this.timeLeftPipe.transform(this.campaign.endDate),
+        },
+        {
+            label: 'Match Funds Remaining',
+            figure: this.currencyPipe.transform(this.campaign.matchFundsRemaining, this.campaign.currencyCode, 'symbol', '1.0-0') as string,
+        },
+      ]);
+    }
+
+    if (!campaignOpen) {
```
alihejazi, line 354: @NoelLH Probs better to use `else`

src/app/meta-campaign/meta-campaign.component.ts
NoelLH: @alihejazi I don't think this really makes sense.

1. We'd add a level of conditional nesting so I think it makes the code harder to follow
2. We'd still need to call the method you mention twice because 1 of the 2 cases is in an async callback from an Observable

alihejazi (approved):

PR #765, opened 2022-11-21
DON-660 DON-661 DON-662

alihejazi:


src/app/campaign.service.ts
```diff
@@ -28,7 +28,7 @@ export class CampaignService {
       return false;
     }
 
-    if (campaign.status === 'Active') {
+    if (campaign.status === 'Active' && new Date(campaign.endDate) >= new Date()) {
```
bdsl, line 31: I don't think github will allow me to make a code suggestion that long, but might be worth simplifying slightly by extracting a variable for the current date instead of instantiating in three places, i.e.:


```ts
    const now = new Date();
    
    if (campaign.status === 'Active' && new Date(campaign.endDate) >= now) {
      return true;
    }

    if (new Date(campaign.startDate) > now) {
      return false;
    }

    return (new Date(campaign.endDate) >= now);
  }
```

PR #766, opened 2022-11-21
DON-660 DON-661 DON-662 use button event emit for url route -> stops 'closed/launches' message being secretly clickable!

alihejazi:


alihejazi:
Merging straight in without waiting for build as it's a simple change

PR #776, opened 2022-11-23
DON-561: Replace youtube video on homepage

bdsl:
Screenshot from 2022-11-23 10-11-10


bdsl:
I'm not 100% sure if we wanted to replace this video or if we wanted to add an additional video the homepage - waiting to hear via JIRA comments before merging.

alihejazi (approved):

PR #787, opened 2022-11-25
Update `develop` with `main`

NoelLH:
Fix GitHub complaining about out-of-dateness and refusing to update in the UI due to non-reviewed commits.


NoelLH:
@bdsl I think if we're going to have any system where we do force pushes, we need to have a solid idea of how we're going to avoid stuff like this. It's quite a bit of extra work and waiting for builds when we have urgent deploys.

bdsl:
@NoelLH I'm not sure what this is - probably best to chat more after the CC as you said.

PR #795, opened 2022-11-28
DON-685 – give totalizer live timing info

NoelLH:
Support a countdown while the page is open, as before. Updates are called client side once per second so should have similar performance impact to when we used `TimeLeftPipe` in the template directly.

### DIFF
--- a/src/app/meta-campaign/meta-campaign.component.ts
+++ b/src/app/meta-campaign/meta-campaign.component.ts
@@ -55,6 +55,7 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
   // scrolling back to the previous metacampaign position, so to be safe we use the highest
   // scroll Y we've seen without intervention, plus a ~25% buffer.
   private smallestSignificantScrollPx = 250;
+  private tickerUpdateTimer: any;
 
   private readonly recentChildrenKey = `${environment.donateUriPrefix}/children/v2`; // Key is per-domain/env
   private readonly recentChildrenMaxMinutes = 10; // Maximum time in mins we'll keep using saved child campaigns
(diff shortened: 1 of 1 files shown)


src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -430,6 +435,18 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
...
+
+    if (this.tickerUpdateTimer) {
+      clearTimeout(this.tickerUpdateTimer);
+    }
+
+    // Load data just once, but refresh e.g. time left to launch summary once
+    // per second.
+    if (isPlatformBrowser(this.platformId)) {
+      this.tickerUpdateTimer = setTimeout(() => {
+        this.setTickerParams()
```
alihejazi, line 447: @NoelLH Looks good but I have a _slight_ concern that recalling _all_ of `this.setTickerParams()` every minute might be computing a lot of unnecessary stuff. In other words the only thing we need to recompute is the time left.

I also have another slight concern that this will happen every second even when it's not really noticeable, i.e. after the campaign launches when the time left will change to days, as opposed to mins and seconds, and this case covers most of the campaign time.

NoelLH: @alihejazi There is very minimal logic in the rest of the method. Compared to all the other computation I expect it to be negligible.

Calling it every 1s is no different from the old approach, or from everywhere TimeLeftPipe is still used in templates.

alihejazi: Ok let's try this out then

alihejazi (approved):

PR #805, opened 2022-11-29
DON-691 – QA map init fix

NoelLH:


bdsl (approved):
LGTM

bdsl:
Setting `strictPropertyInitialization=true` in tsconfig would prevent the error fixed here, but then it also makes TS complain about quite a lot of code we have now that may be fine.

PR #806, opened 2022-11-29
DON-690 – fix float value errors in some PRB cases

NoelLH:
We have to do a slightly weird type dance because TS needs its `parseInt()` inputs to be strings.

This should stop Stripe PRB logic crashing because it's given 8099.9999999 pence or similar.


bdsl (approved):
LGTM

PR #816, opened 2022-12-01
DON-698 – simplify and loosen validation of email addresses' domain part

NoelLH:
Notably, permit subdomains. Also allows a couple of known odd things like "." straight after "@" in the interest of speed + simplicity.

### DIFF
--- a/src/app/donation-start/donation-start.component.ts
+++ b/src/app/donation-start/donation-start.component.ts
@@ -129,7 +129,12 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
 
   private initialTipSuggestedPercentage = 15;
 
-  private emailRegExp : RegExp = /^(?=.{1,254}$)(?=.{1,64}@)[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+)*@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(\.[a-zA-Z]{2,})+$/;
+  // Validators.email regexp rejects most invalid emails but has a few edge-cases slip through.
+  // For example, it allows emails ending with numbers like hello@thebiggive.org.uk.123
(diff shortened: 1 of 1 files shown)


src/app/donation-start/donation-start.component.ts
```diff
@@ -129,7 +129,12 @@ export class DonationStartComponent implements AfterContentChecked, AfterContent
 
   private initialTipSuggestedPercentage = 15;
 
-  private emailRegExp : RegExp = /^(?=.{1,254}$)(?=.{1,64}@)[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+)*@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(\.[a-zA-Z]{2,})+$/;
+  // Validators.email regexp rejects most invalid emails but has a few edge-cases slip through.
+  // For example, it allows emails ending with numbers like hello@thebiggive.org.uk.123
+  // We needed tighter validation, so have adapted the Angular pattern iteratively including
+  // some simplification. We needed to loosen part of it in Dec '22 because subdomains
+  // weren't properly supported in the previous version.
+  private emailRegExp : RegExp = /^(?=.{1,254}$)(?=.{1,64}@)[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+)*@[.a-zA-Z0-9-]{0,180}\.[a-zA-Z]{2,}$/;
```
bdsl, line 137: It's almost impossible to read this regex IMHO, but I've used https://regex101.com/ to try it with a few examples, and it seems to be good. Let's get this out, later it would be good to see if we can either use something off the shelf so we don't have to deal with this, or if we can't do that break up the code to make it more readable and do a unit test with examples of valid and invalid addresses.

bdsl (approved):

alihejazi (approved):

PR #821, opened 2022-12-03
DON-677 – fix multiple location handling in campaign details "branded image"

NoelLH:


alihejazi:
Tested locally to confirm it works, and it does.

alihejazi (approved):

PR #828, opened 2022-12-05
Fix double back link in "?fromFund" view

NoelLH:
i.e. this, as highlighted in a Google Search Console warning today

Screenshot 2022-12-05 at 07 21 49


bdsl (approved):
LGTM

PR #830, opened 2022-12-05
Fix the hour in various `getRelevantDateAsStr()`s

NoelLH:
As discussed, I think shortly after CC22 we should refactor this so we don't have so much extra logic in `<biggive-button/>`. But for now, we should sort out the existing prop approach so that close times are factually correct.


bdsl (approved):
LGTM, changing from 12 hour clock hour to 24 hour clock hour.

PR #832, opened 2022-12-06
DON-687 – fix `<biggive-button/>` scope creep / coupling to campaign contexts

NoelLH:


src/app/campaign-details/campaign-details.component.html
```diff
@@ -246,11 +244,7 @@ <h4 class="b-semi-bold b-mb-0">{{ update.modifiedDate | date : 'longDate' }}</h4
             size="medium"
             rounded="false"
             [url]="'/donate/' + campaign.id"
-            [isFutureCampaign]="campaignInFuture"
-            [isPastCampaign]="campaignInPast"
-            [datetime]="getRelevantDateAsStr(campaign)"
-            >
-          </biggive-button>
+          ></biggive-button>
```
alihejazi, line 247: You need to add this to make sure the donate button doesn't show on desktops when the campaign is not open for donations.
```suggestion
             *ngIf="donateEnabled"
          ></biggive-button>
```

PR #834, opened 2022-12-07
DON-619 update homepage cards

alihejazi:


bdsl:
Screenshot from my local. This is all I can see before I scroll. Should there be so much space above the cards?

image

src/app/home/home.component.html
```diff
@@ -9,17 +9,50 @@
...
+        button-label='Read more'
+        button-colour-scheme='clear-primary'
+        class="basic-card-padding"
+      >
+      </biggive-basic-card>
+
+      <biggive-basic-card
+        space-below='4'
+        icon-colour='brand-5'
+        background-image-url='/assets/images/colour-orange.png'
```
bdsl, line 31: I wonder why we actually need an image of a plain colour. But I tried passing the `background-colour` prop instead with this #f07d00 colour and it it didn't work.

bdsl (approved):
Added a couple of comments, not blockers.

src/app/home/home.component.html
alihejazi: `background-colour` will only work as a CSS style. See the component docs here: https://components-staging.thebiggivetest.org.uk/?path=/docs/components-content--basic-card-component. This is fine as is.

PR #835, opened 2022-12-08
DON-273: Show error instead of password reset form if token invalid

bdsl:
This PR relies on https://github.com/thebiggive/identity/pull/44 so that should be merged first.

We show the error below instead of the new password form if someone clicks a password reset link that has expired or doesn't exist.

Ideally like to make it an actual 404 error page, but I'm not sure how to persuade react router to do that, and this way may be fine, especially since we don't seem to serve a 404 page anywhere at the moment.

Screenshot from 2022-12-08 12-42-28

### DIFF
--- a/src/app/reset-password/reset-password.component.html
+++ b/src/app/reset-password/reset-password.component.html
@@ -10,7 +10,7 @@
     ></biggive-heading>
     <!-- Note: check explicitly for `undefined` here, not for `!saveSuccessful`, because
     `saveSuccessful` has can be undefined, false, or true, all with a unique meaning. -->
-    <div *ngIf="saveSuccessful === undefined && !savingNewPassword">
+    <div *ngIf="saveSuccessful === undefined && !savingNewPassword && tokenValid">
       <mat-dialog-content class="mat-typography">
         <form [formGroup]="passwordForm">
           <p class="b-rt-0">Please enter your new password (min 10 characters)</p>
@@ -51,5 +51,9 @@
     <p *ngIf="saveSuccessful === false" class="error">
       Sorry, there was an error saving your new password. Your link may have expired. Please <a href="https://www.thebiggive.org.uk/s/contact-us" target="_blank">contact us</a> if this message persists.
     </p>
+
+    <div *ngIf="!tokenValid"  class="error">
+      Sorry, there was an error changing your new password. Your link may have expired. Please <a href="https://www.thebiggive.org.uk/s/contact-us" target="_blank">contact us</a> if this message persists.
+    </div>
 
 </div>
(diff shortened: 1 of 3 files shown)


src/app/reset-password/reset-password.component.ts
```diff
@@ -55,6 +56,14 @@ export class ResetPasswordComponent implements OnInit {
     this.route.queryParams
       .subscribe(params => {
         this.token = params.token;
+        this.identityService.checkTokenValid(this.token).subscribe(
```
bdsl, line 59: of course if the backend is slow to respond or times out, the user will first see the new password form and then see it disappear and be replaced with the error message. Not sure what the best way to avoid that is, or if we should just live with it.

alihejazi: @bdsl this is a really good consideration. To work around this, we can simply use a `<mat-spinner>`, like we use elsewhere in the app. The spinner should show whilst the server does its processing, until it returns valid or invalid.
To do this, we can default `tokenValid` to undefined, like this:
`tokenValid: boolean|undefined = true;`
Then, we can render the `<mat-spinner>` in the template file while `tokenValid === undefined`

bdsl: Thanks, done in 2c4f0028

src/app/reset-password/reset-password.component.html
```diff
@@ -52,4 +60,8 @@
       Sorry, there was an error saving your new password. Your link may have expired. Please <a href="https://www.thebiggive.org.uk/s/contact-us" target="_blank">contact us</a> if this message persists.
     </p>
 
-</div>
\ No newline at end of file
+    <div *ngIf="tokenValid === false"  class="error">
+      Sorry, there was an error changing your new password. Your link may have expired. Please <a href="https://www.thebiggive.org.uk/s/contact-us" target="_blank">contact us</a> if this message persists.
```
alihejazi, line 64: @bdsl The user will have never tried to change their password when this message is shown, so let's remove that part:
```suggestion
      Sorry, your link may have expired. Please <a href="https://www.thebiggive.org.uk/s/contact-us" target="_blank">contact us</a> if this message persists.
```

src/app/reset-password/reset-password.component.ts
alihejazi: Awesome!

src/app/reset-password/reset-password.component.html
bdsl: Yeah that's true. I think I was confused because I was thinking about the several different possible reasons for the error.

alihejazi (approved):

PR #836, opened 2022-12-08
COM-28: npm update v1.0.179

alihejazi:


bdsl (approved):
I can't really review the diff here, I'll take it on trust that it's right.

PR #837, opened 2022-12-09
DON-615: Add CSS styles to menu entries

bdsl:
Relies on new styles added in components PR 91, but should be safe to deploy now as nothing is being removed.

See https://github.com/thebiggive/components/pull/91


alihejazi (approved):
@bdsl Nice one. I thought we were missing the usage of `icon-emergency` but I realised it's because we haven't yet added the 'Emergency Match Fund' menu item.

PR #851, opened 2022-12-30
Stop loading global styles both globally and on `AppComponent`

NoelLH:
Global styles are already loaded in `angular.json` so they seem to have been extraneous here. Removing this decreases the `main` bundle's uncompressed size quite a bit.


bdsl (approved):
LGTM, nice saving!

PR #854, opened 2023-01-04
DON-708: Update GMF and C4C highlight cards to show applications open…

bdsl:
… from 9 Jan

At the time set the cards should automatically change to appear as in the second screenshot, although probably only after a browser refresh.

Screenshot from 2023-01-04 11-57-35

image

It's relying on the clock in the browsing device to check the time. I'd prefer to rely on the server's clock to be authoritative instead of the browser, but this may be good enough for now, as I'm not sure what the best way would be to use the server's clock.

### DIFF
--- a/src/app/home/home.component.html
+++ b/src/app/home/home.component.html
@@ -12,20 +12,41 @@
 
     <biggive-grid
       column-count="3">
-      <biggive-basic-card
+      <biggive-basic-card *ngIf="greenMatchApplicationsOpen"
         space-below='4'
         icon-colour='brand-3'
         background-image-url='/assets/images/card-background-gmf.jpg'
-        main-title='Save the Date for the Green Match Fund'
+        main-title='Applications for Green Match Fund now open'
+        subtitle='Deadline is 17 January 2023'
+        button-url='https://www.thebiggive.org.uk/charities/s/login/'
+        button-label='Apply Now'
+        button-colour-scheme='clear-primary'
+        class="basic-card-padding">
+      </biggive-basic-card>
+      <biggive-basic-card *ngIf="! greenMatchApplicationsOpen"
+        space-below='4'
+        icon-colour='brand-3'
+        background-image-url='/assets/images/card-background-gmf.jpg'
+        main-title='Save the Date for Green Match Fund'
         subtitle='Applications open 9 Jan 2023'
         button-url='https://blog.thebiggive.org.uk/green-match-fund/'
         button-label='Read more'
         button-colour-scheme='clear-primary'
-        class="basic-card-padding"
-      >
+        class="basic-card-padding">
       </biggive-basic-card>
 
-      <biggive-basic-card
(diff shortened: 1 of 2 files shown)


src/app/home/home.component.ts
```diff
@@ -21,5 +24,10 @@ export class HomeComponent implements OnInit {
       false,
       'https://images-production.thebiggive.org.uk/0011r00002IMRknAAH/CCampaign%20Banner/db3faeb1-d20d-4747-bb80-1ae9286336a3.jpg',
     );
+
+    const openingDate = new Date("2023-01-09T12:00:00.000+00:00");
```
bdsl, line 28: I would have rather called the timezone Europe/London instead of +00:00, but I don't know if that's supported natively by  browsers.

bdsl: Not essential now, but if we don't have an easy way to this it might be worth bringing in a library like Luxon at some point ( https://moment.github.io/luxon/index.html#/zones)  - although that still doesn't solve the problem of relying on the browser's clock.

src/app/home/home.component.html
```diff
@@ -12,20 +12,41 @@
...
         background-image-url='/assets/images/card-background-gmf.jpg'
-        main-title='Save the Date for the Green Match Fund'
+        main-title='Applications for Green Match Fund now open'
+        subtitle='Deadline is 17 January 2023'
+        button-url='https://www.thebiggive.org.uk/charities/s/login/'
+        button-label='Apply Now'
+        button-colour-scheme='clear-primary'
+        class="basic-card-padding">
+      </biggive-basic-card>
+      <biggive-basic-card *ngIf="! greenMatchApplicationsOpen"
```
bdsl, line 26: I couldn't quite get the else part of ngIf to work, hence the slightly clunky use of *if not* logic.

src/app/home/home.component.ts
alihejazi: As it's such a short-term change I think this is fine for now.

alihejazi:
@bdsl Main thing I've been considering with this PR is if there would've been value in rendering out each card once but use TS variables to set the prop values instead of hard coding them in. But as there are 4 props to change for each card, we'd need 4 new TS variables, so this alongside the fact that it's a short term fix makes me feel it's ok as is.

alihejazi (approved):

bdsl:
> @bdsl Main thing I've been considering with this PR is if there would've been value in rendering out each card once but use TS variables to set the prop values instead of hard coding them in. But as there are 4 props to change for each card, we'd need 4 new TS variables, so this alongside the fact that it's a short term fix makes me feel it's ok as is.

We could put the 4 props into an object, and have no copy in the template file all the copy in typescript. Probably a good idea but let's leave it for now.

src/app/home/home.component.ts
NoelLH: @bdsl Angular already has powerful pipes for this stuff – I doubt we need more libraries

PR #856, opened 2023-01-04
DON-715: WIP allow use of local components

bdsl:
Screenshots show a mix of local changes within this codebase, and local changes in the components repo. Exactly where built code is saved and what we need to to do to update is slightly unclear to me.

XXX's are from local changes in the component repo
image

Yeah's are from local changes in the component repo
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,15 @@ To use `ng` commands directly, e.g. to generate new code scaffolding with the CL
 
     npm install -g @angular/cli
 
+## Developing with local copy of Big Give Components
+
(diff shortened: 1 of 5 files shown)


bdsl:
As discussed, this seems to work at least somewhat on my machine, not really for @alihejazi . But this PR is only adding new functionality and only for the dev environment, so it's not a big problem if it doesn't always work - we can avoid using it or fix it later. Will try out using it as part of a future development task.

alihejazi (approved):

PR #859, opened 2023-01-06
COM-22: Use stricter tsconfig settings to prevent errors similar to t…

bdsl:
…hat fixed in COM PR #86

See https://github.com/thebiggive/components/pull/86 See https://github.com/thebiggive/components/pull/107

### DIFF
--- a/src/app/campaign-groups.service.ts
+++ b/src/app/campaign-groups.service.ts
@@ -89,7 +89,7 @@ export class CampaignGroupsService {
     const beneficiaries: {name: string, icon: IconDefinition}[] = this.getBeneficiaries();
 
     for (let i = 0; i < beneficiaries.length; i++) {
-      const beneficiaryObj: {name: string, icon: IconDefinition} = beneficiaries[i];
+      const beneficiaryObj: {name: string, icon: IconDefinition} = beneficiaries[i]!;
       names.push(beneficiaryObj.name);
     }
 
@@ -182,7 +182,7 @@ export class CampaignGroupsService {
     const categories: {name: string, icon: IconDefinition}[] = this.getCategories();
 
     for (let i = 0; i < categories.length; i++) {
-      const categoryObj: {name: string, icon: IconDefinition} = categories[i];
+      const categoryObj: {name: string, icon: IconDefinition} = categories[i]!;
       names.push(categoryObj.name);
     }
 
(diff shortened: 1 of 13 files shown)


src/app/campaign-groups.service.ts
```diff
@@ -442,7 +442,7 @@ export class CampaignGroupsService {
     );
 
     // For an unknown/invalid beneficiary, show 'other' symbol.
-    return matchingItems[0].icon || faPlus;
```
bdsl, line 445: this line specifically is quite similar to the one fixed before. Would have thrown a TypeError if matchingItems[0] was undefined.

src/app/home/home.component.html
```diff
@@ -27,7 +27,7 @@
         space-below='4'
         icon-colour='brand-3'
         background-image-url='/assets/images/card-background-gmf.jpg'
-        main-title='Save the Date for Green Match Fund'
+        main-title='Save the Date for Green Match Fund (text edited in donate-frontend)'
```
alihejazi, line 30:
```suggestion
        main-title='Save the Date for Green Match Fund'
```

src/app/search.service.ts
```diff
@@ -108,7 +108,8 @@ export class SearchService {
   }
 
   hasTermFilterApplied(): boolean {
-    return (this.getQueryParams().term?.length > 0);
+    const term = this.getQueryParams().term || '';
+    return (term.length > 0);
```
alihejazi, line 112: Not sure this adds much. The previous code was already type-safe.
```suggestion
    return (this.getQueryParams().term?.length > 0);
```

If possible can you please explain your reasoning for this change?

alihejazi (approved):

src/app/search.service.ts
bdsl: The previous version produces an error with the new tsconfig. `length` could be undefined, so we can't use it as a number:

```

Error: src/app/search.service.ts:111:13 - error TS2532: Object is possibly 'undefined'.

111     return (this.getQueryParams().term?.length > 0);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

alihejazi: Ok. Thanks for explaining.

PR #862, opened 2023-01-06
Use components version 1.0.194

bdsl:
See https://github.com/thebiggive/components/pull/110


bdsl:
Build failure is just the Jira update step.

alihejazi (approved):

PR #866, opened 2023-01-09
Don 649 accesibility

bdsl:
Use appropriate heading levels for cards and footer, resolving issue reported by Ligghthouse in chrome dev tools.

image


src/app/home/home.component.html
```diff
@@ -12,8 +12,9 @@
 
     <biggive-grid
       column-count="3">
-      <biggive-basic-card *ngIf="greenMatchApplicationsOpen"
+      <biggive-basic-card
         space-below='4'
+        heading-level = 2
```
alihejazi, line 17: For consistency:
```suggestion
        heading-level=2
```

alihejazi: @bdsl Also not sure but pls test locally that it works when passed as an int and not as a string.

bdsl: @alihejazi I have tested it locally, it works. The screenshot in the PR descriptions shows the issue goes away in Lighthouse.

src/app/home/home.component.html
```diff
@@ -23,21 +24,11 @@
...
-        button-label='Read more'
-        button-colour-scheme='clear-primary'
-        class="basic-card-padding">
-      </biggive-basic-card>
 
-      <biggive-basic-card *ngIf="championsForChildrenApplicationsOpen"
+      <biggive-basic-card
         space-below='4'
         icon-colour='brand-5'
+        heading-level = 2
```
alihejazi, line 31:
```suggestion
        heading-level=2
```

src/app/home/home.component.html
```diff
@@ -46,20 +37,10 @@
...
-        subtitle='Applications open 9 Jan 2023'
-        button-url='https://blog.thebiggive.org.uk/champions-for-children/'
-        button-label='Read more'
-        button-colour-scheme='clear-primary'
-        class="basic-card-padding">
-      </biggive-basic-card>
       <biggive-basic-card
         space-below='4'
         icon-colour='primary'
+        heading-level = '2'
```
alihejazi, line 43:
```suggestion
        heading-level='2'
```

bdsl:
I think in principle we could merge this as-is without breaking anything, but maybe it's worth waiting to include the new version and avoid referencing a prop that doesn't exist in the current version.

alihejazi (approved):

PR #867, opened 2023-01-09
DON-707 Handle invalid slug

bdsl:


src/app/campaign.resolver.ts
```diff
@@ -35,6 +38,16 @@ export class CampaignResolver implements Resolve<any> {
       );
     }
 
+    if (campaignSlug && fundSlug && campaignSlug !== "campaign") {
+      const query = this.campaignService.buildQuery(this.searchService.selected, 0, campaignId ?? undefined, campaignSlug, fundSlug);
+      const searchResult = this.campaignService.search(query as SearchQuery).subscribe(
+        () => {},
+        () => {
+          this.router.navigateByUrl(`/${campaignSlug}/`);
```
bdsl, line 46: I'm not sure if this error handler will or is guaranteed to run during SSR

src/app/campaign.resolver.ts
```diff
@@ -35,6 +38,16 @@ export class CampaignResolver implements Resolve<any> {
       );
     }
 
+    if (campaignSlug && fundSlug && campaignSlug !== "campaign") {
```
alihejazi, line 41: Why do we need to check the following?
`campaignSlug !== "campaign"`

alihejazi:
@bdsl I've pushed this commit: https://github.com/thebiggive/donate-frontend/pull/867/commits/5272d755cc1066103e2f24f7fd326b994f0339f6

src/app/campaign.resolver.ts
bdsl: My reading of the requriement was we only want to do this redirect when there's a specific metacampaign - e.g.

from https://donate.thebiggive.org.uk/christmas-challenge-2022/xxxxx to https://donate.thebiggive.org.uk/christmas-challenge-2022/ but not from https://donate.thebiggive.org.uk/campaign/abcd to https://donate.thebiggive.org.uk/campaign/

alihejazi (approved):

PR #868, opened 2023-01-09
Fix deadline on GMF card

bdsl:


alihejazi (approved):

bdsl:
The message from github that says this was merged into `main` is not actually true. @alihejazi and I checked, it was in fact merged into develop. I think github got confused because I changed the base on the PR at almost exactly the same time as Ali did the merge. I wanted to check how it would look as a PR against main since this is effectively a bugfix.

PR #870, opened 2023-01-09
Fix deadline on GMF card

alihejazi:


bdsl (approved):
Beautiful

bdsl:
Closing because the checks will never pass since `deploy-staging` doesn't run if the PR isn't from `develop`.

PR #873, opened 2023-01-10
DON-667 – start updating Angular Material to v15

NoelLH:


NoelLH:
While I think some of my findings may be useful, this branch is looking increasingly worse than starting over, because:
* I have confirmed that builds break when using the provided tools on the versions locked here so far, but work with newer 15.x libraries; and
* The migration covers styling we have already deleted or can do so soon

I'll close and repeat the process on a new branch from `develop` once https://github.com/thebiggive/donate-frontend/pull/986 is merged there.

NoelLH:
Superseded by https://github.com/thebiggive/donate-frontend/pull/988 which I hope to get good enough for testing today.

PR #879, opened 2023-01-17
DON-718: Update links on homepage tiles

bdsl:


NoelLH (approved):
Looks good – though probably worth waiting until the blog's stable before merging

PR #890, opened 2023-01-27
DON-702 – use component property to expand button

NoelLH:


bdsl:
This makes the button quite a bit narrower than it is on develop

Develop

image


This branch:

image

The width we have now was requested in the QA feedback.

NoelLH:
> On my machine changing between `full-width="true"` and `full-width="false"` has no visible effect.

@bdsl I tested the button in isolation as I didn't have long.

What you're saying and the fact that the property can be seen working and filling the entire browser in Storybook suggests to me that the containing element is too small for what you're trying to do, and that the `nbsp;`s work by overflowing it.

This doesn't feel like a great solution compared to providing a containing element that is the desired size. We already use `full-width="true"` in the app so I don't understand the rationale for introducing this hack instead of getting the layout to play nicely.

bdsl:
I think the rationale was just that I couldn't find an easy way to get the layout to play nicely, it felt like it would have been quite a bit more work adjusting the component in `com` and then using it here, the design instructions I was given were specific about the width on this particular page, and I didn't know whether we'd have the same thing on other pages.

If I was going to do a less hacky solution I probably would have looked to add a `min-width` prop or something that could be passed into the button from the page.

NoelLH:
OK – I think it's OK to leave it for now @bdsl.

I was concerned about accessibility as well as maintainability, but I  just tried it with VoiceOver in Chrome and it seemed to bypass the space and not visually overlay the same amount of blank space in the caption of what it was reading.

Additionally, I found that we probably have a bigger issue in that these shadow elements aren't receiving focus like vanilla links are. Adding `tabindex="0"` appears to fix this and should probably happen in `components`.

I would still suggest we ought to be setting `full-width="true"` so that the stated intention is not directly contradicting the workaround.

bdsl:
@NoelLH OK, no harm in changing setting `full-width="true"` and keeping the `&nbsp`'s.

bdsl (approved):

PR #893, opened 2023-02-01
DON-593: Show specific error message if we can't register user becaus…

bdsl:
…e they are already registered

Screenshot from 2023-02-01 18-09-03

### DIFF
--- a/src/app/donation-complete/donation-complete.component.ts
+++ b/src/app/donation-complete/donation-complete.component.ts
@@ -37,6 +37,7 @@ export class DonationCompleteComponent implements OnInit {
   encodedPrefilledText: string;
   recaptchaIdSiteKey = environment.recaptchaIdentitySiteKey;
   registerError?: string;
+  duplicateEmailAddressWithPassword: boolean = false;
   registrationComplete = false;
   shareUrl: string;
   timedOut = false;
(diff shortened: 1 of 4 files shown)


src/app/donation-complete/donation-complete.component.ts
```diff
@@ -158,6 +159,10 @@ export class DonationCompleteComponent implements OnInit {
           }
         },
         (error: HttpErrorResponse) => {
+          if (error.error?.error?.type === "DUPLICATE_EMAIL_ADDRESS_WITH_PASSWORD") {
```
bdsl, line 162: Not critical, but would be nice to have a more type safe way to do this. At the moment it's relying on `any` a lot. If it doesn't work we shouldn't get a crash, it will just show the old error message like it did before this PR.

NoelLH: I agree, I don't love these!

NoelLH (dismissed):

bdsl:
I've adjusted colours as discussed on Jira.

bdsl:
And changed again. New screenshot:

image

bdsl:
@NoelLH Please don't merge, I want to hold this back and deploy other changes first.

src/app/donation-complete/donation-complete.component.ts
bdsl: hmm no of course we would get a crash if one of the objects was not null or undefined but didn't have the property we need. But I can cope, it's unlikely and it's only crashing in what we already know is an error handling task.

PR #894, opened 2023-02-02
REG-23: Use unique nav titles for footer nav elements

bdsl:
This will also require updating the components libary version to work

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -126,7 +126,13 @@
 
 <router-outlet></router-outlet>
 
-<biggive-footer heading-level=3>
+<biggive-footer
+  heading-level=3
+  primary-nav-title="Match Funding Opportunities"
+  secondary-nav-title="Resources"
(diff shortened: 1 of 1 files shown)


src/app/app.component.html
```diff
@@ -126,7 +126,13 @@
 
 <router-outlet></router-outlet>
 
-<biggive-footer heading-level=3>
+<biggive-footer
+  heading-level=3
+  primary-nav-title="Match Funding Opportunities"
+  secondary-nav-title="Resources"
+  tertiary-nav-title="About"
+  postscript-nav-title="postscript"
```
NoelLH, line 134: Not sure about this label. Should we request copy that gives a better idea what's inside the area?

bdsl: Yes let me ask on Slack.

bdsl:
Changes in https://github.com/thebiggive/components/pull/126 make this redundant.

PR #895, opened 2023-02-03
Support our future "global" domain

NoelLH:


bdsl (approved):
LGTM, matches much better what's in the staging and regression files.

PR #898, opened 2023-02-06
DON-720: Change banner on thank you page

bdsl:
Before and after images
Screenshot from 2023-02-06 10-06-11

image

New image was from Jira attachment as PNG. I converted it from PNG to JPEG to reduce the size from 1.66 MB down to 366KB, for faster loading especially on slow mobile connections. There's almost no visible difference between the images - if I zoom in to 500% and switch between them I can just see that the edges of the triangles are slightly softer on the JPEG than they are on the PNG I was given:

PNG:
image

JPEG:

image


bdsl:
Edited image:

image

NoelLH (approved):

PR #899, opened 2023-02-06
ID-36: Display password related error as sent from server

bdsl:
image

### DIFF
--- a/src/app/donation-complete/donation-complete.component.html
+++ b/src/app/donation-complete/donation-complete.component.html
@@ -48,15 +48,9 @@
         using your chosen email address and password.
       </p>
 
-      <p *ngIf="duplicateEmailAddressWithPassword" class="error" aria-live="assertive">
+      <p *ngIf="registerError" class="error" aria-live="assertive">
         <fa-icon [icon]="faExclamationTriangle"></fa-icon>
-        <i class="fa fa-exclamation-triangle" aria-hidden="true"></i> Your password could not be set. There is already a password set for your email address.
-      </p>
-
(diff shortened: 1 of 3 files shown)


src/app/donation-complete/donation-complete.component.html
```diff
@@ -48,15 +48,9 @@ <h2>Thank you!</h2>
         using your chosen email address and password.
       </p>
 
-      <p *ngIf="duplicateEmailAddressWithPassword" class="error" aria-live="assertive">
+      <p *ngIf="registerError" class="error" aria-live="assertive">
```
NoelLH, line 51: Probably not new or critical for this, but based on my understanding now I think we would probably be better surrounding just the error description with a span that has the `aria-live` attribute. Normally these regions should be empty when the browser first parses the page. This also sidesteps any potentially a11y impact from us losing `aria-hidden` with this PR's change.

NoelLH: @bdsl I think that `assertive` is appropriate for important errors, _if_ the region is narrow enough i.e. doesn't possibly include meaningless icon info.

I was thinking of pages like https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live which show empty elements being used first. I don't think this is a requirement of the spec but is frequently recommended for compatibility because more readers pick up changes inside elements if they exist on first page load. In fact, what I mentioned would not be sufficient to cover that eventuality: we would also need to make only the text contents of the region, and not the region itself, `*ngIf`-conditional.

bdsl: OK. Let's leave it as is for now since it's not a change from this PR.

PR #904, opened 2023-02-08
DON-729: DEC-Turkey-Syria homepage card

bdsl:
Currently looks like so:

image

Before this is ready we need to

- [x] Agree copy to replace "You donate. We double it" (or keep as-is)
- [x] Choose and set a button url
- [x] Ensure campaign (or any other content) is live at chosen URL


bdsl:
If we were to change the subtitle to "Double your donation" it would look like this:

image

src/app/home/home.component.html
```diff
@@ -12,6 +12,22 @@
...
       column-count="3">
+
+      <biggive-basic-card
+        space-below='4'
+        icon-colour='brand-4'
+        heading-level='2'
+        background-image-url='/assets/images/emergency-card.png'
+        main-title='Turkey-Syria Earthquake Appeal'
+        [subtitle]="'Double the impact of your donation'"
+        button-url='/turkey-syria-earthquake-appeal'
```
bdsl, line 23: https://donate.thebiggive.org.uk/turkey-syria-earthquake-appeal is live already

PR #914, opened 2023-02-09
DON-731 Reorder nav menu on donate to match menu on Wordpress

bdsl:
Before:

image

After:

image


bdsl:
Additional screenshot from Firefox emulating Iphone 11 Pro:

image

PR #915, opened 2023-02-09
DON-610: WIP new donation stepper

bdsl:
Initially the new is the same as the old. To enable it, add `?2023-design` to the end of the address bar.


bdsl:
I've started with the background and shadow. Before / after:

image

image

bdsl:
Closing for now, may choose to re-open in future.

PR #916, opened 2023-02-10
DON-586: Reskin Donation page's side/lower campaign info card

bdsl:
I've got the same text appearing on the sidebar of the donation start page as on the campaign info page, but the styling is not currently the same, needs fixing before merging.

Before:

image

On this branch HEAD currently:

image

Sidebar from info page that we should be matching:

image


bdsl:
OK, so I think I've finally got something usable, although the code is a mess. I took a long time struggling to get the campaign-info component to style correctly in the new context. In the end I gave up and copied the code out of that component into the donation start component. This will mean we'll have to edit in both places separately in future.

Sidebars on the info and donate start pages now look identical except for a slight variation in the icon size that I don't know the cause of, and of course the presence of the button on the info page:

Info (unchanged):

image


Donation start:

image

PR #921, opened 2023-02-14
DON-738: Homepage updates

bdsl:
This makes the homepage vary depending the time its viewed. At the current time it will look like this:

image

Depends on changes from https://github.com/thebiggive/components/pull/128


package.json
```diff
@@ -36,8 +36,8 @@
     "@angular/platform-browser-dynamic": "^15.0.4",
     "@angular/platform-server": "^15.0.4",
     "@angular/router": "^15.0.4",
-    "@biggive/components": "^202302031402.0.0",
-    "@biggive/components-angular": "^202302031402.0.0",
+    "@biggive/components": "^202302141635.0.0",
+    "@biggive/components-angular": "^202302141635.0.0",
```
bdsl, line 40: changes above required to bring in work from https://github.com/thebiggive/components/pull/128/files

PR #922, opened 2023-02-14
Revert "Revert "DON-586: Reskin Donation page's side/lower campaign info card""

bdsl:
Reverts thebiggive/donate-frontend#919, and then adds a banner image to the top of the second column.


src/app/donation-start/donation-start.component.scss
```diff
@@ -1,6 +1,14 @@
 @use '@angular/material' as mat;
 @import '../../abstract';
 
+* {
+  //outline: 1px dotted red;
+}
+
```
bdsl, line 7:
```suggestion
```

PR #931, opened 2023-02-17
DON-727: Improve payment method layout

bdsl:
I think this might be all we need for display of payment methods. There is more information that we have, and that the API is sending to the browser so people can see if they no how, but I think things like the billing address attached to each payment method and what country it's in would just clutter the page if we showed them here.

This is how the page looks if you have several cards saved:

image

I'm pretty sure we don't allow any payment methods other than cards, but if someone does have a different method stored (e.g. [Cleaplay](https://www.clearpay.co.uk/en-GB) then it would be displayed like this:

image

If people agree that this is enough for the display I'll move on to implementing deletion.

### DIFF
--- a/src/app/my-account/my-account.component.html
+++ b/src/app/my-account/my-account.component.html
@@ -20,7 +20,8 @@
         </div>
       </biggive-page-section>
       <biggive-page-section>
-        <table *ngIf="person" class="personal-details">
+        <h2>Your Details</h2>
+        <table *ngIf="person" id="personal-details">
           <!--
           Not sure why Typescript doesn't allow me to concatenate like this.
           <h2>{{person.first_name + person.last_name}}</h2>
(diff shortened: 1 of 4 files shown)


src/app/my-account/my-account.component.html
```diff
@@ -45,19 +46,21 @@ <h2>{{person.first_name + person.last_name}}</h2>
...
-              {{displayPaymentCard(method.card)}}
-            </div>
-          </div>
-        </div>
+        <table id="paymentMethods" *ngIf="paymentMethods !== undefined && paymentMethods.length > 0">
+          <tr *ngFor="let method of paymentMethods">
+            <td>{{(method.card?.brand || method.type).toUpperCase()}}</td>
+            <td *ngIf="method.card !== undefined">
+              Card Ending: {{method.card.last4}} <br />
+              Expiry Date: {{method.card.exp_month.toString().padStart(2, "0")}}/{{method.card.exp_year}}
```
bdsl, line 60: `padStart` doesn't work in I.E, but thankfully I.E went away this week. https://uk.pcmag.com/browsers/145433/microsoft-permanently-disables-internet-explorer-with-valentines-day-update

PR #932, opened 2023-02-17
DON-727: Put account details over a blue background

bdsl:
Now appears like this on desktop:

image

And like this on mobile:

image


bdsl:
Updated screenshots:

image

image

PR #938, opened 2023-02-20
DON-724: Fix error handling & page update on payment method deletion

bdsl:


src/app/my-account/my-account.component.html
```diff
@@ -40,7 +40,7 @@ <h2>{{person.first_name + person.last_name}}</h2>
       <h2>Your Payment Methods</h2>
 
       <div *ngIf="paymentMethods === undefined">
-        <mat-spinner [diameter]="40" aria-label="Loading payment methods"></mat-spinner>
+        <mat-spinner color="accent" [diameter]="40" aria-label="Loading payment methods"></mat-spinner>
```
bdsl, line 43: Blue spinner was invisible on a blue background. This makes it green.

src/app/my-account/my-account.component.ts
```diff
@@ -66,8 +66,15 @@ export class MyAccountComponent implements OnInit {
   }
 
   deleteMethod(method: PaymentMethod) {
-    this.donationService.deleteStripePaymentMethod(this.person, method, this.jwtAsString()).
-    subscribe(this.loadPaymentMethods)
+    this.paymentMethods = undefined;
```
bdsl, line 69: Makes the spinner re-appear the moment someone clicks the delete button

src/app/my-account/my-account.component.ts
```diff
@@ -66,8 +66,15 @@ export class MyAccountComponent implements OnInit {
...
+
+    this.donationService.deleteStripePaymentMethod(this.person, method, this.jwtAsString()).subscribe(
+      this.loadPaymentMethods.bind(this),
+      error => {
+        this.loadPaymentMethods.bind(this)()
+        alert(error.error.error)
+      }
```
bdsl, line 76: then whether the backend does what we ask of it or returns an error, we update the list of payment methods. If we had an error we display it using the native alert function.

PR #955, opened 2023-03-01
Fix server crashes from `window` use

NoelLH:
And document that we need to look out for this


bdsl:
Looks good to me - will merge.

I wonder why typescript's lib.dom.ts has defines 
```
declare var window: Window & typeof globalThis;
```

It seems like if instead they had what seems like a more accurate reflection of potentially polymorphic reality:

```
declare var window: Window & typeof globalThis | undefined;
```

Then the problem would go away as the compiler would force us to check we have a window before using it. Probably a good reason why they don't do that though.

PR #972, opened 2023-03-09
Remove login after donate

bdsl:
It looks like the "Log me in" function within the donation complete page has not been working since January. IMHO whenever something is not working at all we should immediately fix or remove it. I don't have a fix ready to apply so this removes it instead.

Before:

Screenshot from 2023-03-09 15-05-41

After:

image


bdsl:
This is a PR aimed at `main`. Created second PR from the same branch https://github.com/thebiggive/donate-frontend/pull/973 aimed at `develop` to allow testing there before merging this as a hotfix.

dee-jo (approved):

PR #978, opened 2023-03-13
BG2 2292 update icon design on campaign-info and  donation-start page

dee-jo:
### Campaign Page
before:
Screenshot from 2023-03-13 14-10-02

fotor_2023-3-13_14_21_5

after:

Screenshot from 2023-03-13 14-02-15

Screenshot from 2023-03-13 14-02-35

### DIFF
--- a/src/app/donation-start/donation-start.component.scss
+++ b/src/app/donation-start/donation-start.component.scss
@@ -299,24 +299,43 @@ a {
   flex-wrap: wrap;
 }
 
+// .icon-wrapper {
+//   display: flex;
+//   padding: 10px;
+
+//   // in campaign-details.component.scss the selector here is `svg` rather than `fa-icon`, but that doesn't seem to work
+//   // here.
+//   fa-icon {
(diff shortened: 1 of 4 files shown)


src/app/donation-start/donation-start.component.scss
```diff
@@ -296,27 +296,46 @@ a {
...
+    flex-direction: row;
+
+    .icon-wrapper:first-child span {
+      padding-left: 10px;
+    }
+
+    .icon-wrapper:first-child {
+      padding-left: 0;
+    }
+  }
```
bdsl, line 317: Looks like there's some duplication between this and the camaign details css. Do you think it would be worth de-duplicating, maybe moving the shared scss into `_mixins.scss` ? Not sure.

bdsl (commented):
Added comments inline. The only real blocker for me is I think we need some comments to explain the split and join and \u200B. Feel free to take or leave the ideas in the other comments.

bdsl (approved):
LGTM

PR #979, opened 2023-03-14
DON-725 refinements

NoelLH:
* Treat ID service as always-on to reduce complexity
* Handle captcha errors
* Set staging HTML `canonical` to the temporary domain


bdsl (approved):
LGTM

PR #984, opened 2023-03-15
BG-2292 align icons and text

dee-jo:
Untitled design-1 (1)

### DIFF
--- a/src/assets/scss/_mixins.scss
+++ b/src/assets/scss/_mixins.scss
@@ -221,8 +221,18 @@
   display: flex;
   padding: 10px 10px;
   align-items: center;
-  min-width: 33%; // works when there are 3 icons in each row
-  max-width: 33%; // works when there are less than 3 items in a row
+  min-width: 28%;
+  padding-left: 20px;
+
+  @media#{$breakpoint-sm} {
+    max-width: 28%;
+  }
+  @media#{$breakpoint-md} {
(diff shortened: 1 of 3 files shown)


bdsl:
Code looks fine but this stuff is hard to review as text. Looking at the display before and after:

before

image


After

image


So on mobile the texts aren't all quite aligned - the first one has some extra space on the left:

image

Not sure what's making the first one different. 


Also we may need a smaller breakpoint, because at a certain range of width in firefox the third text is cut off by the edge of the screen and there's no way to read it:

image

src/assets/scss/_mixins.scss
```diff
@@ -208,27 +208,27 @@
...
   display: flex;
-  padding: 20px 30px;
+  padding: 10px 10px;
   align-items: center;
-  width: 33%;
-  @media#{$breakpoint-md} {
-    padding-left: 0;
-  }
+  min-width: 33%; // works when there are 3 icons in each row
+  max-width: 33%; // works when there are less than 3 items in a row
```
bdsl, line 225: Is there a difference between `min-width: 33%; max-width: 33%;` and just `width: 33%;` ?

bdsl:
Alignment looks good - but there's still a problem of some text being cut off when I view in a wide Firefox window:

image

Everything else LGTM

bdsl (approved):
LGTM

PR #985, opened 2023-03-16
Update libs

NoelLH:


bdsl (approved):
Not reviewed in any detail, but I'll assume this is good.

PR #988, opened 2023-03-17
DON-717 – Update Angular Material to v15

NoelLH:
Largely the goal here is to use the new stuff to get the same-ish outcome.

The most notable known visual differences are

1. Fixed **position of various content vs. the menu** at desktop size.

Before (staging):
Screenshot 2023-03-17 at 17 01 17

After:
Screenshot 2023-03-17 at 17 02 42

2. Tweaked overrides on **"special" donation amount input fields** including the custom tip amount one. (I don't love the new grey background here, but think we need a clear approach to new form stuff agreed to know what to do with text fields more generally, if we're going to do something else.)

Before (staging):
Screenshot 2023-03-17 at 17 04 10

After:
Screenshot 2023-03-17 at 17 04 21

3. **Tabs** (soon to be replaced anyway, I think)

Before (staging):
Screenshot 2023-03-17 at 17 06 59

After:
Screenshot 2023-03-17 at 17 07 06


src/app/campaign-details/campaign-details.component.scss
```diff
@@ -150,16 +128,20 @@ h1 {
...
+  margin-top: 30px;
+}
+
+#campaign-highlights-beneficiaries-and-categories-desktop {
   display: none;
 }
 
-@media screen and (max-width: 960px){
-  #campaign-highlights-beneficiaries-and-categories-desktop {
+@media #{$breakpoint-lg} {
```
NoelLH, line 139: Just tweaking this so effective breakpoints are consistent. I'm not really convinced we need 2 totally separate elements for different screen sizes here, but don't have time to refactor that out today.

bdsl:
I think the new donation amount input is better and worse than the current one - the text is more readable because it doesn't overlap the boarder like the current one does, on the other hand the darker background when editing reduces contrast with the number. On balance I don't think its worse so we shouldn't block us merging this.

Code changes all look fine - I thought it was going to be more work to get rid of Material v14 everywhere.

bdsl (approved):

PR #989, opened 2023-03-17
BG2-2286: Use preset menu content from component library

bdsl:
Doing the same thing in the wordpress repo will mean we end up maintaining just one copy of the menu contents, not two.

For changes pulled in from components see
https://github.com/thebiggive/components/compare/b78498f69ce78252d5dd6d1eb563e56beddd4da4...44755acc4329b9f3b122f0c4adde1e4cbe109f5e

There are many changes but most are for new components and components used only in Wordpress.

Comparing my local to production - there a few minor spacing changes, will try and work out exactly what's caused them. This should be a pure refactor with no visible change.

Production and my then my local:

image
image


src/app/app.component.html
```diff
@@ -1,134 +1,10 @@
...
-    </li>
-    <li><a href="https://www.thebiggive.org.uk/s/contact-us">Contact us</a></li>
-    <li><a href="https://www.thebiggive.org.uk/charities/s/login">Charity login</a></li>
-  </ul>
-</biggive-main-menu>
+<biggive-main-menu
+  use-preset-menu-content="true"
+  donate-url-prefix="https://donate.thebiggive.org.uk"
+  blog-url-prefix="https://blog.thebiggive.org.uk"
+  experience-url-prefix="https://www.thebiggive.org.uk"
```
bdsl, line 5: these three were already hardcoded but I think I may was well take the chance to make them environment appropriate while before I merge this.

bdsl:
> There are many changes but most are for new components and components used only in Wordpress.

That's a lie, @bdsl . There are only a few changes in components, most of them from you for this ticket. Nothing there is for Wordpress.

bdsl:
I'm trying to make the links environment specific by doing something like the following, I'm not sure why it isn't working - it's giving me the "Donate URL prefix must set and start with http" error from the menu component. Will investigate on Monday.

```diff
diff --git a/src/app/app.component.html b/src/app/app.component.html
index 7284d177..ef19da38 100644
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,6 +1,6 @@
 <biggive-main-menu
   use-preset-menu-content="true"
-  donate-url-prefix="https://donate.thebiggive.org.uk"
+  donate-url-prefix="{donateUriPrefix}"
   blog-url-prefix="https://blog.thebiggive.org.uk"
   experience-url-prefix="https://www.thebiggive.org.uk"
   is-logged-in="false"
diff --git a/src/app/app.component.ts b/src/app/app.component.ts
index 92921e26..39399ed3 100644
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -11,6 +11,7 @@ import {NavigationService} from './navigation.service';
 import {Person} from "./person.model";
 import {IdentityService} from "./identity.service";
 import {flags} from "./featureFlags"
+import {environment} from "../environments/environment";
 
 @Component({
   selector: 'app-root',
@@ -22,6 +23,8 @@ export class AppComponent implements AfterViewInit, OnInit {
   public isLoggedIn: boolean = false;
   public flags: { profilePageEnabled: boolean };
 
+  public donateUriPrefix = environment.donateUriPrefix;
+
   constructor(
     private analyticsService: AnalyticsService,
     private identityService: IdentityService,
```

bdsl:
> I'm trying to make the links environment specific by doing something like the following, I'm not sure why it isn't working - it's giving me the "Donate URL prefix must set and start with http" error from the menu component. Will investigate on Monday.

This is resolved, thanks @NoelLH

bdsl:
With a slight change I'll make in the components library the before and after is now almost identical as it should be. There's a slight spacing change still in the primary menu on the right, but I don't think it makes it any worse, maybe not partuclarly worthwhile to try and track down exactly what causes it. Before and after will be:

image
image

dee-jo (approved):
LGTM

PR #990, opened 2023-03-18
DON-727: Don't try to get details for people with no password

NoelLH:


bdsl:
I added DON-727 to the PR title, as that's the only place that the code edited here is used.

bdsl (approved):

PR #991, opened 2023-03-18
Give thank you page longer, with an expotential backoff

NoelLH:


src/app/donation-complete/donation-complete.component.ts
```diff
@@ -53,7 +53,7 @@ export class DonationCompleteComponent implements OnInit {
   private maxTries = 5;
   private patchedCorePersonInfo = false;
   private person?: Person;
-  private retryInterval = 2; // In seconds
+  private retryBaseInterval = 2; // In seconds
```
bdsl, line 56: not a blocker but maybe worth renaming to `retryBaseIntervalSeconds` so that it doesn't need the comment and the seconds part can be repeated on every usage.

bdsl (approved):

bdsl:
@NoelLH I made some tiny adjustments in https://github.com/thebiggive/donate-frontend/pull/992 after merging this.

PR #995, opened 2023-03-21
DON-726 – redirect to new base URI + simplify config

NoelLH:


src/app/app.component.ts
```diff
@@ -70,11 +70,11 @@ export class AppComponent implements AfterViewInit, OnInit {
...
-      if (window.location.host === 'donate.biggive.org') {
-        // donations are currently not working on the new domain. No-one should be visiting that domain yet, but
-        // in case they do we redirect them to the old domain for now:
-
-        window.location.host = "donate.thebiggive.org.uk";
+      // Temporarily client-side redirect the previous non-global domain to the new one.
+      // Once most inbound links are updated, we can probably replace the app redirect
+      // with an infrastructure-level one a la parked domains.
+      if (window.location.host === 'donate.thebiggive.org.uk') {
+        window.location.host = 'donate.biggive.org';
```
bdsl, line 77: Instead of reversing the redirect, shouldn't we remove the current redirect, test the new domain manually in prod, and then put in the reversed redirect once were satisfied that it all works?

bdsl: IF we did that I guess it would also mean leaving out the changes in `src/app/app.module.ts` for now.

bdsl (commented):
Added comment inline, otherwise all LGTM.

src/app/app.component.ts
NoelLH: @bdsl Good call. Done on https://github.com/thebiggive/donate-frontend/pull/995/commits/099bf7f5d20dca0dcfcf61c3f1d898e2a4abb2d9

bdsl (approved):
LGTM

PR #1001, opened 2023-03-23
BG2-2296: Add link to PHP DonationStatus enum for reference

bdsl:


src/app/donation-status.type.ts
bdsl, line 1: As far as I can see we don't get any benefit by first defining an `enum` and then using `keyof typeof` to define a union of strings based on it. The enum isn't exported from this file, so I think we could replace the whole file with the comment and the below without losing anything. Maybe there's something I'm not seeing.

```ts
export type DonationStatus = 
  "Cancelled" 
  | "Chargedback" 
  | "Collected" 
  | "Failed"
  | "NotSet"
  | "Pending"
  | "Paid"
  | "PendingCancellation"
  | "Refunded"
  | "RefundingPending"
  | "Reserved"
```

src/app/donation-status.type.ts
```diff
@@ -1,3 +1,7 @@
+/**
+ * For details of status semantics see PHP enum MatchBot\Domain\DonationStatus at
+ * https://github.com/thebiggive/matchbot/blob/main/src/Domain/DonationStatus.php
```
bdsl, line 3: This link is dead right now because the PHP Enum hasn't been merged to main - but https://github.com/thebiggive/matchbot/blob/develop/src/Domain/DonationStatus.php has the enum and there's an open PR to add a docblock to every case.

dee-jo (approved):

PR #1002, opened 2023-03-23
ID-40: Bring back payment details to populate on Donate page

dee-jo:


bdsl (commented):
LGTM if you just delete the blank line at the top. No real idea wrapping the page in a conditional before broke the payment thing but I saw that this fixes it, and the conditional is only needed for the purposes of the menu.

bdsl (approved):

PR #1003, opened 2023-03-24
DON-751 – make all buttons explicitly not (very) round

NoelLH:
Probably safe to merge as-is.

There's also a discussion to conclude around whether / why we have the property to vary this at all.

### DIFF
--- a/src/app/donation-complete/donation-complete.component.html
+++ b/src/app/donation-complete/donation-complete.component.html
@@ -35,7 +35,7 @@
             open-in-new-tab="false"
             full-width="true"
             size="medium"
-            rounded="true"
+            rounded="false"
             centered="false"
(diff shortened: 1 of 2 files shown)


bdsl (commented):
I'm slightly confused by this - the description makes it sound like they're currently implicitly not round. In that case I like the move from implicit to explicit but I'm not sure how `rounded="true"` would be implicitly not very round.

Is this implementing design change, or just a refactor?

NoelLH:
@bdsl The reason I put "very" is that the `rounded="false"` style still has a non-zero border radius, just a smaller one.

The change is to use the "not round" style consistently, as requested most recently on the Jira issue. No refactor or change to what the property means for today.

bdsl (approved):

PR #1005, opened 2023-03-27
DON-735 - start preparing to take homepage content from Salesforce

bdsl:
So far this is pure refactoring, no changes made to homepage content or appearance. I think even if we didn't go further this has a bit of value as the new version seems like it might be slightly easier to edit, especially for non-devs, and is type checked.

image

### DIFF
--- a/src/app/home/home.component.ts
+++ b/src/app/home/home.component.ts
@@ -1,6 +1,8 @@
 import {Component, OnInit} from '@angular/core';
 
 import {PageMetaService} from '../page-meta.service';
+import {highlightCard} from "./hilightCard";
+import {environment} from "../../environments/environment";
 
 @Component({
   selector: 'app-home',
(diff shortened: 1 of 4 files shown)


src/app/home/home.component.ts
```diff
@@ -10,32 +12,38 @@ import {PageMetaService} from '../page-meta.service';
...
+        bodyText: "You donate.\nWe double it.",
+        iconColor: "primary",
+        backgroundImageUrl: new URL('/assets/images/blue-texture.jpg', environment.donateUriPrefix),
+        button: {
+          text: "Explore now",
+          href: new URL('/explore', environment.donateUriPrefix),
+        }
+      },
+    ];
+
```
bdsl, line 42: I suppose while this is hard-coded it could be in constructor instead of ngOnInit. But I don't think it matters much, and it would just have to move back to ngOnInit when it starts using a service call to get the data.

dee-jo (approved):

PR #1006, opened 2023-03-27
DON-735: Fake loading list of metacampaigns from SF and function to c…

bdsl:
This is starting to show the structure for implementing all the scenarios in the ticket - will still need to add in code to deal with dates & times, and the carousel (unless we avoid that as @NoelLH suggested)

### DIFF
--- a/src/app/metaCampaign.model.ts
+++ b/src/app/metaCampaign.model.ts
@@ -1,0 +1,12 @@
+/**
+ * Represents a metacampaign as it exists in Salesforce.
+ */
+export class MetaCampaign {
+  constructor(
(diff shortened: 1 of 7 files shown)


src/app/metaCampaign.model.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Represents a metacampaign as it exists in Salesforce.
```
bdsl, line 2:
```suggestion
 * Represents a metacampaign as it exists in Salesforce.
 * As discussed with @NoelLH we may be likely to delete this class and use Campaign instead.
```

dee-jo (approved):

PR #1007, opened 2023-03-28
Fix Home cards column count

NoelLH:


bdsl (commented):
This does change the appearance at least in the case where there is only one card, and it will conflict with changes I have in https://github.com/thebiggive/donate-frontend/pull/1006 . Does this fix a problem with the layout we have now?

NoelLH:
@bdsl Yes, most noticeably at this kind of breakpoint we are leaving space for 3 cards but showing 2 (this is Production)
Screenshot 2023-03-28 at 12 25 40

Making the column count dynamic makes the flex grid work as I think it's intended in `components`.

bdsl:
@NoelLH Thanks, this seems like a good fix then. I've taken the liberty of merging dev into your branch and resolving the conflict, please check you're happy with that before merging.

bdsl (dismissed):

PR #1009, opened 2023-03-28
DON-726 – Finalise domain move

NoelLH:
Put config simplification and redirects back, now we know the new domain works in general.

> Revert "DON-726 – prepare a safer phased move to the new domain"
> This reverts commit 099bf7f5d20dca0dcfcf61c3f1d898e2a4abb2d9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,8 +122,7 @@ To test re-building the image (use FontAwesome token marked "Pro Package Token" 
 
     docker build --rm --build-arg BUILD_ENV=staging --build-arg FONTAWESOME_NPM_AUTH_TOKEN=**token** -t thebiggive/donate-frontend:staging .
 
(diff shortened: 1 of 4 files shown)


bdsl (commented):
LGTM but it's be worth updating the value of donateURI prefix in the production environment file as part of this - otherwise every menu click will trigger a redirect.

NoelLH:
@bdsl I think everything I was aware of up to Tues was moved over to the global env key. I think it might be good to keep the old domain allowed in CSP rules etc. just for a few days until we're certain everything is stable on Production, at which point we can retire the extraneous domain.

I've pushed a commit updating new refs and the readme for now, on that basis.

bdsl (approved):

PR #1013, opened 2023-03-29
REG-33: Skip prompts to continue donation if in regression test envir…

bdsl:
…onment

It's not ideal to make the site behave more differently in regression environment than in prod than it has to, but it's better than leaving tests failing while other dev work continues.

It seems like it may be hard to make the regression test runner cope with this popup properly since it can appear at an unpredictable moment and interrupts other tasks. Removing it in regression environment for now should let us confirm for sure that it is what makes the tests fail, and restore our ability to detect regressions.


dee-jo (approved):
LGTM!

PR #1014, opened 2023-03-29
DON-727: Remove old release toggle

bdsl:
We're unlikely to ever need to turn off this feature now, so the feature flag is just cruft


dee-jo (approved):
LGTM!

PR #1024, opened 2023-04-04
DON-746 – implement Meta Pixel

NoelLH:
See also https://github.com/thebiggive/donate-frontend/pull/249 This was tested & went live in 2020 and a lot of code is borrowed, so this PR is a close relative of that one. But it also includes measures to improve on the security and privacy implications of having the tag – more detail in commit and code comments.

### DIFF
--- a/src/app/meta-pixel.service.ts
+++ b/src/app/meta-pixel.service.ts
@@ -2,6 +2,7 @@ import { Injectable } from '@angular/core';
 import { NavigationEnd, Router } from '@angular/router';
 
 import { environment } from '../environments/environment';
+import { Donation } from './donation.model';
 
 declare const fbq: ((action: string, eventOrId: string, params?: object) => void) | undefined;
 
@@ -38,12 +39,15 @@ export class MetaPixelService {
     document.head.appendChild(scriptInitMetaPixel);
   }
 
-  trackConversion(amount: number) {
+  trackConversion(donation: Donation) {
     if (!fbq) {
       return; // Skip the call gracefully if loading fails or 3rd party JS is blocked.
     }
 
-    fbq('track', 'Donate', {value: amount, currency: 'GBP'});
+    fbq('track', 'Donate', {
+      value: donation.donationAmount,
+      currency: donation.currencyCode,
+    });
   }
 
   private listenForRouteChanges() {
--- a/src/app/meta-pixel.service.ts
+++ b/src/app/meta-pixel.service.ts
@@ -19,6 +19,7 @@ export class MetaPixelService {
     this.listenForRouteChanges();
 
     const scriptInitMetaPixel = document.createElement('script');
+    scriptInitMetaPixel.setAttribute('nonce', 'OT22mYwcUVPp');
     scriptInitMetaPixel.innerHTML = `
       !function(f,b,e,v,n,t,s)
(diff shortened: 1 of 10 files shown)


src/app/meta-pixel.service.ts
```diff
@@ -0,0 +1,60 @@
...
+declare const fbq: (action: string, eventOrId: string, params?: object) => void;
+
+@Injectable({
+  providedIn: 'root',
+})
+export class MetaPixelService {
+  constructor(private router: Router) {}
+
+  init() {
+    if (!environment.metaPixelId) {
```
bdsl, line 15: would this be easier to test etc if we made the metaPixelId a constructor param instead of using it from the environment directly? Not sure.

src/app/meta-pixel.service.ts
```diff
@@ -0,0 +1,60 @@
+import { Injectable } from '@angular/core';
+import { NavigationEnd, Router } from '@angular/router';
+
+import { environment } from '../environments/environment';
+
+declare const fbq: (action: string, eventOrId: string, params?: object) => void;
```
bdsl, line 6: To be explicit about the fact that fbq might not be initialised if the stuff from meta doesn't load. This means the TS compiler will force us to check that it's a function before we call it.
```suggestion
declare const fbq: ((action: string, eventOrId: string, params?: object) => void) | undefined;
```

src/app/meta-pixel.service.ts
```diff
@@ -0,0 +1,60 @@
...
+    `;
+    document.head.appendChild(scriptInitMetaPixel);
+  }
+
+  trackConversion(amount: number) {
+    if (!fbq) {
+      return; // Skip the call gracefully if loading fails or 3rd party JS is blocked.
+    }
+
+    fbq('track', 'Donate', {value: amount, currency: 'GBP'});
```
bdsl, line 46: could make slightly briefer, fine either way imho:

```suggestion
    // Skip the call gracefully if loading fails or 3rd party JS is blocked.
    fbq && fbq('track', 'Donate', {value: amount, currency: 'GBP'});
```

src/app/meta-pixel.service.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  private listenForRouteChanges() {
+    this.router.events.subscribe(event => {
+      if (event instanceof NavigationEnd) {
+        if (!fbq) {
+          return; // Skip the call gracefully if loading fails or 3rd party JS is blocked.
+        }
+
+        fbq('track', 'PageView');
```
bdsl, line 56:
```suggestion
        fbq && fbq('track', 'PageView');
```

src/app/meta-pixel.service.ts
bdsl, line 46: Thought - should we take the currency from the donation object here instead of hard-coding GBP?

NoelLH: As discussed, bit more overhead to know that later lines are safe when reading

src/app/meta-pixel.service.ts
NoelLH: I think that given the constructor has only an injected param at the moment, this is maybe not worth the work to change how it's instantiated, given the functionality is not critical for the journey and the class not currently unit tested

bdsl (approved):

PR #1026, opened 2023-04-05
Com 55 main menu fix

bdsl:
Before mobile (on staging)

image

After mobile (on local)

image

Looks almost exactly the same, but now clicking to open sub & sub-sub menu items works. Also some words are changed to upper case.


dee-jo (approved):
LGTM

PR #1033, opened 2023-04-11
Fix many errors shown in tests like "ERROR: 'NG0303: Can't bind to 'n…

bdsl:
…gIf' since it isn't a known property of"

See https://stackoverflow.com/questions/63813806/angular-karma-jasmine-test-cant-bind-to-ngif-since-it-isnt-a-known-property

All spec.ts files now declare their component in the `declarations` section.


bdsl:
Hmm, this doesn't actually seem to fix that much - there are still many `ERROR`s shown under **npm run ci** at https://app.circleci.com/pipelines/github/thebiggive/donate-frontend/3469/workflows/dc7b5057-32b2-4433-9c9e-2024afe6a457/jobs/10078 . Will close for now.

PR #1034, opened 2023-04-11
DON-723: Allow editing saved card with modal

bdsl:
https://github.com/thebiggive/matchbot/pull/518 should be merged and deployed before this one.

image

image

image

Error messages are displayed as sent by matchbot. There will be a PR to that which will make it pass on error messages verbatim from the Stripe API. They're not all exactly as I would choose to write them, but I think it's probably not worthwhile to map them to custom messages.

### DIFF
--- a/src/app/my-account/my-account.component.ts
+++ b/src/app/my-account/my-account.component.ts
@@ -84,10 +84,10 @@ export class MyAccountComponent implements OnInit {
     return this.identityService.getJWT() as string;
   }
 
-  updateCard(paymentMethod: PaymentMethod) {
+  updateCard(card: PaymentMethod.Card, billingDetails: PaymentMethod.BillingDetails) {
     const updateCardDialog = this.dialog.open(UpdateCardModalComponent);
     console.log(this.paymentMethods);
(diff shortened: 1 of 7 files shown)


src/app/my-account/my-account.component.ts
```diff
@@ -84,13 +86,62 @@ export class MyAccountComponent implements OnInit {
...
 
-  updateCard(card: PaymentMethod.Card) {
+  updateCard(methodId: string, card: PaymentMethod.Card, billingDetails: PaymentMethod.BillingDetails) {
     const updateCardDialog = this.dialog.open(UpdateCardModalComponent);
     console.log(this.paymentMethods);
-    updateCardDialog.componentInstance.setCard(card);
+    updateCardDialog.componentInstance.setPaymentMethod(card, billingDetails);
     updateCardDialog.afterClosed().subscribe((data: unknown) => {
-      // todo - send card update to backend and update view.
+      if (data === "null") {
```
bdsl, line 94: not sure if this is the best way to check whether OK was pressed.

dee-jo (approved):

PR #1036, opened 2023-04-11
Don 704 split up donation start component

dee-jo:


bdsl:
@dee-jo can this now be closed if it's to be replaced by #1038 and following PRs?

PR #1042, opened 2023-04-13
Revert "DON-759 enforce captcha on every step during the donation journey"

dee-jo:
Reverts thebiggive/donate-frontend#1039


dee-jo:
Reverting as the change introduced a bug which makes the captcha reappear twice on consecutive steps

PR #1052, opened 2023-04-17
DON-732: Release edit card feature to production

bdsl:
I forgot to do this when DON-723 passed QA. Just went to check it in prod and saw that it wasn't released.


dee-jo (commented):
LGTM

PR #1054, opened 2023-04-18
DON-761 Updated text for Green match fund and swapped to first homepa…

BethBigGive:
…ge card


bdsl:
Thanks @BethBigGive . I've added a commit to change remove the exclamation mark. Since we want to add the "Not before noon on Thursday rule" I'll close this PR in favour of mine for the rest of the work at https://github.com/thebiggive/donate-frontend/pull/1055 

You still show up as contributor on that though: 
image

PR #1055, opened 2023-04-18
Don 761 gmf card update

bdsl:
Includes logic to keep the home page as it is for now, and then change to the below when the clock on the users' computer ticks past noon UK time on Thursday.

image

### DIFF
--- a/src/app/home/cardsForMetaCampaigns.ts
+++ b/src/app/home/cardsForMetaCampaigns.ts
@@ -32,7 +32,12 @@ const campaignToCard = function (metaCampaign: MetaCampaign, donateUriPrefix: st
  * There is an open question about whether this logic should live here, or in SF, or somewhere else - for now it seems
  * to make sense to put it here since we have limited SF development capacity.
  */
-export const cardsForMetaCampaigns = function (metacampaigns: readonly MetaCampaign[], donateUriPrefix: string, blogUriPrefix: string): readonly HighlightCard[]
+export const cardsForMetaCampaigns = function (
+  viewingDate: Date,
+  metacampaigns: readonly MetaCampaign[],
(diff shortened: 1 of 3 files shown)


src/app/home/cardsForMetaCampaigns.ts
```diff
@@ -53,6 +58,33 @@ export const cardsForMetaCampaigns = function (metacampaigns: readonly MetaCampa
...
   }
 
+  if (viewingDate >= new Date("2022-04-20T12:00:00+01:00")) {
+    return [
+      ...metaCampaignCards,
+      {
+        headerText: 'Double your donation in the Green Match Fund',
+        backgroundImageUrl: new URL('/assets/images/card-background-gmf.jpg', donateUriPrefix),
+        iconColor: 'brand-3',
+        bodyText: 'Donate between 20/04/23 - 27/04/23',
```
NoelLH, line 68: I think this would look better if the date format matched the next card. We might also be able to omit "23"?

bdsl: Agreed, and generally the numbers only format is probably best avoided since its ambiguous for the first 12 days of the month. Also to me "between" reads as redundant with "-". We could change it to the below. @BethBigGive what do you think?

```suggestion
        bodyText: 'Donate from 20 April to 27 April 2023',
```

dee-jo (approved):

src/app/home/cardsForMetaCampaigns.ts
BethBigGive: Happy to make that change @bdsl

bdsl: image

PR #1057, opened 2023-04-18
DON-762 – load credit balance on login

NoelLH:
I think we lost this while splitting / tidying `loadAuthedPersonInfo()`


bdsl (approved):
LTGTM, I see this is restoring code accidentally deleted in 13e4748.  Sorry for not spotting that!

PR #1065, opened 2023-04-20
DON-758 – Matomo CSP additions

NoelLH:
Allow our Matomo host for `connect` and `img` sources too, based on observed CSP errors and
https://matomo.org/faq/general/faq_20904/

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -25,6 +25,7 @@ const donationsApiHost = (new URL(environment.donationsApiPrefix)).host;
 const donateGlobalHost = (new URL(environment.donateGlobalUriPrefix)).host;
 const donateHost = (new URL(environment.donateUriPrefix)).host;
 const identityApiHost = (new URL(environment.identityApiPrefix)).host;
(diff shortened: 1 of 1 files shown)


server.ts
```diff
@@ -72,10 +74,12 @@ export function app() {
           `'self'`,
           'data:',
           'https:',
+          matomoHost,
```
bdsl, line 77: Is it right that the three items above are URI schemes but what you've added is a hostname? I'm not sure where the docs are on what's supposed to go in this list.

NoelLH: It looks like both are allowed. https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

I just went with host out of habit as it's documented that way more often / matches other places in this file. But obviously if we can be stricter about SSL we should, so I'll amend.

bdsl (approved):

server.ts
bdsl: Thanks, useful link.

PR #1066, opened 2023-04-20
modified angular.json to fix broken test runner

dee-jo:
https://stackoverflow.com/a/68163709


bdsl (approved):
LGTM, changing dev settings is safe enough, we can always change back if we don't like them.

PR #1072, opened 2023-04-25
Remove redundant `noNav()` event interceptor

NoelLH:
Pretty sure this is permanently redundant since the latest menu changes


bdsl (approved):
LGTM - I wonder if there's an easy auto check we could to have to disallow dead code (except were we mark it as intentionally dead)

PR #1075, opened 2023-04-25
DON-770: Update components verison.

bdsl:
Components changes: https://github.com/thebiggive/components/compare/b6b29...4e03bd2


bdsl:
Will merge without review when checks pass.

PR #1076, opened 2023-04-26
DON-771 Replace GMF card with K2M card

BethBigGive:
- Brand icon colour still needs to be updated #62CFC9

- Card should be updated after noon Thursday

image


bdsl:
I've pushed some updates to this to change the colour and to get the automatic tests to pass. But the icon is showing grey not purple:

image

Looks like the issue is that there's some contradiction in our components library - in [one place](https://github.com/thebiggive/components/blob/8f348fa7260a57bf8713d0c2456bf73fbdc753ef/src/globals/variables.scss#L38) `brand-6` is defined as turquoise, but [somewhere else](https://github.com/thebiggive/components/blob/0445a47388d9691a3e5acb545ae8ed5e77c52df0/src/globals/mixins.scss#L1085) it's defined as grey.

Going to push another update to make the icon purple without having to edit the components library if I can.

PR #1071, opened 2023-04-25
Don 704 move secondary column into separate component

dee-jo:

PR #1077, opened 2023-04-27
Revert "Don 704 move secondary column into separate component"

dee-jo:
Reverts thebiggive/donate-frontend#1071


bdsl:
I assume this can be closed now.

PR #1079, opened 2023-04-27
Don 704: Separate donation start header and use donation start login from donation start form

bdsl:
I'm nibbling away at the top of `donation-start-primary.component.html` with a plan to leave nothing but the form in there.

### DIFF
--- a/src/app/donation-start/donation-start-container/donation-start-container.component.html
+++ b/src/app/donation-start/donation-start-container/donation-start-container.component.html
@@ -13,8 +13,22 @@
         <h2 class="c-header__to b-grey b-light">For</h2>
         <h1 class="b-rh-2 b-bold">{{ campaign.title }}</h1>
       </div>
+
+      <app-donation-start-login
+        [logout]="this.logout"
+        [personId]="this.personId"
+        [email]="this.loggedInEmailAddress"
+        [personIsLoginReady]="personIsLoginReady"
+        [loadAuthedPersonInfo]="loadAuthedPersonInfo"
+        [creditPenceToUse]="this.donationStartForm && this.donationStartForm.creditPenceToUse || 0"
+        [campaign]="campaign"
+        [canLogin]="this.canLogin">
+      </app-donation-start-login>
+
       <app-donation-start-primary
+        #donation_start_form
       [campaign]="campaign"
+      [personId]="personId"
     />
     </div>
     <div class="b-secondary-column">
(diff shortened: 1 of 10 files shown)


src/app/donation-start/donation-start-container/donation-start-container.component.html
```diff
@@ -7,8 +7,28 @@
...
+        [email]="this.loggedInEmailAddress"
+        [personIsLoginReady]="personIsLoginReady"
+        [loadAuthedPersonInfo]="loadAuthedPersonInfo"
+        [creditPenceToUse]="this.donationStartForm && this.donationStartForm.creditPenceToUse || 0"
+        [campaign]="campaign"
+        [canLogin]="this.canLogin">
+      </app-donation-start-login>
+
       <app-donation-start-primary
+        #donation_start_form
```
bdsl, line 29: not sure if this ID is required for the @ViewChild or if we could select it more simply by the component name.

src/app/donation-start/donation-start-container/donation-start-container.component.ts
```diff
@@ -1,21 +1,65 @@
...
+  }
+
+  loadAuthedPersonInfo = (id: string, jwt: string) => {
+    if (!this.identityService) {
+      // This feels like an anti-pattern, but currently seems to be required. Since the "contained"
+      // login component is passed this public fn and could call it any time, it is not safe to assume
+      // that this page has its normal service dependencies. The current behaviour post-login seems to
+      // be that this is called as a no-op once, but then there's a reload during which it works?
+
+      return;
```
bdsl, line 51: took me far too long to work out why it was always returning here - eventually realised it was because I didn't have it as an arrow function so it wasn't bound to the object. Now fixed.

src/app/donation-start/donation-start-container/donation-start-container.component.html
NoelLH: I'm not an expert but think I've only seen it done this way or with an Angular class that defines the component. I think we'd probably want to look into the performance of looking it up in a DOM-based way.

NoelLH (commented):
Looking good!

bdsl:
> I'm not an expert but think I've only seen it done this way or with an Angular class that defines the component. I think we'd probably want to look into the performance of looking it up in a DOM-based way.

OK, let's leave that as it is.

NoelLH (approved):

PR #1085, opened 2023-05-02
Various domain updates

NoelLH:


src/app/campaign.resolver.ts
```diff
@@ -40,7 +40,7 @@ export class CampaignResolver implements Resolve<any> {
 
     if (campaignSlug && fundSlug && campaignSlug !== "campaign") {
       const query = this.campaignService.buildQuery(this.searchService.selected, 0, campaignId ?? undefined, campaignSlug, fundSlug);
-      const searchResult = this.campaignService.search(query as SearchQuery).subscribe(
+      this.campaignService.search(query as SearchQuery).subscribe(
```
bdsl, line 43: Nice - maybe would be worth enabling the `@typescript-eslint/no-unused-vars` eslint rule to do this more generally. Could just extend "plugin:@typescript-eslint/recommended", but that finds a lot more errors and I'm not sure how to set a baseline.

image

With that recommended setup it finds 191 problems (42 errors, 149 warnings)
28 warnings just for no-unused-vars. But actually I'm not really a fan of warnings, think its better to either have things error or ignored.

bdsl (approved):

src/app/campaign.resolver.ts
bdsl: See https://github.com/thebiggive/donate-frontend/pull/1086

PR #1086, opened 2023-05-02
Ban & remove unused vars

bdsl:


src/app/campaign-list.resolver.ts
```diff
@@ -7,6 +7,7 @@ import { CampaignService } from './campaign.service';
 export class CampaignListResolver implements Resolve<any> {
   constructor(private campaignService: CampaignService) {}
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
bdsl, line 10: Generally not removing function params since we may want to force callers to pass them even if we don't use them for now. In PHP you can start the variable or param name with a _ to show its not accidentally unused, but doesn't seem to work here.

PR #1088, opened 2023-05-02
DON-656 – Search + sort fixes

NoelLH:


src/app/explore/explore.component.html
```diff
@@ -15,7 +15,7 @@
       [beneficiaryOptions]="beneficiaryOptions"
       [locationOptions]="locationOptions"
       [fundingOptions]="fundingOptions"
-      [selectedSortByOption]="searchService.getSelectedSortLabel()"
+      [selectedSortByOption]="searchService.selectedSortLabel"
```
bdsl, line 18: I'm confused by this - it looks like the prop is typed as `'Most raised' | 'Match funds remaining' | 'Relevance' | null` but you're now passing a function rather than a string or null - how does that work?
https://github.com/thebiggive/components/blob/7aa4dfad27472c6aa070f6ed90148e7adfd77f26/src/components/biggive-campaign-card-filter-grid/biggive-campaign-card-filter-grid.tsx#L95

bdsl, line 18: The prop type is `'Most raised' | 'Match funds remaining' | 'Relevance' | null` but you're passing a `string` which shouldn't be assignable - looks like we don't have SA working between the template and the prop type, but would be good if we could get them matching up.

bdsl: I've made a PR to your PR: https://github.com/thebiggive/donate-frontend/pull/1089

bdsl (commented):
LGTM other than where I've commented

src/app/explore/explore.component.html
NoelLH: @bdsl Sure. Not had time to test it in context but I've made a `components` change to make the prop non-nullable: https://github.com/thebiggive/components/pull/164

The types of the lib are checked independently and it should be safe to assign the narrower type to the wider one afaik. So I think these could probably be merged in either order / without necessarily waiting for a new `components` version.

bdsl: In theory narrow types can be assigned to wide, not vise versa. So your PR 164 should be a BC break - but I don't think it matters too much while these aren't being checked and yes they can be merged in either order. Really 1089 should be changed to remove the 'null' part from it now.

PR #1089, opened 2023-05-02
DON-656: Match types between search SearchService.selectedSortLabel a…

bdsl:
…nd BiggiveCampaignCardFilterGrid.selectedSortByOption

The previous version seems like it should have been a compile error, but the compiler doesn't seem to check this

### DIFF
--- a/src/app/search.service.ts
+++ b/src/app/search.service.ts
@@ -18,7 +18,7 @@ export class SearchService {
   changed: EventEmitter<boolean>; // Value indicates if an interactive UI change triggered this.
 
   nonDefaultsActive: boolean;
-  selectedSortLabel: string;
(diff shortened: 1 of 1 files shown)


src/app/search.service.ts
```diff
@@ -18,7 +18,7 @@ export class SearchService {
   changed: EventEmitter<boolean>; // Value indicates if an interactive UI change triggered this.
 
   nonDefaultsActive: boolean;
-  selectedSortLabel: string;
+  selectedSortLabel: 'Most raised' | 'Match funds remaining' | 'Relevance' | null;
```
NoelLH, line 21:
```suggestion
  selectedSortLabel: 'Most raised' | 'Match funds remaining' | 'Relevance';
```
Per https://github.com/thebiggive/donate-frontend/pull/1088#discussion_r1182685483

NoelLH (approved):

PR #1090, opened 2023-05-02
DON-744: Extract updateDonationFromForm and sanitiseCurrency function…

bdsl:
…s from Donation Start Form component

### DIFF
--- a/src/app/donation-start/updateDonationFromForm.ts
+++ b/src/app/donation-start/updateDonationFromForm.ts
@@ -1,0 +1,65 @@
+import {StepperSelectionEvent} from "@angular/cdk/stepper";
+import {AnalyticsService} from "../analytics.service";
+import {FormGroup} from "@angular/forms";
+import {DonationService} from "../donation.service";
+import {Donation} from "../donation.model";
+import {Campaign} from "../campaign.model";
+import {sanitiseCurrency} from "./sanitiseCurrency";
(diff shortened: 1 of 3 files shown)


bdsl:
We can safely ignore the  Incomplete string escaping or encoding alert - the code its complaining about is not new or really edited here, just moved. Could unit test that `sanitiseCurrency` and rewrite it in a way that won't generate an alert, but not sure its necassary now.

src/app/donation-start/updateDonationFromForm.ts
```diff
@@ -0,0 +1,65 @@
+import {StepperSelectionEvent} from "@angular/cdk/stepper";
+import {AnalyticsService} from "../analytics.service";
+import {FormGroup} from "@angular/forms";
+import {DonationService} from "../donation.service";
+import {Donation} from "../donation.model";
+import {Campaign} from "../campaign.model";
+import {sanitiseCurrency} from "./sanitiseCurrency";
+
+
+export function updateDonationFromForm
```
bdsl, line 10: There are actually no tests that cover this function - returning void at the first line doesn't make anything fail. So no tests to move out of the component spec file.

dee-jo (approved):
LGTM, compared with the develop branch (in particular the updateDonationFromForm(...)) and the code is equivalent

PR #1092, opened 2023-05-03
DON-774: Move tracking code out of donation start component into new …

bdsl:
…service

I've tested locally just to make sure that Angular does inject the new service (and the existing services are injected into it) and the function can be called.

I think after this it maybe time to call DON-774 done. There's always more splitting we could do, but for now we could probably move on to work on other tasks and just do any more splitting as and when we need to.


dee-jo (approved):
LGTM (pulled branch and compared code statically)

PR #1094, opened 2023-05-03
BG2-2342: Cancel pending donation if logged in user has a different P…

bdsl:
…SP ID

It isn't possible for a logged in user to pay for a donation created for a different user account (which would normally be an anon account), so if someone logs in after starting a donation we cancel their existing donation and let them start again.

### DIFF
--- a/src/app/donation-start/donation-start-container/donation-start-container.component.html
+++ b/src/app/donation-start/donation-start-container/donation-start-container.component.html
@@ -32,7 +32,9 @@
         <span class="b-rt-1 b-bold">{{ reservationExpiryDate | timeLeft }}</span>
       </p>
 
-      <app-donation-start-form *ngIf="campaignOpenOnLoad"
+      <!-- We don't want to let the form's ngInit run before ours completes, or we would have odd behaviour
+           around loading the logged in user - hence conditional on dataLoaded -->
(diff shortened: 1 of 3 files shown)


src/app/donation-start/donation-start-container/donation-start-container.component.html
```diff
@@ -32,10 +32,12 @@ <h1 class="b-rh-2 b-bold">{{ campaign.title }}</h1>
         <span class="b-rt-1 b-bold">{{ reservationExpiryDate | timeLeft }}</span>
       </p>
 
-      <app-donation-start-form *ngIf="campaignOpenOnLoad"
+      <!-- We don't want to let the form's ngInit run before ours completes, or we would have odd behaviour
+           around loading the logged in user - hence conditional on dataLoaded -->
+      <app-donation-start-form *ngIf="dataLoaded && campaignOpenOnLoad"
```
bdsl, line 37: I'm not actually 100% sure if this change is needed - yesterday it seemed like it was needed, in my testing now on my local I wasn't able to keep reproduce the bug just by changing this line back. But I think its worth keeping anyway, easier to see what's going on if the form doesn't load until the onnginit from the container is finished.

dee-jo (approved):
LGTM, tested locally and see no error when navigating back to the campaign page or logging in after starting a donation, donation appears to be cleared as I'm not being offered to resume when I return back to donate

PR #1096, opened 2023-05-04
DON-764: Use separate instances of TimeLeftPipe for opening and closi…

bdsl:
…ng campaign dates

When we share one instance for two purposes the dates can get mixed up


dee-jo (approved):
LGTM

PR #1102, opened 2023-05-05
DON-610: updated design for 1st step of donation stepper form

dee-jo:
WIP, this PR has issues that need to be fixed:
- if the user enters a custom tip value and then goes back to percentage tip, the custom value should be applied on the slider and the handle position adjusted. In case it value fell out of slider's percentage range, 'back to percentage tip' link should be hidden
- the slider also has a bug, which makes the slider's handle jump too far on click

Screenshot from 2023-05-25 10-58-35

Screenshot from 2023-05-25 10-58-44

Screenshot from 2023-05-25 10-59-05


src/app/donation-start/donation-start-container/donation-start-container.module.ts
```diff
@@ -44,6 +46,8 @@ import {DonationStartFormParentComponent} from "../donation-start-form/donation-
     MatRadioModule,
     MatSelectModule,
     MatStepperModule,
+    MatExpansionModule,
+    MatDividerModule,
```
bdsl, line 50: Are these actually used? I'm guessing maybe not as there are no other changes in the donation-start-container folder.

src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -59,8 +62,8 @@
           </mat-checkbox>
         </div>
 
-        <div *ngIf="!campaign.feePercentage && creditPenceToUse === 0">
-          <mat-form-field appearance="outline">
+      <div *ngIf="!campaign.feePercentage && creditPenceToUse === 0" style="text-align: center; margin-bottom: 50px;">
+            <!-- <mat-form-field appearance="outline">
```
bdsl, line 66:
```suggestion
            <!-- We may want to bring something like the below back to test a/b test against the slider, although using our
                   custom select component instead of mat-select.
            <mat-form-field appearance="outline">
```

src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -69,21 +72,17 @@
...
-                <div class="floating-label">
-                  Donation to the Big Give Trust
-                </div>
-              </mat-label>
-              <input formControlName="tipAmount" id="tipAmount" matInput aria-describedby="custom-tip-status">
-            </mat-form-field>
+          </mat-form-field>-->
+
+          <div *ngIf="showCustomTipInput" class="donation-input donation-input-main">
+            <!-- <p id="custom-tip-status" class="b-grey b-rt-sm donation-input-tip__status-help" aria-live="assertive">You've selected to give {{ tipAmount() | exactCurrency:campaign.currencyCode }} to the Big Give Trust. You can edit the amount in the next field.</p> -->
```
bdsl, line 78: No reason to keep this, is there?

```suggestion```

Comment (bdsl) [inline src/app/donation-start/donation-start-form/donation-start-form-new.component.html:78]: There's no reason to keep this, is there?

```suggestion
```

src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -102,6 +101,53 @@
           of main donation
           </mat-hint>
         </div>
+        <app-donation-tipping-slider
+            *ngIf="!showCustomTipInput"
+            #donationTippingSlider
+            [percentageStart]="1"
+            [percentageEnd]="maximumTipPercentage"
+            [donationAmount]="donationAmount"
```
bdsl, line 109: Shouldn't this be something like

```suggestion
            [donationAmount]="tipAmount()"
```

Within the tip slider I think donation amount refers to the amount they're tipping, not the amount they're donating to the campaigning charity.

src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -102,6 +101,53 @@
...
+              donations to good causes.
+            </li>
+            <li>Providing match funds in our Anchor Match Fund. Through tips, charities will receive funding to launch their own match funded campaign on Big Give.
+            </li>
+          </ul>
+
+  <p>
+  Thank you for your generosity! Your tips ensure that we can continue to support charities and causes that are important to you.
+
+          </p>
```
bdsl, line 148: Is this paragraph meant to be here? Feels like its repeating the stuff above, could have been left in by mistake?

src/app/donation-start/donation-start-form/donation-start-form-new.component.ts
```diff
@@ -15,4 +16,12 @@ import { DonationStartFormParentComponent } from "./donation-start-form-parent.c
...
 export class DonationStartFormNewComponent extends DonationStartFormParentComponent {
 
-}
+  panelOpenState = false;
+  percentage = 1;
+  showCustomTipInput = false;
+  @ViewChild('tippingSlider') tippingSlider: DonationTippingSliderComponent;
+
+  toggleTipInput = () => {
+    this.showCustomTipInput = this.showCustomTipInput === true ? false : true;
```
bdsl, line 25:
```suggestion
    this.showCustomTipInput = !this.showCustomTipInput;
```

src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -160,8 +163,8 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
   private stepHeaderEventsSet = false;
   private tipPercentageChanged = false;
 
-  private initialTipSuggestedPercentage = 15;
-
+  tipPercentage = 12.5;
```
bdsl, line 166: why has this changed from 15 to 12.5?

src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -837,7 +840,8 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
   }
 
   tipAmount(): number {
-    return sanitiseCurrency(this.amountsGroup.value.tipAmount);
+    //return sanitiseCurrency(this.amountsGroup.value.tipAmount);
```
bdsl, line 843:
```suggestion
```

src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -1571,13 +1575,13 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
...
           tipPercentage?: number | string,
-          tipAmount?: string,
+          tipAmount?: string
         } = {};
 
         donationAmount = sanitiseCurrency(donationAmount);
 
         if (!this.tipPercentageChanged) {
-          let newDefault = this.initialTipSuggestedPercentage;
+          let newDefault = this.tipPercentage;
```
bdsl, line 1584: not sure what this change is about

src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -1858,4 +1862,13 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
     this.prefillRarelyChangingFormValuesFromPerson(person);
     this.loadFirstSavedStripeCardIfAny(id, jwt);
   }
+
+  onDonationSliderMove = (tipPercentage: number, tipAmount: number) => {
+    this.tipPercentage = tipPercentage;
```
bdsl, line 1867: maybe something to note on Jira and fix later rather than here - but I'm slightly concerned that taking the percentage means we don't have the rounding that's done for display inside the slider, so the amount they actually pay could be a few pence different to what they expected. We should round once and then use that output for display and for actually taking the money.

I think maybe we should get rid of the tipPercentage field in the class and make everything work off the tipAmount.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.scss
```diff
@@ -0,0 +1,68 @@
+// @include backgrounds();
+// @include fills();
+
```
bdsl, line 3:
```suggestion
```

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.scss
```diff
@@ -0,0 +1,68 @@
...
+:host {
+  display: contents;
+//   @include standard-font();
```
bdsl, line 6:
```suggestion
```

src/app/donation-start/donation-start-form/donation-start-form-new.component.html
bdsl: ah no I was wrong, ignore me.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+  | 'grey-light'
+  | 'grey-medium'
+  | 'grey-dark' = 'primary';
+
+  @Input() percentageCurrent: number;
+  @Input() percentageStart: number;
+  @Input() percentageEnd: number;
+  @Input() donationAmount: number;
+  //ISO-4217 currency code (e.g. GBP, USD)
+  @Input() donationCurrency!: 'GBP' | 'USD';
```
bdsl, line 35: Would love to get stronger types and make this more like:

```suggestion
  @Input() percentageCurrent: percentage;
  @Input() percentageStart: percentage;
  @Input() percentageEnd: percentage;
  @Input() donationAmount: donationAmount;
```

Percentages would be constrained to between zero and 100, donation amount would be constrained to between 1 and 25,000 and would encapsulate the currency and the number together. But definitely not for today!

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+  @Input() onHandleMoved: (tipPercentage: number, tipAmount: number) => void;
+
+  @ViewChild('handle', {static: true}) handle: ElementRef;
+  @ViewChild('bar', {static: true}) bar: ElementRef;
+  @ViewChild('percentageValue', {static: true}) percentageWrap: ElementRef;
+  @ViewChild('donationValue', {static: true}) donationWrap: ElementRef;
+
```
bdsl, line 42:
```suggestion
```

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  containerClass: string;
+  derivedPercentage: number;
```
bdsl, line 45: not sure what the difference between percentageCurrent and derivedPercentage is - maybe could be named better, maybe we don't need both properties.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+  @Input() donationCurrency!: 'GBP' | 'USD';
+  @Input() onHandleMoved: (tipPercentage: number, tipAmount: number) => void;
+
+  @ViewChild('handle', {static: true}) handle: ElementRef;
+  @ViewChild('bar', {static: true}) bar: ElementRef;
+  @ViewChild('percentageValue', {static: true}) percentageWrap: ElementRef;
+  @ViewChild('donationValue', {static: true}) donationWrap: ElementRef;
+
+
+  containerClass: string;
```
bdsl, line 44: we can just hard-code the space below and lose this property - I don't think there's any reason to expect to need this component in two places any time soon.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+  derivedPercentage: number;
+  tipAmount: number;
+  currencyFormatted: string;
+
+  documentMouseupUnlistener: () => void;
+  documentMousemoveUnlistener: () => void;
+  documentTouchmoveUnlistener: () => void;
```
bdsl, line 51: not sure what these are about - I see they're only invoked in ngOnDestroy.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  isMoving = false;
+  max: number;
+  pageX: any;
+  mousePos: number;
+  position: number;
```
bdsl, line 57: maybe clearer if we rename max and position to something like `maxPosition` and `currentPosition` and/or add docblocks to explain what they are?

Also not sure what pageX is, and would be great to have any type that isn't `any` on it.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+  };
+
+
+  ngOnInit() {
+    this.containerClass = 'container space-below-' + this.spaceBelow;
+    this.calcAndSetPercentage();
+    this.adjustDonationPercentageAndValue();
+    this.updateHandlePosition(undefined)
+  }
+
```
bdsl, line 83: What's the reasoning for doing some stuff OnInit and waiting for AfterContentInit for other stuff? I guess the nativeElements aren't ready to have event listeners OnInit?

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+      this.bar.nativeElement.addEventListener('touchend', () => {
+        this.isMoving = false;
+      });
+    }
+  };
+
+  // detect changes in the donationAmount input
+  ngOnChanges(changed: any) {
+    this.max = this.bar.nativeElement.offsetWidth - this.handle.nativeElement.offsetWidth;
+    if (changed.donationAmount.currentValue != changed.donationAmount.previousValue) {
```
bdsl, line 105: I've still got my `any` allergy - and the comment seems redundant.

```suggestion
  ngOnChanges(changed: SimpleChanges) {
    this.max = this.bar.nativeElement.offsetWidth - this.handle.nativeElement.offsetWidth;
    if (changed.donationAmount!.currentValue != changed.donationAmount!.previousValue) {
```

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+      this.calcAndSetPercentage();
+      this.calcAndSetTipAmount();
+      this.adjustDonationPercentageAndValue();
```
bdsl, line 108: Looks like these three a similar, maybe could have one function that calls them all in turn from here. Particularly if we're going to add unit tests for the business logic.

Feels like there's a bit of tension between whether we're doing reactive, ZoneJS based programming, or more traditional OOP programming with explicit function calls and arguments - we've got the function calls but with arguments. I'm not necessarily familiar enough with angular and TS to know if it's better to go one way or the other.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+        this.derivedPercentage = 7.5;
+      } else if (this.donationAmount >= 300) {
+        this.derivedPercentage = 10;
+      } else if (this.donationAmount >= 100) {
+        this.derivedPercentage = 12.5;
+      } else {
+        this.derivedPercentage = 15;
+      }
+    }
+  }
```
bdsl, line 140: logic above definitely looks worth testing. Did we do the thing where we don't set the default percentage after someone choose moves the slider?

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+          this.adjustDonationPercentageAndValue();
+          this.onHandleMoved(this.derivedPercentage, this.tipAmount);
+        }
+      }
+    }
+
+  }
+
+  calcAndSetTipAmount() {
+    if (this.donationAmount < 55) {
```
bdsl, line 175: what's special about the number 55? Maybe give it a name.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+      this.tipAmount = this.donationAmount * (this.derivedPercentage / 100);
+    } else {
+      this.tipAmount = Math.round(this.donationAmount * (this.derivedPercentage / 100));
```
bdsl, line 178: maybe getting nitpicky but better not to repeat the calculation code? Could just calculate it outside the conditional and then conditionally round?

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+
+  updateHandlePosition(e: MouseEvent | TouchEvent | undefined) {
+    if(!e) { // the handle position is driven by the donation input change
+      this.calcAndSetPercentage();
+      this.pageX = 122;
```
bdsl, line 185: What is 122?

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  calcAndSetTipAmount() {
+    if (this.donationAmount < 55) {
+      this.tipAmount = this.donationAmount * (this.derivedPercentage / 100);
+    } else {
+      this.tipAmount = Math.round(this.donationAmount * (this.derivedPercentage / 100));
+    }
+  }
+
+  updateHandlePosition(e: MouseEvent | TouchEvent | undefined) {
```
bdsl, line 182: When we call this method we always know whether e is defined or undefined I think - and then 90% of the body of the method is different accordingly. So why not have two separate methods - one takes an event argument that has to be defined, and one doesn't take any argument.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+    if(!e) { // the handle position is driven by the donation input change
+      this.calcAndSetPercentage();
+      this.pageX = 122;
+      this.mousePos = this.pageX - this.bar.nativeElement.offsetLeft - this.handle.nativeElement.offsetWidth / 2;
```
bdsl, line 186: `mousePos` is never used outside this function, is it? Can it be scoped to the function instead of a class property?

Also here in the `!e` case it seems like mousePos is written and never read.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -0,0 +1,214 @@
...
+      this.calcAndSetPercentage();
+    }
+
+    this.handle.nativeElement.style.marginLeft = this.position + 'px';
+  }
+
+  adjustDonationPercentageAndValue() {
+    this.percentageWrap.nativeElement.innerHTML = this.derivedPercentage.toString();
+    this.currencyFormatted = this.format(this.donationCurrency, this.tipAmount);
+    this.donationWrap.nativeElement.innerHTML = this.currencyFormatted;
```
bdsl, line 201: can we set `.text` instead of `.innerHtml`, since neither the percentage or the currency are HTML snippets?

And I guess we don't need to update the currencyFormatted do we?

bdsl:
>     if the user enters a custom tip value and then goes back to percentage tip, the custom value should be applied on the slider and the handle position adjusted. In case it value fell out of slider's percentage range, 'back to percentage tip' link should be hidden
>     the slider also has a bug, which makes the slider's handle jump too far on click

I don't think these are blockers, but before merging let's make sure they're captured somewhere we'll see them - either on the DON-610, or comments in the code, or maybe as individual Jira tickets since 610 is getting unwieldy.

src/app/donation-start/donation-start-container/donation-start-container.module.ts
dee-jo: I think we still need the Expansion Module, will delete the Devider Module

src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -102,6 +101,53 @@
           of main donation
           </mat-hint>
         </div>
+        <app-donation-tipping-slider
+            *ngIf="!showCustomTipInput"
+            #donationTippingSlider
+            [percentageStart]="1"
+            [percentageEnd]="maximumTipPercentage"
+            [donationAmount]="donationAmount"
+            donationCurrency="GBP"
```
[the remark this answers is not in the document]
dee-jo: Good shout, could use the one on the campaign object

src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
dee-jo: TODO: needs to be reverted to 12.5

bdsl: Changed back to 15

bdsl (approved):
Approving as this won't break the existing page, and it's more efficient to fix the remaining issues on develop branch. See Jira

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
bdsl: see https://github.com/thebiggive/donate-frontend/pull/1133

bdsl:
Logged issues: https://thebiggive.atlassian.net/issues/?jql=issueKey%20in%20(BG2-2359%2CBG2-2360)

PR #1107, opened 2023-05-11
DON-777/776: Update front-end homepage cards

BethBigGive:
I have updated the cards; only additional change needs to be the background card image for Christmas Challenge. Cards should go live on 15th May


bdsl:
The Jira didn't say what the Christmas background should be but I took a look at https://github.com/thebiggive/donate-frontend/tree/develop/src/assets/images and found https://github.com/thebiggive/donate-frontend/blob/develop/src/assets/images/card-background-cc.jpg so I've set that. Now appears as below. @BethBigGive let me know if it's what you want.

image

bdsl (dismissed):
LGTM, just a query about the timing. If it needs precise timing I can add code for that, if not let's not merge this until Friday afternoon ready to deploy on Monday.

BethBigGive:
- We want the 'donate now' for K2M to appear just before 12pm (midday) on Monday, happy to go with your suggestions.
- For the banner, can we have the Christmas Challenge image linked in this ticket: https://thebiggive.atlassian.net/browse/DON-735

bdsl:
OK I think we can just push it to live manually around 11:40am, might appear 5 or 10 minutes early.

I'll add the image - looks like its an image we never used before so will download from Jira.

bdsl:
Pushed an update, looks like so now:

image

dee-jo (approved):
LGTM

bdsl:
Thanks @dee-jo Let's merge this tomorrow afternoon and deploy on Monday.

bdsl:
Merging without status checks because the automatic build & check doesn't run as this is a PR from outside the repo. At some point it might be worth either giving @BethBigGive write access to make PRs from within the main repo, or setting up Circle to run builds on her copy of the repo.

PR #1118, opened 2023-05-18
DON-610: WIP use new drop-down component for billing country in donat…

bdsl:
…ion start form

Still need to fix:

- Background colour inside should be grey not white
- Initial value (GB by default) should load automatically
- Actually taking the value selected
- Use whole column width

Currently appears like so (initially, during and after selection):
image
image
image


bdsl:
Closing - as discussed on Slack we need to substantially change the component before it's good enough to use.

PR #1119, opened 2023-05-18
DON-610 update donation stepper design / step 5

dorota-joanna:
image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
@@ -555,7 +555,6 @@
             <tr>
               <td colspan="2"><hr></td>
             </tr>
-            <!-- TODO: think about to include matchfunding in the total -->
             <tr class="total">
               <td><div><strong>Total amount you will be charged</strong></div></td>
               <td><strong>{{ donationAndExtrasAmount | exactCurrency:donation.currencyCode }}</strong></td>


src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -476,45 +477,67 @@
...
+            <tr *ngIf="donation.matchedAmount > 0">
+              <td>Matched Amount</td>
+              <td>{{ expectedMatchAmount() | exactCurrency:donation.currencyCode }}</td>
+            </tr>
+            <tr *ngIf="giftAidAmount() > 0">
+              <td>Gift Aid Amount</td>
+              <td>{{ giftAidAmount | exactCurrency:donation.currencyCode }}</td>
+            </tr>
+            <tr *ngIf="feeCoverAmount() > 0 || (creditPenceToUse === 0 && !campaign.feePercentage)">
+              <td>Platform fee</td>
```
bdsl, line 496: Noticing that the screenshot shoews platform fee £0 - I don't think we want it to show platform fee with any current campaigns, not sure why it's showing up. This logic I think is copied from what we had already, so not sure exacltly what's going on - or why the second part of the condition exists.

NoelLH: Yeah, this definitely looks wrong. It looks like a previous containing element probably had the full condition clause in its `ngIf` and it got moved to the wrong place, as it looks like a combination of what should be
`[the case where fee shows] || [the case where tip shows]`

bdsl: Right. Fixed in 4fc0a17c .

src/app/donation-start/donation-start-form/donation-start-form-new.component.scss
```diff
@@ -402,3 +402,45 @@ div.payment-details > * {
...
+        font-size: 1.4rem;
+      }
+    }
+  }
+}
+//hr {
+//  margin: 0;
+//  width: 100%;
+//}
+.receives {
```
bdsl, line 438:
```suggestion
.receives {
```

src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -521,45 +522,65 @@ <h3 class="b-rt-0 b-m-0 b-bold">How does Big Give use tips?</h3>
...
+              *ngIf="creditPenceToUse === 0 && !campaign.feePercentage"
+            >
+              <td>Tip Amount</td>
+              <td>{{ donation.tipAmount | exactCurrency:donation.currencyCode }}</td>
+            </tr>
+
+            <tr>
+              <td colspan="2"><hr></td>
+            </tr>
+            <!-- TODO: think about to include matchfunding in the total -->
```
NoelLH, line 558: I don't think I agree with this _todo_, as IMO it is very likely we would get donors nervous about being overcharged.

NoelLH, line 558:
```suggestion
```

bdsl: I think I put this in because I was looking at the thanks page, idea was to align the two better. Thinking now maybe we'd even want them to share code if we want them really aligned. Not sure.

This is the thanks page we have now:

image

PR #1122, opened 2023-05-19
DON-782 – Matomo: fix line item ecommerce amounts not set

NoelLH:
Despite the library docs, currently an undefined category makes it skip setting the value of a line item


bdsl (approved):
LGTM - notes on issue tracker were helpful, thanks.

PR #1123, opened 2023-05-19
Updating hero component wording

alexbiggive:


alexbiggive:
Just testing pull requests with Barney.

PR #1124, opened 2023-05-23
DON-760: fixed layout issues on reset-password page

dorota-joanna:
Before:
Screenshot from 2023-05-23 13-41-01

After:
Screenshot from 2023-05-23 13-12-43

To test locally:
- comment out the spinner on lines 12-17 in html template
- comment out the *ngIf condition on line 21 in html
- comment out lines 55-70 in .ts to remove token check

### DIFF
--- a/src/app/reset-password/reset-password.component.scss
+++ b/src/app/reset-password/reset-password.component.scss
@@ -5,3 +5,15 @@
     margin-bottom: 20px;
     color: mat.get-color-from-palette($donate-warn);
 }
+
+main {
+    display: flex;
+    flex-direction: column;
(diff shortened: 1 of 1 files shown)


NoelLH (commented):
Nice, this looks way better!

My only question isn't actually about your change, but connects to my assumption (noted on Jira) that we were missing a module import. The existing HTML & TypeScript use/import `<mat-dialog-content/>`, but this seems to actually just be a standalone page and doesn't seem to be opened in a Dialog. So it feels like we'd maybe be better off removing that container and the module to avoid confusion?

bdsl (approved):

PR #1126, opened 2023-05-23
Don 610: country drop down restyle

bdsl:
Note the "Billing country" prompt is in the wrong position below, but this will be fixed by pulling future a components version that includes https://github.com/thebiggive/components/pull/179 . It should be just above "United Kingdom"

image


src/app/donation-start/donation-start-form/donation-start-form-new.component.ts
```diff
@@ -15,4 +15,5 @@ import { DonationStartFormParentComponent } from "./donation-start-form-parent.c
 })
 export class DonationStartFormNewComponent extends DonationStartFormParentComponent {
 
+    protected readonly Object = Object;
```
bdsl, line 18:
```suggestion
```

src/app/donation-start/donation-start-form/donation-start-form-new.component.scss
```diff
@@ -366,18 +366,10 @@ div.binary-option {
...
   margin-bottom: 2.5rem;
 }
 
 div.payment-details > * {
-  // 35% is > 1/3 to ensure we can't have 3 in a row, well under 1/2 so we can have two with gap between
-  flex: 1 0 35%;
-}
-
-div.payment-details > .sr-combo-inputs-row {
-  flex-basis: 100%; // contains space for card logo, PAN, Expiry & CVV, half the column width is not enough.
```
bdsl, line 382: No need for the complexity of flex when we're just putting one div below another.

src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -14,43 +14,49 @@ import {
...
-import { getCurrencyMaxValidator } from '../../validators/currency-max';
-import { getCurrencyMinValidator } from '../../validators/currency-min';
-import { EMAIL_REGEXP } from '../../validators/patterns';
-import { ValidateBillingPostCode } from '../../validators/validate-billing-post-code';
+import {FormBuilder, FormGroup, Validators} from '@angular/forms';
+import {MatAutocompleteSelectedEvent} from '@angular/material/autocomplete';
+import {MatCheckboxChange} from '@angular/material/checkbox';
+import {MatDialog} from '@angular/material/dialog';
+import {MatStepper} from '@angular/material/stepper';
+import {ActivatedRoute, Router} from '@angular/router';
```
bdsl, line 22: There are a few formatting changes in this file, most disappear if you select Hide whitespace in Github, the other is just adding line breaks in the imports from stripe-js. I think they're all fine to leave, probably were made automatically by PHP Storm. If the old version is better I can remove those changes and just keep the relevant stuff about countries.

src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -172,6 +179,7 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
   private stepChangedBlockedByCaptcha = false;
   @Input() donor: Person | undefined;
 
+  public countryOptionsObject: { [p: string]: string };
```
dee-jo, line 182: Note to self: change p to more meaningful name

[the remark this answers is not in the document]
bdsl: Was going to rename, then switched to use the [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) type and added a docblock in c249a55e88081ccfe180b8c7574741ea362e121c.

PR #1128, opened 2023-05-24
Fix "finish password reset" page being sort-of-standalone

NoelLH:
* Define modules consistently in one place
* Do routing the same as other pages
* Remove redundant `MatDialog` container stuff


NoelLH:
Following https://github.com/thebiggive/donate-frontend/pull/1124#pullrequestreview-1439800020 – I realised that there are other ways in which this page's code was an inconsistent mix of a standalone dialog and a page with its own module. This is an attempt to make it the latter, and more consistent with other pages to which we route in this way.

bdsl (approved):

PR #1131, opened 2023-05-25
DON-775: fixed part one of the donation journey bug

dee-jo:
Completed part 1 of the bug:
- as logged in user, when I come back to step one and change the donation amount, or when I cancel and start donation again, the ‘Next’ button on step 3 shouldn't be greyed out


src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -190,6 +190,7 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
    * Keys are ISO2 codes, values are names.
    */
   public countryOptionsObject: Record<string, string>;
+  paymentMethodChecked: boolean;
```
bdsl, line 193: I think we haven't been sticking this for all the new stepper work, but noticing here since it's between a private and a public: probably worth specifying which one a property is generally instead of letting it be public by default.

bdsl:
```suggestion
  private paymentMethodChecked: boolean;
```

bdsl: I'm also not clear if or why new should have an additional property. Shouldn't `!!this.selectedSavedMethod` always evaluate to the same thing as `this.paymentMethodChecked`?

src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -1481,6 +1488,9 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
     this.stripeResponseErrorCode = undefined;
 
     this.stripePaymentMethodReady = false;
+    if (this.stripeSavedMethods.length < 1) {
+      this.paymentMethodChecked = false;
```
bdsl, line 1492: wouldn't `this.paymentMethodChecked` have to already be false if there are no saved methods?

src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -894,6 +896,7 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
       this.stepper.next();
       return;
     }
+    this.stripePaymentMethodReady = this.paymentMethodChecked || this.stripeManualCardInputValid;
```
bdsl, line 899: Why should clicking next affect whether we consider the payment method to be ready?

bdsl (commented):
Added comments inline.

bdsl (approved):

PR #1132, opened 2023-05-26
DON-610: Add data driven unit test for DonationTippingSliderComponent…

bdsl:
….calcAndSetPercentage

Introducing data driven testing ( https://en.wikipedia.org/wiki/Data-driven_testing ) to the repo, would be good to see what people think.

If there were test failures they would show up like so:

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.spec.ts
+++ b/src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.spec.ts
@@ -32,13 +32,42 @@ describe('DonationTippingSliderComponent', () => {
     expect(component).toBeTruthy();
   });
 
-  it('should set the percentage defaults according to tipAmounts', () => {
-    const slider =  new DonationTippingSliderComponent({
-      listen: () => { },
-    } as unknown as Renderer2);
-    slider.donationAmount = 100;
-    slider.calcAndSetPercentage();
-    expect(slider.derivedPercentage).toBe(12.5);
-  });
+  usingDataIt('should have default percentage according to donation size', [
+  // [donation , expected default percentage]
+    [0, 15],
+    [99, 15],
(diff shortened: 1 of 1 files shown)


src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.spec.ts
```diff
@@ -32,13 +32,42 @@ describe('DonationTippingSliderComponent', () => {
...
+  usingDataIt('should have default percentage according to donation size', [
+  // [donation , expected default percentage]
+    [0, 15],
+    [99, 15],
+    [100, 12.5],
+    [299, 12.5],
+    [300, 10],
+    [999, 10],
+    [1000, 7.5],
+    ], function (name: string, args: Array<any>){
```
bdsl, line 44: would be nice to get rid of the `any`s, I feel like there should be a way to get rid of it with generics. But it's not as important for tests as it is in production.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.spec.ts
```diff
@@ -32,13 +32,42 @@ describe('DonationTippingSliderComponent', () => {
...
+    [300, 10],
+    [999, 10],
+    [1000, 7.5],
+    ], function (name: string, args: [number, number]){
+    it(name, () => {
+      const donationAmount = args[0];
+      const expectedDerivedPercentage = args[1];
+
+      const stubRenderer = {listen: () => { }} as unknown as Renderer2;
+      const slider =  new DonationTippingSliderComponent(stubRenderer);
```
bdsl, line 50:
```suggestion
      const dummyRenderer = {listen: () => { }} as unknown as Renderer2;
      const slider =  new DonationTippingSliderComponent(dummyRenderer);
```

PR #1135, opened 2023-05-30
COM-57 – use improved components

NoelLH:
Add Web "social" link; improve UX & accessibility


src/app/donation-complete/donation-complete.component.html
```diff
@@ -151,11 +151,31 @@ <h2>Thank you!</h2>
         <p>Boost your impact by sharing on social media</p>
         <div class="social-icons-wrapper">
           <span class="separator"></span>
-          <biggive-social-icon service="Facebook" url="https://www.facebook.com/sharer/sharer.php?u={{ shareUrl }}" background-colour="tertiary" icon-colour="black" wide="true"></biggive-social-icon>
+          <biggive-social-icon
+            service="Facebook"
+            label-prefix="Share"
+            url="https://www.facebook.com/sharer/sharer.php?u={{ shareUrl }}" background-colour="tertiary"
```
bdsl, line 157: I know not a new issue, but maybe worth encodeURI'ing the shareUrl we use here now.

src/app/donation-complete/donation-complete.component.ts
```diff
@@ -283,7 +283,7 @@ export class DonationCompleteComponent implements OnInit {
 
   private setSocialShares(campaign: Campaign) {
     const prefix = environment.donateGlobalUriPrefix;
-    this.shareUrl = `${prefix}/campaign/${campaign.id}`;
+    this.shareUrl = encodeURIComponent(`${prefix}/campaign/${campaign.id}`);
```
bdsl, line 286: ah you're quite right, it should be `encodeURIComponent` not `encodeURI`. Look at the docs on MDN I don't really understand now what the point of encodeURI is or when anyone would want that - encoding always has to be based on the output context, this is correct in the context we have that it's going in a URI. I don't know what context encodeURI is supposed to be used in.

bdsl: Not a blocker but might be worth renaming the property to `this.encodedShareUrl` to make sure we don't double encode it and make sure we don't forget to encode it if we assign it from anywhere else. Effectively it's a different type.

bdsl (approved):

src/app/donation-complete/donation-complete.component.ts
bdsl: Looking at https://stackoverflow.com/a/9247370/2526181 `encodeURI` seems to be a useless function.

PR #1136, opened 2023-05-30
DON-778 – improve donation maximum validation message

NoelLH:
Give donors with funds a more helpful error


bdsl (commented):
Looks good but we probably need to make the same change in the old version of this page too.

bdsl (approved):
LGTM

PR #1140, opened 2023-06-01
DON-791: Update tip slider with value set in custom tip field

bdsl:
If a custom tip outside of the min / max suggested percentage is selected, then we don't allow people to go back to the slider

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
@@ -76,7 +76,7 @@
             </mat-select>
           </mat-form-field>-->
 
-          <div *ngIf="showCustomTipInput" class="donation-input donation-input-main">
+          <div [style]="showCustomTipInput ? '' : 'display: none;'" class="donation-input donation-input-main">
 
             <biggive-text-input currency="{{campaign.currencyCode}}"  prompt="Donation to the Big Give Trust">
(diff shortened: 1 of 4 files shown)


src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -103,20 +103,20 @@ <h2 class="b-rt-0 b-m-0 b-bold">Tip The Big Give</h2>
...
             [onHandleMoved]="onDonationSliderMove"
           />
 
-          <span class="button" (click)="toggleTipInput()" *ngIf="!showCustomTipInput && donationAmount">
+          <span class="button" (click)="displayCustomTipInput()" *ngIf="!showCustomTipInput && donationAmount">
             Other amount
           </span>
           <!-- TODO: this check sets an infinite loop, need to check the logic -->
-          <span *ngIf="shouldShowBackToSlider()"  class="button" (click)="toggleTipInput()" >
+          <span *ngIf="showCustomTipInput && tipIsWithinSuggestedPercentRange"  class="button" (click)="displayPercentageTipInput()" >
```
bdsl, line 119: using an expression rather than a function so ZoneJS can detect when to re-evaluate the expression automatically - afaik it won't re-run an explicit function call.

dee-jo (approved):
LGTM

PR #1141, opened 2023-06-01
DON-790: Fix mobile version of new stepper

bdsl:
Various elements widened / made taller & made vertically stacked instead of side by side for narrow screens

image


dee-jo (approved):
LGTM!

PR #1143, opened 2023-06-02
DON-793: Fix movement of tip slider to match mouse

bdsl:
Appearance is unchanged, but now the slider moves correctly with the mouse pointer, and also recalculates its position when the window is resized.

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
@@ -208,9 +208,19 @@ export class DonationTippingSliderComponent implements OnInit, AfterContentInit,
     this.handle.nativeElement.style.marginLeft = this.position + 'px';
   }
 
+  private updateWidth() {
+    this.width = this.bar.nativeElement.getBoundingClientRect().width - this.handle.nativeElement.getBoundingClientRect().width;
+  }
+
   updateHandlePositionFromClick(pageX: number) {
-    const mousePos = pageX - this.bar.nativeElement.getBoundingClientRect().left - this.handle.nativeElement.getBoundingClientRect().width / 2;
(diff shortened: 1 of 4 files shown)


src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -208,9 +222,19 @@ export class DonationTippingSliderComponent implements OnInit, AfterContentInit,
     this.handle.nativeElement.style.marginLeft = this.position + 'px';
   }
 
+  private detectWidth() {
+    this.width = this.bar.nativeElement.getBoundingClientRect().width - this.handle.nativeElement.getBoundingClientRect().width;
+  }
+
   updateHandlePositionFromClick(pageX: number) {
-    const mousePos = pageX - this.bar.nativeElement.offsetLeft - this.handle.nativeElement.offsetWidth / 2;
```
bdsl, line 212: The source of the bug was that offsetLeft is the offset from the containing positioned element, not the viewport edge, bug pageX was measured from the viewport width. Replacing it with `this.bar.nativeElement.getBoundingClientRect().width` fixed the bug.

src/app/donation-start/donation-start-form/donation-tipping-slider/donation-tipping-slider.component.ts
```diff
@@ -56,8 +57,13 @@ export class DonationTippingSliderComponent implements OnInit, AfterContentInit,
   position: number;
   disableDefaults: boolean = false;
 
+  private readonly viewportChange = this.viewPortRuler.change(100)
```
bdsl, line 60: Don't know if 100ms is the perfect throttle delay. Shorter would give us smoother movement, longer would reduce resource usage on the client. The ideal might be to have a long throttle but then use a native browser animation to do the movement, but it's probably not worth learning how to do that right now.

PR #1144, opened 2023-06-02
DON-778: Let overidden donation limit completely replace standard limit

bdsl:


src/app/validators/currency-max.ts
```diff
@@ -12,7 +12,7 @@ export function getCurrencyMaxValidator(limitOverride?: number): ValidatorFn {
 
       let effectiveLimit = environment.maximumDonationAmount;
```
bdsl, line 13: Possibly we should have unit tests for the validators - there aren't any at present, I could start writing some.

dee-jo (approved):
LGTM

PR #1146, opened 2023-06-05
DON-803: Link from new donation stepper & buy credits to Anchor Match…

bdsl:
… Fund Page

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
@@ -139,7 +139,9 @@
             <li>Supporting our website and team, allowing charities to double their
               donations to good causes.
             </li>
-            <li>Providing match funds in our Anchor Match Fund. Through tips, charities will receive funding to launch their own match funded campaign on Big Give.
+            <li>Providing match funds in our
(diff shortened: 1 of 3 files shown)


src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -139,7 +139,9 @@ <h3 class="b-rt-0 b-m-0 b-bold">How does Big Give use tips?</h3>
             <li>Supporting our website and team, allowing charities to double their
               donations to good causes.
             </li>
-            <li>Providing match funds in our Anchor Match Fund. Through tips, charities will receive funding to launch their own match funded campaign on Big Give.
+            <li>Providing match funds in our
+              <a href="https://biggive.org/anchor-match-fund/" target="_blank"><mat-icon class="b-va-bottom" aria-hidden="false" aria-label="Open in new tab">open_in_new</mat-icon>Anchor Match Fund</a>
```
bdsl, line 143: Could have made this link to the matching environment, e.g. dev / staging / production, but for something like this I think we might as well just keep it simple and always link to production.

dee-jo (dismissed):

PR #1152, opened 2023-06-06
Revert "DON-778: Let overidden donation limit completely replace standard limit", fix error message

bdsl:
Reverts thebiggive/donate-frontend#1144


dee-jo (approved):
LGTM!

PR #1153, opened 2023-06-08
DON-772: Increase donation limit based on funds balance

bdsl:
This should not be deployed (and therefore maybe should not be merged to develop?) before changes from https://github.com/thebiggive/matchbot/pull/559 are in production.

image

image


dee-jo (approved):
LGTM

PR #1155, opened 2023-06-09
DON-830: Upgrade to RxJS 7

bdsl:
Seems to work, builds, tests passing and site loads on local. And will stop us doing things like:

image

(not that we have been doing that)

Needs more testing before merging for real, so this is just a spike I'm planning to close.


bdsl:
Let's just hold this back for Monday.

PR #1164, opened 2023-06-16
Don 815 no default custom tip

bdsl:
As clearing the custom tip input would remove the ability to go back to the slider if we kept hte existing behaviour that the slider is only available when the custome input is within the 1%-30% range, I changed the logic to make make the link to return to the slider always available, and when the user goes to the slider we adjust their tip to be the closet to what they have selected that lies within that 1-30 range.


NoelLH (approved):
Looks good. Only suggestion is for a comment typo

PR #1165, opened 2023-06-20
DON-824: card for DEC Ukraine appleal

dee-jo:


bdsl (approved):
LGTM, just worth adding a screenshot either here or on Jira and mentioning that you've done see results too. Let's keep this unmerged until Thursday or Friday so we can deploy then.

PR #1171, opened 2023-06-23
DON-805: add padding between closed steps

dee-jo:


bdsl:
@dee-jo For visual stuff like this it would be really helpful generally if you could put before/after screenshots on the PRs. 

Also as I'm reviewing this alongside https://github.com/thebiggive/donate-frontend/pull/1172 I'm wondering how the page will look when both PRs are merged together. I'm not sure how much sense to makes to review them separately. But if you want them reviewed separately will you add screenshots to that one too? Thanks.

bdsl (approved):
LGTM

PR #1173, opened 2023-06-26
DON-788: Allow switching between tip slider and drop-down

bdsl:
To test this add `?tipControl=dropdown` on to the address bar when looking at the new stepper - this will replace the slider control with a drop down control. We still need to work out how to run an a/b test of the two, but probably worth getting the implementation with the dropdown selected this way QA'd first.

The edges of the input are a bit finicitky and can disappear depending on zoom in Firefox - but I think that's a generall issue with all these dropdowns, not specific to tipping.

Screenshots as on Jira:

Screenshot from 2023-06-27 11-52-37
Screenshot from 2023-06-27 11-52-56


dee-jo (approved):
lgtm!

PR #1181, opened 2023-06-29
Componets merge wip

dee-jo:


bdsl:
This was made by me and @dee-jo pairing. It's just an experimental spike to see how possible this would be. We didn't get as far as actually having the frontend load with the merged components. Doing it for real would be quite a bit of work - I do think it might be a good idea but almost certainly not something we want to do for real in the immediate future.

NoelLH:
Given Barney's comment I will close this for now. I think we should also discuss the relative benefits vs. the complexity, repo pull size needed etc. for devs and agencies working just on WordPress with the components.

PR #1184, opened 2023-07-04
DON-802: update styling of Material accordion and stepper "Continue" buttons

dorota-joanna:
Added horizontal line to the tips info section of the stepper and decreasd font sizes on step labels and buttons:

Screenshot from 2023-07-12 13-22-30

Screenshot from 2023-07-12 13-25-30


src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -139,9 +139,9 @@ <h2 class="b-rt-0 b-m-0 b-bold">Tip Big Give</h2>
           <mat-expansion-panel (opened)="panelOpenState = true"
                                (closed)="panelOpenState = false">
             <mat-expansion-panel-header>
-              <h3 class="b-rt-0 b-m-0 b-bold">How does Big Give use tips?</h3>
+              <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give use tips?</h3>
```
NoelLH, line 142: Couldn't we apply a border below the `h3` itself? This feels like it would be easier to follow and less brittle than adding an absolutely positioned extra element.

dorota-joanna: Hey @NoelLH , yes I remember trying that but couldn't make the border extend to the edges this way
Screenshot from 2023-07-18 12-16-41

NoelLH: @dorota-joanna OK. I think let's not spend more time on this part as we can hopefully fix the components version of the accordion at some point and substitute it in.

NoelLH (approved):

NoelLH:
@dorota-joanna I've tried to make the PR title reflect what I understand us to actually be doing now. Maybe see if you think this is accurate now before we merge, so the history gives a better sense of what we did?

PR #1190, opened 2023-07-17
Refactor: Required-input all the things

bdsl:
Taking advantage of the new feature in Angular 16 to express when an input is always set. This gets checked at compile time in all templates that use the component

See https://netbasal.com/from-good-to-great-required-inputs-in-angular-47374c8a43cf

I dif a fairly niave search for all the @Input decotators and added required - then took it away anywhere it triggered a compile error. Seems reasonable to assume that if we currently allways pass an input to a componen then it's probably required.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
@@ -84,7 +84,7 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
   requestButtonShown = false;
   showChampionOptIn = false;
 
-  @Input() campaign: Campaign;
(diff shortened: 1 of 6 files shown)


src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -84,8 +84,7 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
   requestButtonShown = false;
   showChampionOptIn = false;
 
-  @Input() campaign: Campaign;
-  @Input() column: 'primary'|'secondary'
```
bdsl, line 88: looks like this column field was never used. Not sure what it was for.

dorota-joanna (approved):
LGTM

PR #1191, opened 2023-07-17
DON-810: Inherit site font into input fields in new stepper

bdsl:
See e.g. https://stackoverflow.com/a/2875030

I'm also wondering if there's a better place to put this code to mkae it not specific to the donation page.

And on the other hand whether there's a good enough reason for browsers to bring their own default styling to input fields that means we should leave it alone. But I know using css-reset files is a widespread practice. I couldn't really find any description of exactly why input elements have font family set by browsers.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form-new.component.scss
+++ b/src/app/donation-start/donation-start-form/donation-start-form-new.component.scss
@@ -168,6 +168,10 @@ input:disabled {
   }
 }
 
+input {
+  font-family: inherit;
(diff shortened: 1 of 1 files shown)


NoelLH:
@bdsl There is a global override file `_overwrite-mat.scss`. Mostly it's been used sparingly and to focus on Angular Material global class overrides, but as we have a `.donation-input { ... }` block there already it might be worth a try to see if it would also work for overriding this app-wide?

PR #1193, opened 2023-07-18
DON-810: Adjust new stepper font sizes: Header and payment card inputs

bdsl:
Changes as discussed in last comment on Jira: https://thebiggive.atlassian.net/browse/DON-810?focusedCommentId=31188

Before:

image

After:

image

Before:

image

After:

image

Also note the font family inside the payment card input isn't as it should be above - but that seems to just be an issue with the local dev environment. It should be fine when we deploy this code to staging / prod.

Note the font size of the country drop down "United Kingdom" is still 17px rather than 18 - this will need changing in the component library. Easiest change is to just make it 18px there - that will also affect the filter drop downs in the explore page, which may be OK, or we could make it inherit a font size or allow passing one in. I guess it's not ideal that we're picking sizes looking at one page at a time rather than properly following a system across the whole site.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form-new.component.scss
+++ b/src/app/donation-start/donation-start-form/donation-start-form-new.component.scss
@@ -25,7 +25,7 @@ div.mdc-form-field label {
   // which is different for the selected and non-selected step. Also overriding that by being more specific
   // in the selector above, including 'div', but don't want to rely just on that.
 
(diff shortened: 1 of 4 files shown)


bdsl:
(screenshots are slightly out of date, 18px font sizes reduced by 5.5% since then. Not sure I can be bothered to update them)

dorota-joanna (approved):
LGTM

PR #1195, opened 2023-07-20
DON-836 – minor Angular 16 tuning

NoelLH:
* Reduce a bit of boilerplate with new route param bind to `@Input()` – doesn't seem to be a thing in route resolvers unfortunately
* Minor update to get bug fix that might impact our load performance (not tested with SSR locally)

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -48,6 +48,7 @@ const matomoTrackers = environment.matomoSiteId ? [
       }
     }),
     RouterModule.forRoot(routes, {
+      bindToComponentInputs: true,
       initialNavigation: 'enabledBlocking', // "This value is required for server-side rendering to work." https://angular.io/api/router/InitialNavigation
       onSameUrlNavigation: 'reload', // Allows Explore & home logo links to clear search filters in ExploreComponent
       scrollPositionRestoration: 'enabled',
--- a/src/app/donation-complete/donation-complete.component.ts
+++ b/src/app/donation-complete/donation-complete.component.ts
@@ -1,5 +1,5 @@
 import { HttpErrorResponse } from '@angular/common/http';
-import { Component, OnInit, ViewChild } from '@angular/core';
+import { Component, Input, OnInit, ViewChild } from '@angular/core';
 import { DomSanitizer, SafeHtml } from "@angular/platform-browser";
 import { MatDialog } from '@angular/material/dialog';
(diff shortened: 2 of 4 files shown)


src/app/app.module.ts
```diff
@@ -48,6 +48,7 @@ const matomoTrackers = environment.matomoSiteId ? [
       }
     }),
     RouterModule.forRoot(routes, {
+      bindToComponentInputs: true,
```
bdsl, line 51: +1
I read https://itnext.io/bind-route-info-to-component-inputs-new-router-feature-1d747e559dc4 , looks like a good feature.

src/app/donation-complete/donation-complete.component.ts
```diff
@@ -65,10 +65,7 @@ export class DonationCompleteComponent implements OnInit {
     private route: ActivatedRoute,
     private sanitizer: DomSanitizer,
   ) {
-    route.params.pipe().subscribe(params => {
-      this.donationId = params.donationId;
-      this.checkDonation();
-    });
+    this.checkDonation();
```
bdsl, line 68: nice simplification - and did the old version have a memory leak because it didn't unsubscribe to the params on destroy?

src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -45,11 +46,16 @@ const endPipeToken = 'timeLeftToEndPipe';
   ],
 })
 export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnInit {
+  // Campaign ID may be passed instead
+  @Input({ required: false }) private campaignSlug: string;
+  // Passed only on the fund-filtered view of this page.
+  @Input({ required: false }) private fundSlug: string;
```
bdsl, line 52: `required: false` doesn't do anything, but I like it - maybe it's a good idea to make it our coding standard that we always pass required true or false to input decorators.

bdsl (approved):

src/app/donation-complete/donation-complete.component.ts
NoelLH: Maybe!

src/app/meta-campaign/meta-campaign.component.ts
NoelLH: Yep – I think for new or newly-modified inputs at least, it's helpful to show we have thought about it and explain why things are optional.

PR #1198, opened 2023-07-20
DON-744: Add Anchor Match Fund highlight card on Homepage

dorota-joanna:
Updated Homepage Cards:
Screenshot from 2023-07-20 12-18-27


bdsl (approved):
LGTM

PR #1201, opened 2023-07-21
DON-836 – minor Angular 16 tuning (mk II)

NoelLH:
* Reduce a bit of boilerplate with new route param bind to `@Input()` – doesn't seem to be a thing in route resolvers unfortunately
* Minor update to get bug fix that might impact our load performance (not tested with SSR locally)

Basically same as https://github.com/thebiggive/donate-frontend/pull/1195 except using `ngOnInit()` appropriately in `DonationCompleteComponent`. This also lets us stop injecting `ActivatedRoute` in that case.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,12 +108,12 @@
       }
     },
     "node_modules/@angular-devkit/architect": {
-      "version": "0.1601.4",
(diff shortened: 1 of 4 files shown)


bdsl (approved):
LGTM

PR #1202, opened 2023-07-21
DON-706: Store donor dator on transfer funds tip

bdsl:
This relies on matchbot changes at https://github.com/thebiggive/matchbot/pull/577 to be useful, but the PRs can safely be deployed in either order - matchbot will just ignore the additional fields at the moment.

There are no changes here visible to the donor, but after donating (given the matchbot changes is also present) the entry in the matchbot DB looks like this:

image

### DIFF
--- a/src/app/transfer-funds/transfer-funds.component.ts
+++ b/src/app/transfer-funds/transfer-funds.component.ts
@@ -406,26 +406,15 @@ export class TransferFundsComponent implements AfterContentInit, OnInit {
       donation.pspCustomerId = this.identityService.getPspId();
     }
 
-    this.createDonation(donation);
-  }
-
-  /**
-   * Creates a Donation itself. Both success and error callbacks should unconditionally set `creatingDonation` false.
-   */
-  private createDonation(donation: Donation) {
-    if (this.donor === undefined) {
-      throw new Error("Cannot create donation without logged in donor");
-    }
-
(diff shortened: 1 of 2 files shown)


src/app/transfer-funds/transfer-funds.component.ts
```diff
@@ -401,26 +402,19 @@ export class TransferFundsComponent implements AfterContentInit, OnInit {
...
       donation.pspCustomerId = this.identityService.getPspId();
     }
 
-    this.createDonation(donation);
-  }
-
-  /**
-   * Creates a Donation itself. Both success and error callbacks should unconditionally set `creatingDonation` false.
-   */
-  private createDonation(donation: Donation) {
```
bdsl, line 414: inlined this function - as a function called just once and containing just a single function call I don't think there was much value in having it separate - and having it separate meant repeating the check for `this.donor` being defined.

src/app/transfer-funds/transfer-funds.component.ts
```diff
@@ -367,7 +361,14 @@ export class TransferFundsComponent implements AfterContentInit, OnInit {
       return;
     }
 
+    if (this.donor === undefined) {
+      throw new Error("Cannot create donation without logged in donor");
+    }
```
bdsl, line 366: move guard clause to top of function

dorota-joanna (approved):
LGTM

PR #1203, opened 2023-07-21
DON-845: added 'Match Funded' badge to highlight card on donate page

dorota-joanna:
Screenshot from 2023-07-21 14-03-24

### DIFF
--- a/src/app/donation-start/donation-start-secondary/donation-start-secondary.component.html
+++ b/src/app/donation-start/donation-start-secondary/donation-start-secondary.component.html
@@ -6,12 +6,12 @@
             sizes of icons
             -->
     <div>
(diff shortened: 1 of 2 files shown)


bdsl (dismissed):
LGTM. Did you find out how the badge disappeared?

dorota-joanna:
> LGTM. Did you find out how the badge disappeared?

Not really, there was no definition for the match-funded-badge class in the scss file, so just defined it from scratch.

bdsl:
Might be another good case for git-bisect - if we find out how it disappeared we can think about if there's anything we can do different to stop stuff disappearing like that when we don't want it to.

bdsl (approved):

PR #1207, opened 2023-07-25
DON-816 – fix odd layout on new inputs

NoelLH:
Remove `float: right` which makes right padding
significant, and variable across browsers. And reset `padding` to 0 to further improve consistency
across browsers.


bdsl (approved):
LGTM - think the float right was there to make the money numbers go right, not needed now as width: 100% along with text-align right does the same thing better

PR #1209, opened 2023-07-25
DON-819: Allow spaces at start and end of donation / tip amount input

bdsl:
Removes one source of disabled button

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form-new.component.spec.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form-new.component.spec.ts
@@ -475,7 +475,7 @@ describe('DonationStartNewPrimaryComponent', () => {
     const donationAmountErrors: any = component.donationForm.controls.amounts!.get('donationAmount')?.errors;
     expect(Object.keys(donationAmountErrors).length).toBe(1);
     expect(donationAmountErrors.pattern).toEqual({
-      requiredPattern: '^ *[£$]?[0-9]+?(\\.00)? *$',
+      requiredPattern: '^\\s*[£$]?[0-9]+?(\\.00)?\\s*$',
       actualValue: '8765,21',
     });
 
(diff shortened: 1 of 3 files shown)


src/app/donation-start/donation-start-form/donation-start-form-parent.component.ts
```diff
@@ -1652,7 +1652,10 @@ export class DonationStartFormParentComponent implements AfterContentChecked, Af
       if (this.creditPenceToUse === 0) {
         this.amountsGroup.controls.tipAmount!.setValidators([
           Validators.required,
-          Validators.pattern('^[£$]?[0-9]+?(\\.[0-9]{2})?$'),
+          // We allow spaces at start and end of amount inputs because people can easily paste them in
+          // by mistake, and they don't do any harm. Maxlength in the HTML makes sure there can't be so much as
```
NoelLH, line 1656: Not sure we can be totally confident they're harmless without trimming?

There are tests that need updating relating to the validation here.

bdsl: Right. I've updated the tests and added a call to `.trim()` before sanitiseCurrency.

src/app/donation-start/donation-start-form/donation-start-form-new.component.spec.ts
```diff
@@ -475,7 +475,7 @@ describe('DonationStartNewPrimaryComponent', () => {
     const donationAmountErrors: any = component.donationForm.controls.amounts!.get('donationAmount')?.errors;
     expect(Object.keys(donationAmountErrors).length).toBe(1);
     expect(donationAmountErrors.pattern).toEqual({
-      requiredPattern: '^[£$]?[0-9]+?(\\.00)?$',
+      requiredPattern: '^\s*[£$]?[0-9]+?(\\.00)?\s*$',
```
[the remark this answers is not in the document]
bdsl: @NoelLH You're quite right. Fixed.

NoelLH (approved):

PR #1210, opened 2023-07-25
Use `autocapitalize` where appropriate

NoelLH:
Save mobile donors a little effort on typing by hinting where simplified keyboards can be used.

(& improve the chance of us saving better quality postcodes prior to formatting)

Just focusing on the new donation form as we expect it to be live soon.

Identity email is compared as `_ci` so case variation shouldn't block existing donors from logging in.


bdsl (approved):
LGTM, don't remember if I knew about autocapitalize before.

PR #1211, opened 2023-07-25
DON-834: Fix autocmplete not working on email in new stepper

bdsl:
Not sure why it wasn't working here specifically, but telling the browser exactly what we want autocompleted seems to fix it. I tested with 1password on my local

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
@@ -266,6 +266,7 @@
         <div *ngIf="!giftAidGroup.value.homeOutsideUK">
           <biggive-text-input prompt="Enter your home postcode">
             <input
+              autocomplete="postal-code"
               formControlName="homePostcode"
               id="homePostcode"
               autocapitalize="characters"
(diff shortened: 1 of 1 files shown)


src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -320,7 +321,7 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
 
           <div>
             <biggive-text-input  prompt="Email Address *">
-              <input slot="input" formControlName="emailAddress" id="emailAddress" type="email" matInput>
+              <input slot="input" formControlName="emailAddress" id="emailAddress" type="email" autocomplete="email" matInput>
```
bdsl, line 324: Having a `label` element associated with the input might be a better or good additional way to make the autocomplete work - but I can't see an easy way to do that given that the input element is here and if we had a label it would need to be positioned as part of the components in the components library.

NoelLH: There might be a more important point here about accessibility @bdsl – have we tried the new approach with a screen reader?

If `prompt` is not used in an ARIA attribute or an actual `label`, I'm not sure we are really associating it semantically in a way that will keep the form usable for everyone.

dorota-joanna (dismissed):
LGTM

src/app/donation-start/donation-start-form/donation-start-form-new.component.html
bdsl: @NoelLH You're right, we should start using labels. I've made a ticket for that https://thebiggive.atlassian.net/browse/DON-850

bdsl:
This had a merge conflict with changes adding `autocapitalize` on the same lines. Rebased on latest develop.

PR #1212, opened 2023-07-26
New stepper: fix postcss plugin build warning; ensure consistent layouts

NoelLH:
> "Warning: Module Warning (from ./node_modules/postcss-loader/dist/cjs.js) (1550:3) from "autoprefixer" plugin: start value has mixed support, consider using flex-start instead"


bdsl:
Thanks @NoelLH

PR #1213, opened 2023-07-26
DON-838: New stepper / Reduce match funding banner to fit on mobile

bdsl:
Before:

image

After:

image

I would ideally like to get rid of the tiny bit of black visible on the right edge as well, but that seems fiddly and not sure how much time it's worth right now.


dorota-joanna (approved):
LGTM

PR #1214, opened 2023-07-26
DON-802: New stepper / remove box shadow around "How Does Big Give Us…

bdsl:
…e Tips" & keep horizontal line visible

Before:

image

image

After:

image

image


dorota-joanna (approved):
LGTM

PR #1216, opened 2023-07-27
Don 849 delete old stepper elements

bdsl:
image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form-new.component.html
@@ -601,15 +601,8 @@
               <td><div><strong>Total amount you will be charged</strong></div></td>
               <td><strong>{{ donationAndExtrasAmount | exactCurrency:donation.currencyCode }}</strong></td>
             </tr>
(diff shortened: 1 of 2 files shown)


dorota-joanna (dismissed):

src/app/donation-start/donation-start-form/donation-start-form-new.component.html
```diff
@@ -29,7 +29,19 @@
...
+               -->
+              <label
+                *ngIf="campaign.charity.name.length < 20; else donationLabelTruncated"
+                slot="label"
+                for="donationAmount"
+              >
+                Donation to {{campaign.charity.name}}
+              </label>
+              <ng-template #donationLabelTruncated>
+              <label slot="label" for="donationAmount">Donation to {{campaign.charity.name.slice(0, 17)}}…</label>
```
bdsl, line 43: this is not perfect because slice splits by code units, would be better to split by graphemes. Maybe using Intl.Segmenter is the proper way, but the chance we have a very long charity name with more than one code unit per grapheme (e.g. a letter combined with a diacritic, or an emoji with a skin tone applied) is pretty low.

NoelLH (approved):

PR #1221, opened 2023-07-27
DON-850: Use new components version, taking labels for text inputs in…

bdsl:
… new stepper page

New components version also includes changes to components used on other parts of the site which need testing.

Pulls in changes from components released here: https://github.com/thebiggive/components/pull/248

We've already found at least one bug caused by pulling in the component updates.


tomwpaterson-shape:
Hi @bdsl, The Slot is deleted so that we have some control over the CSS for child content. If child elements are added in the slot, the CSS cannot be inherited from the parent CSS. Can you think of an alternative?

bdsl:
Updated components version again: 202307271726.0.0 has the slot removal reverted so we can see the all participating campaigns again:

image

PR #1227, opened 2023-08-01
DON-579: new endpoint for campaign stats

dorota-joanna:
Just some setup - not including the dynamic variable in the template yet until the SF part is ready


src/app/campaign-stats.model.ts
```diff
@@ -0,0 +1,6 @@
+export class CampaignStats {
+    constructor(
+        public totalRaisedWithAdjustment: number,
+        public totalCampaignCount: number
```
bdsl, line 4: It would be good to have docblocks for these two properties to explain what they mean and what units they're measured in. I don't know if it's worth making the first one into an object that contains the currency code as well as the number, if not then the fact that it's GBP counted in whole pounds should be in the docblock and/or the property name.

src/app/campaign.service.spec.ts
```diff
@@ -98,6 +105,23 @@ describe('CampaignService', () => {
...
+    service.getCampaignImpactStats().subscribe(campaignStats => {
+      expect(campaignStats).toEqual(dummyCampaignStats);
+    }, () => {
+      expect(false).toBe(true);
+    });
+
+    const request = httpMock.expectOne(`${environment.apiUriPrefix}/campaigns/services/apexrest/v1.0/campaigns/stats`);
+    expect(request.request.method).toBe('GET');
+    request.flush(dummyCampaignStats);
+  });
```
bdsl, line 123: the Subject Under Test of this code is just the one liner function `getCampaignImpactStats`, right? That function seems so simple I'm not convinced it's worth testing, especially with a test this long. What sort of mistakes would we expect this test to catch / prevent that the compiler won't catch automatically?

Also the last three lines don't seem to be anything to do with the SUT - what are they for?

src/app/campaign.service.ts
```diff
@@ -190,6 +191,10 @@ export class CampaignService {
     return of([
     ]);
   }
+
+  getCampaignImpactStats(): Observable<CampaignStats>{
```
bdsl, line 195: if you want to be able to show this in a template right now without waiting for the sf part to be done you could temporarily add a line like:
```ts
    return of(new CampaignStats(42, 65)); // 
```

then you just have to delete that line when we're ready to actually make the get request to SF.

src/app/campaign.service.spec.ts
bdsl: On the question of whether to test every function - this is a good answer from Kent Beck, author of Test Driven Development:

https://stackoverflow.com/questions/153234/how-deep-are-your-unit-tests/153565#153565

src/app/campaign.service.ts
NoelLH: Yep – there are also examples in the repo which are closely analogous like `CampaignResolver`. (Although for a value used once in the app we would probably want to simplify by removing `TransferState` stuff, which I think is a per-session cache. https://levelup.gitconnected.com/a-beginners-guide-to-implementing-transferstate-in-angular-a313f3244f03)

src/app/campaign-stats.model.ts
```diff
@@ -0,0 +1,12 @@
+export class CampaignStats {
+    constructor(
+        /**
+         * @param: total amount raised (including adjustment) for all except master campaigns
```
bdsl, line 4: I only did half a physics degree, the main thing I kept from it was that a number without a unit is meaningless. Is this right?

```suggestion
         * @param: total amount raised (including adjustment) for all except master campaigns in whole pounds GBP
```

bdsl: Also not sure what 'including adjustment' means here, no if we need to say excluding master campaigns. We only need to document the meaning of this field, not necessarily the method used to calculate it.

src/app/campaign.service.ts
```diff
@@ -190,6 +191,11 @@ export class CampaignService {
...
+    return of(new CampaignStats(1000, 5));
```
bdsl, line 196: Doesn't this need to be as below? These are the numbers I see now at https://donate.biggive.org/
```suggestion
    return of(new CampaignStats(242_309_804, 15_000));
```

src/app/campaign-stats.model.ts
NoelLH: Whole pounds, not sure – I don't think we really need to round it in Apex.

GBP, yes – as discussed on the SF counterpart, it should do an approx conversion to our "home" currency. I think for this endpoint, given we have no non-GBP real fundraising happening yet, it's good enough.

I also have a slight preference for just calling it `totalRaised` in the API.

src/app/home/home.component.ts
```diff
@@ -36,5 +37,9 @@ export class HomeComponent implements OnInit {
       'Big Give – discover campaigns and donate',
       'https://images-production.thebiggive.org.uk/0011r00002IMRknAAH/CCampaign%20Banner/db3faeb1-d20d-4747-bb80-1ae9286336a3.jpg',
     );
+
+    this.campaignService.getCampaignImpactStats().subscribe(campaignStats => {
+      this.dynamicImpactStats = campaignStats;
```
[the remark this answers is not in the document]
NoelLH: Yeah, all our route resolvers would maybe do better to pipe. (The equivalent of this will probs move to one)

bdsl (approved):

PR #1228, opened 2023-08-02
DON-861: Update homepage cards

bdsl:
New homepage appears as below:

image

I took converted the PNG from Jira to a Jpeg for a much lower file size.

There's also a refactoring commit to simplify the code for the homepage by removing the code that was put in to get ready for pulling campaign data from SF to the homepage. We can put that back in when we're going to use it, but while we're hard coding the home page contents it's easier without that.


dorota-joanna (approved):
LGTM

PR #1236, opened 2023-08-08
Don 783 cookie banner

bdsl:
Looks like this so far. A bit more work to do.

image

image

### DIFF
--- a/src/app/cookie-consent-banner/cookie-consent-banner.component.scss
+++ b/src/app/cookie-consent-banner/cookie-consent-banner.component.scss
@@ -1,0 +1,4 @@
+div.cooke-consent-container {
+  background: white;
+  border: 1px solid black;
+}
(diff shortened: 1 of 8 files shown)


src/app/cookie-consent-banner/cookie-consent-banner.component.scss
```diff
@@ -0,0 +1,30 @@
...
+  width: 100%;
+  flex-direction: column;
+
+  @media #{$breakpoint-md} {
+    flex-direction: row;
+  }
+
+    biggive-button {
+      gap: 1em;
+      margin-bottom: 10em;
```
bdsl, line 28: not sure yet why neither of these are working to create a vertical gap between buttons on narrow screens.

bdsl:
Closing this as the banner is going to be built in components so it can be shared with wordpress.

PR #1238, opened 2023-08-09
DON-783: Use alternate recaptcha URL to avoid accessing the many cook…

bdsl:
…ies from google.com

As suggested at https://developers.google.com/recaptcha/docs/faq#does-recaptcha-use-cookies and as shown in the ng-recaptcha readme: https://github.com/DethAriel/ng-recaptcha/blob/master/README.md

Before:

image
image

After:

image
image

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -60,11 +60,8 @@ export function app() {
           'fonts.googleapis.com',
           'fonts.gstatic.com',
           'js.stripe.com',
-          'optimize.google.com',
           'player.vimeo.com',
-          // Next 2 needed for reCAPTCHA to fully load.
-          'recaptcha.google.com',
-          'www.google.com',
+          'recaptcha.net',
           'www.youtube.com',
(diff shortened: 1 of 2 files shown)


dorota-joanna (approved):
LGTM

NoelLH:
> > `server.ts` will require corresponding CSP changes for the service to continue working w/ Angular Universal.
> > (We can also remove `optimize.google.com` while we're there.)
> 
> Thanks - do you think [25cea7a](https://github.com/thebiggive/donate-frontend/commit/25cea7a8b254e1c3e20becaef9686e1bdfb7695a) should do it?

@bdsl Probably – but in future I would recommend running the SSR local/Dockerised setup described in the readme and checking a browser's network logs to be sure, before merging ideally, to make sure no secondary resources we would expect to get through are blocked.

As it is we can do the same on Staging.

PR #1240, opened 2023-08-15
DON-783: Save donations & campaigns in session storage not local storage, save ID & JWT in cookies

bdsl:
This means the thank you page will now only be viewable in the same tab where the donation was made, and not at all after the tab is closed. While it might be nice to have the thank you page available for longer I think this is OK and avoids storing more info in the browser than we need. Donors should be able to see all the details of their old donations in their emails.

Session storage is scoped to the browser tab and automatically expires when the tab is closed, whereas local storage never expires.

Ideally I would like something that persists for the session only (or for a short time like 1 day) but is available across all tabs. That seems more work to set up, would perhaps require using a cookie and relying on the server side db for the content instead.

### DIFF
--- a/src/app/donation.service.ts
+++ b/src/app/donation.service.ts
@@ -237,7 +237,7 @@ export class DonationService {
   }
 
   /**
-   * This funtion can be deleted 1 day after this version is deployed - when we store only in session storage there is no need
+   * @todo This funtion can be deleted 1 day after this version is deployed - when we store only in session storage there is no need
    * to remove anything.
    */
   removeOldLocalDonations() {
--- a/src/app/donation.service.ts
+++ b/src/app/donation.service.ts
@@ -236,6 +236,10 @@ export class DonationService {
     this.sessionStorage.set(this.storageKey, donationCouplets);
   }
 
+  /**
+   * This funtion can be deleted 1 day after this version is deployed - when we store only in session storage there is no need
+   * to remove anything.
+   */
   removeOldLocalDonations() {
     const donationsOlderThan8Hours: Array<{ donation: Donation, jwt: string }> = this.getDonationCouplets().filter(donationItem => {
       return (!donationItem.donation.createdTime || this.getCreatedTime(donationItem.donation) < (Date.now() - (8 * 60 * 60 * 1_000)));
--- a/src/app/donation.service.ts
+++ b/src/app/donation.service.ts
@@ -232,14 +232,15 @@ export class DonationService {
       1,
     );
 
+    this.storage.set(this.storageKey, donationCouplets);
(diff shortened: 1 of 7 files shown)


src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -97,7 +96,7 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
     private route: ActivatedRoute,
     public searchService: SearchService,
     private state: TransferState,
-    @Inject(TBG_DONATE_STORAGE) private storage: StorageService,
+    @Inject(SESSION_STORAGE) private storage: StorageService,
```
NoelLH, line 99: I guess the thinking here is just that maintaining previous cards isn't useful after a while, rather than a PII/privacy consent issue?

I can't immediately see a problem with that, but it might be worth renaming the var for consistency with the other services, to avoid confusion should we want to expand what we store here.

NoelLH (commented):
Looks good! Minor things inline, no blockers

src/app/donation.service.ts
```diff
@@ -230,14 +231,20 @@ export class DonationService {
       donationCouplets.findIndex(donationItem => donationItem.donation.donationId === donation.donationId),
       1,
     );
+
     this.storage.set(this.storageKey, donationCouplets);
+    this.sessionStorage.set(this.storageKey, donationCouplets);
   }
 
+  /**
+   * This funtion can be deleted 1 day after this version is deployed - when we store only in session storage there is no need
```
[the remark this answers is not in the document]
bdsl:
```suggestion
   * @todo This funtion can be deleted 1 day after this version is deployed - when we store only in session storage there is no need
```

src/app/meta-campaign/meta-campaign.component.ts
bdsl: Renamed in 263848d9

src/app/donation.service.ts
```diff
@@ -230,14 +231,20 @@ export class DonationService {
...
     );
+
     this.storage.set(this.storageKey, donationCouplets);
+    this.sessionStorage.set(this.storageKey, donationCouplets);
   }
 
+  /**
+   * @todo This funtion can be deleted 1 day after this version is deployed - when we store only in session storage there is no need
+   * to remove anything.
+   */
```
bdsl, line 242:
```suggestion
```

That isn't really true, we should keep deleting stuff indefinitely, since we can't rely on the browser to ever automatically remove what we put in local storage in the past.

NoelLH (commented):
As discussed, cookie addition for more persistence looks good, and the SSR-ready version of the new lib would probably be even better

NoelLH (approved):

PR #1242, opened 2023-08-15
CI installation change WIP

NoelLH:
I started trying to do a bigger refactor of the npm bits, but not that much is actually shared so I'm not sure it's worth it. Pushing and resetting (mentally!)


NoelLH:
Think I'm over this one already! It feels like there should be a way to make the config more concise, but this is probably not it.

PR #1244, opened 2023-08-15
Turn inlining critical styles on

NoelLH:
Making a draft PR, as I did this locally last week and hadn't committed yet. (And conflicts in these areas are unlikely.)

I don't suggest we merge before some local SSR build testing has happened – a sanity check of the bundle sizes at least and preferably a real app load.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -59,7 +59,7 @@
                 "scripts": true,
                 "styles": {
                   "minify": true,
-                  "inlineCritical": false
+                  "inlineCritical": true
                 },
                 "fonts": true
               },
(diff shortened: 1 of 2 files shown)


NoelLH:
Note too sure why the CI workflow now called `commit` hasn't run on this, given I merged the latest `develop` into this branch.

NoelLH:
I did a quick test of the Angular 17 equivalent today (same thing but the server value is in `renderOptions`).

From very anecdotal evidence it doesn't seem we get a big difference between when it's off:
don-950 perf with inline critical off

and on:
don-950 perf with inline critical on

So in the interest of not changing more than necessary at once I think we should stick with it being off for right now. I will close this PR as it won't quite align with the change needed on Angular 17+ anyway.

We might want to review if and when we identify the cause of incomplete components in SSR mode as that is probably a much bigger source of layout shift which might be preventing benefits from inlining styles.

PR #1245, opened 2023-08-16
DON-783: Properly clear out donations from local storage, removing st…

bdsl:
…orage key instead of saving useless empty array


src/app/donation.service.spec.ts
```diff
@@ -54,9 +54,11 @@ describe('DonationService', () => {
       RouterTestingModule,
     ],
     providers: [
-      // Inject in-memory storage for tests, in place of local storage.
-      { provide: TBG_DONATE_STORAGE, useExisting: InMemoryStorageService },
-      { provide: SESSION_STORAGE, useExisting: InMemoryStorageService },
+      // Inject in-memory storage for tests, in place of local storage and ession storage.
```
dorota-joanna, line 57:
```suggestion
      // Inject in-memory storage for tests, in place of local storage and session storage.
```

src/app/donation.service.ts
```diff
@@ -232,17 +232,14 @@ export class DonationService {
       1,
     );
 
-    this.storage.set(this.storageKey, donationCouplets);
     this.sessionStorage.set(this.storageKey, donationCouplets);
   }
 
+  /**
+   * Removes donations that were set in local storage before we switched ot using session storage.
```
dorota-joanna, line 239:
```suggestion
   * Removes donations that were set in local storage before we switched to using session storage.
```

dorota-joanna (approved):
LGTM!

PR #1246, opened 2023-08-16
Don-868 new stepper gift aid a11y changes

bdsl:
image


NoelLH (approved):
Nice!

(Sorry the `tabindex` change turned out to be the exact opposite of what I suspected! 😅)

PR #1254, opened 2023-08-21
Don 783 cookie consent form

bdsl:
So far the form doesn't do anything useful, so should only be displayed in dev environment - this PR is just for the way we display the form. Subsequent PRs (or additions to this if this isn't merged first) will make the form actually control what cookies we set.

image


src/app/cookiePreference.service.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+      next: value => {
+        expect(value).toBe(true);
+        done();
+      }
+    })
+  })
+
+  it('Should set a cookie to show the user accepts all cookies', done => {
+    // @ts-ignore - we don't need to take all the params, only the ones our SUT will actually pass.
+    mockCookieService.set = (name, value: string, duration: number) => {
```
bdsl, line 45: realised we we also need to set the cookie path and domain so it can be shared across the site and with the wordpress site.

bdsl: Done in [7634540](https://github.com/thebiggive/donate-frontend/pull/1254/commits/7634540caa741778cf883ddaaac4cdef56093a84)

src/app/cookiePreference.service.spec.ts
```diff
@@ -0,0 +1,62 @@
+import {CookiePreferenceService} from "./cookiePreference.service";
+import {CookieService} from "ngx-cookie-service";
+
+describe('CookiePreferenceService', () => {
+
+  const mockCookieService: CookieService = {
+    get: undefined,
+    set: undefined,
+  } as unknown as CookieService;
```
bdsl, line 9: if someone knows a simpler way to do mocking using Jasmine or something I'm happy to change it to that - I couldn't quite see how to to do it and this seems to work.

dorota-joanna (approved):
LGTM!

PR #1255, opened 2023-08-21
DON-579: use new sf api to dysplay dynamic stats

dorota-joanna:
Screenshot from 2023-08-21 18-27-36
Temporarily using SF prod url to display live stats.

### DIFF
--- a/src/app/campaign-stats.model.ts
+++ b/src/app/campaign-stats.model.ts
@@ -9,8 +9,4 @@ export class CampaignStats {
          */
         public totalCampaignCount: number
     ) {}
-
-    formattedTotalRaised = () => ("£" + this.totalRaised.toLocaleString('en-GB'));
-
-    formattedTotalCount = () => this.totalCampaignCount.toLocaleString('en-GB');
   }
--- a/src/app/campaign.service.ts
+++ b/src/app/campaign.service.ts
@@ -181,11 +181,8 @@ export class CampaignService {
     return this.http.get<Campaign>(`${environment.apiUriPrefix}${this.apiPath}/campaigns/slug/${campaignSlug}`);
   }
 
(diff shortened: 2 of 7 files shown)


src/app/campaign-stats.model.ts
```diff
@@ -9,8 +9,4 @@ export class CampaignStats {
          */
         public totalCampaignCount: number
     ) {}
-
-    formattedTotalRaised = () => ("£" + this.totalRaised.toLocaleString('en-GB'));
-
-    formattedTotalCount = () => this.totalCampaignCount.toLocaleString('en-GB');
```
bdsl, line 15: makes sense we can't put functions on here that we want to call because this class isn't actually used at runtime at all, only used in the type checking stage.

src/app/campaign.service.ts
```diff
@@ -182,10 +182,10 @@ export class CampaignService {
   }
 
   getCampaignImpactStats(): Observable<CampaignStats>{
-    // temporarily hardcoded until the endpoint is ready on SF side
-    // TODO: replace with the commented line below
-    return of(new CampaignStats(242_309_804, 15_000));
-    // return this.http.get<CampaignStats>(`${environment.apiUriPrefix}${this.apiPath}/campaigns/stats`)
```
bdsl, line 188: I think before we merge this we should switch to using a route resolver in the component instead of using the observable directly - see e.g. https://www.digitalocean.com/community/tutorials/angular-route-resolvers

bdsl: Also we probably need to fix the staging version of the API in SF + maybe think about what the frontend should do in case of an error in salesforce.

bdsl (commented):
Added comments inline, otherwise all LGTM.

bdsl (approved):

PR #1257, opened 2023-08-22
DON-865: Use SVG arrow instead of text arrow as list style to stop sc…

bdsl:
…reen reader reader it as text

SVG file downloaded from https://iconmonstr.com/arrow-right-lined-svg/ and then modified - has a permissive licence and it's so simple I'm not sure it's copyright anyway.


src/assets/images/arrow-right.svg
bdsl, line 1: Could download an svg from https://fontawesome.com/icons/arrow-right?f=classic&s=regular instead - but I tried and I don't think it looks any better than this one.

NoelLH (approved):

PR #1262, opened 2023-08-23
DON-783: Only track users with Matomo if they consent to marketing co…

bdsl:
…okies.

We could tell Matomo to remember the consent, using the `rememberConsentGiven` but since we're remembering it already with our own `cookie-preferences` I think it's simpler to rely on that, and just re-send the consent to Matomo every time the app loads.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -13,6 +13,7 @@ import {environment} from "../environments/environment";
 import {flags} from "./featureFlags";
 import {CookiePreferenceService} from "./cookiePreference.service";
 import {Subscription} from "rxjs";
+import {MatomoTracker} from "ngx-matomo";
 
 @Component({
   selector: 'app-root',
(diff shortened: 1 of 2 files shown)


src/app/app.module.ts
```diff
@@ -46,7 +47,8 @@ const matomoTrackers = environment.matomoSiteId ? [
       trackers: matomoTrackers,
       routeTracking: {
         enable: true,
-      }
+      },
+        requireConsent: flags.cookieBannerEnabled,
```
bdsl, line 51: This means we won't send tracking data to Matamo at all without user consent. There is also an option to send tracking data without setting cookies, but I think we want to avoid both.

src/app/app.component.ts
```diff
@@ -78,9 +80,12 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
         this.marketingCookieOptInSubscription = this.cookiePreferenceService.userOptInToMarketingCookies().subscribe(() => {
           this.getSiteControlService.init();
           this.marketingCookieOptInSubscription?.unsubscribe();
+          this.matomoTracker.setCookieConsentGiven();
         });
       } else {
         this.getSiteControlService.init();
+        // no-need to simulate user consent for matomo here, if the banner isn't enabled we don't have
+        //  `requireConsent: true` in the matomo config.
```
bdsl, line 88:
```suggestion
        //  `requireCookieConsent: true` in the matomo config.
```

PR #1266, opened 2023-08-25
DON-811: Delete old version of donation start form

bdsl:
Should make no visible difference to the site, just removing the old version of the donation form, and getting rid of a lot of code to simplify maintenance. It makes it explicit that any future changes only need to be done on the new version of the form, which will mean we will lose the ability to easily revert to the old version.

image


src/app/donation-start/donation-start-form/donation-start-form.component.spec.ts
```diff
@@ -21,12 +21,18 @@ import { of } from 'rxjs';
...
+import {DonationStartFormComponent} from "./donation-start-form.component";
 
-describe('DonationStartPrimaryComponent', () => {
+/**
+ *  This file is currently identical to donation-start-form.component.spec.ts other than using the
+ *  new redesigned form component. Ideally I'd like to just have one file to test both, but since
+ *  we'll hopefully be deleting the old component soon anyway it isn't a big concern to have some
+ *  duplicate test code.
+ */
+
```
bdsl, line 34:
```suggestion
```

dorota-joanna (approved):

PR #1268, opened 2023-08-29
DON-855: Migrate to stripe payment element

bdsl:
Must not be deployed to any environment in advance of https://github.com/thebiggive/matchbot/pull/586

image


src/app/stripe.service.ts
```diff
@@ -55,35 +55,30 @@ export class StripeService {
...
+          family: 'Euclid Triangle',
+          src: `url('${environment.donateGlobalUriPrefix}/d/EuclidTriangle-Regular.1d45abfd25720872.woff2') format('woff2')`,
+          weight: '400',
+        },
+      ],
+      mode: 'payment',
+      currency: donation.currencyCode.toLowerCase(),
+      amount: amountInMinorUnit,
+      setup_future_usage: 'on_session',
+      on_behalf_of: campaign.charity.stripeAccountId,
```
bdsl, line 80: `on_behalf_of` has to be set here to match how it is already set when matchbot creates the payment intent, otherwise we get an error from stripe: `The provided on_behalf_of (acct_xxx) does not match the expected on_behalf_of (null). Try confirming with a Payment Intent that is configured to use the same parameters as Stripe Elements` when pressing the pay button.

Along with the future usage, it makes this message appear within the payment element:

image

bdsl: (discovered later that we can turn off display of the terms within the stripe payment element, e.g. if we want to rely on our own copy instead)

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -642,37 +649,11 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     // Jump back if we get an out of band message back that the card is *not* valid/ready.
     // Don't jump forward when the card *is* valid, as the donor might have been
     // intending to edit something else in the `payment` step; let them click Next.
-    if (!this.donation || !this.stripePaymentMethodReady || !this.card) {
+    if (!this.donation || !this.stripePaymentMethodReady || !this.stripePaymentElement || !this.stripeElements) {
       this.jumpToStep('Payment details');
 
       return;
     }
-
```
bdsl, line 650: I'm not sure if/why we needed to explicitly create stripe payment
    methods. Stripe says that:
    
            In most integrations, you will not need to use this method [createPaymentMethod].
            Instead, use methods like stripe.confirmCardPayment, which
            will automatically create a PaymentMethod when you confirm a PaymentIntent.
            https://stripe.com/docs/js/payment_methods/create_payment_method
    
    I tried calling createPaymentMethod by passing the elements instance,
    but it gives an error saying you have to call elements.submit first,
    which I think we don't want to do before the donor presses the pay now
    button.

dorota-joanna (commented):
Code looks good, but I'm getting this error in the console:
Screenshot from 2023-08-31 17-52-35

bdsl:
I just tried donating again on my local and was able to make the donation multiple getting any error like you saw. Not sure what else to try. Maybe try again on your local @dorota-joanna ? Or once @NoelLH has reviewed we can deploy to staging if we can't find anything to fix in the code and test it there.

bdsl:
Tested with the additional test cards from stripe:

`4000 0025 0000 3155` requires authentication. Payment goes through fine with authentication pass, with authentication failure it scrolls back up to the payment element on pressing pay button and shows a fine error message;

image


----

`4000 0000 0000 9995` declines with insufficient funds. Error shows as expected: "Payment processing failed: Your card has insufficient funds. "

`6205 500 0000 0000 0004` is a longer card number. Payment goes through OK.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
NoelLH: @bdsl This sounds very closely related to what you mentioned about collecting billing address info like `country`. I think that if we kept this as it was, we could hide the address fields.

This lets us provide a smoother journey because we can pre-fill country from CDN header and postcode from the Gift Aid step.

src/app/stripe.service.ts
```diff
@@ -391,4 +361,42 @@ export class StripeService {
...
+    //       },
+    //       email: donation.emailAddress,
+    //       phone: ''
+    //     }
+    // };
+
+    return await this.stripe?.confirmPayment({
+      elements: elements,
+      clientSecret: donation.clientSecret as string,
+      redirect: 'if_required',
```
NoelLH, line 395: I'm a bit nervous about the possibility that we could end up with different, regional payment method types enabled on the dashboard that we have never tested or specifically coded for.

It seems like maybe this is the only way the new element can be used though – is that your understanding @bdsl?

Have we done any testing of a method that redirects if so?

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1010,13 +998,30 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
-      this.card.mount(this.cardInfo.nativeElement);
-      this.card.on('change', this.cardHandler);
+    this.stripePaymentElement = this.stripeElements.create(
+        "payment",
+        {
+          terms: {
+            card: "never" // we have our own terms copy for the future payment in donation-start-form.component.html
+          },
+          fields: {
+            billingDetails: 'auto', // we could set 'never' here, but then we'd have to pass all the details in our `confirmPayment` call, in more detail than we collect.
```
NoelLH, line 1016: Per my other comment, I'm not sure about this – it definitely feels like the same rules ought to apply and I'd like to be 100% sure and see what error we get, before we start replacing our better-optimised address fields with the Stripe ones.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1010,13 +998,30 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+    }
+
     // Card element is mounted the same way regardless of donation info. See
     // this.createDonationAndMaybePerson().subscribe(...) for Payment Request Button mount, which needs donation info
     // first and so happens in `preparePaymentRequestButton()`.
-    this.card = this.stripeService.getCard();
-    if (this.cardInfo && this.card) { // Ensure #cardInfo not hidden by PRB success.
-      this.card.mount(this.cardInfo.nativeElement);
-      this.card.on('change', this.cardHandler);
+    this.stripePaymentElement = this.stripeElements.create(
```
NoelLH, line 1009: Should we set a `paymentMethodOrder` so that e.g. wallet placement is predictable over time? I would suggest putting wallets first, then `card`.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1360,7 +1365,11 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
       complete: () => {},
     };
 
-    this.paymentRequestButton = this.stripeService.getPaymentRequestButton(donation, paymentRequestResultObserver);
+    if (! this.stripeElements) {
+      throw new Error("Stripe elements not ready");
+    }
+
+    this.paymentRequestButton = this.stripeService.getPaymentRequestButton(donation, this.stripeElements, paymentRequestResultObserver);
```
NoelLH, line 1372: Wallets seem to be switched on inside this PR's use of Payment so far. So shouldn't we also remove the PRB completely?

Or, if we want a smaller first step, we could turn off wallets in the new element for now.

src/app/stripe.service.ts
```diff
@@ -391,4 +361,42 @@ export class StripeService {
...
+  {
+    const {error: submitError} = await elements.submit();
+    if (submitError) {
+      console.error(submitError); // @todo handle this error better.
+      return;
+    }
+
+    // const donorName = donation.firstName || '' + ' ' +  donation.lastName;
+
+    // If we want to not show billing details inside the Stripe payment element we have to pass billing details
```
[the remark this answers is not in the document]
bdsl: @NoelLH Yes I did - each of those lines in the commented out code I added when I got an error saying it was required. After a while I gave up and commented it all out.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
NoelLH: As discussed, the docs strongly suggest `submit()` is something we can/should do around this point in the journey. So something close to what we had before seems right.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
NoelLH: For now, we'll probably add a `wallets` key to turn them off (keeping PRBs as the way we support wallets), as the smallest next step.

src/app/stripe.service.ts
```diff
@@ -55,35 +54,30 @@ export class StripeService {
...
     }
 
-    // See https://stripe.com/docs/js/payment_methods
-    const result = await this.stripe.createPaymentMethod({
-      type: 'card',
-      card: cardElement,
-      billing_details: {
-        name: donorName,
-      },
+    const amountInMinorUnit = Math.floor((donation.tipAmount + donation.donationAmount) * 100);
```
NoelLH, line 65:
```suggestion
    const amountInMinorUnit = Math.round((donation.tipAmount + donation.donationAmount) * 100);
```

src/app/stripe.service.ts
bdsl: Yes I think this is what the new element makes us do - butting redirect: 'never' would be a type error: 
image

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: Looks like trying to manually create the payment method if we don't allow the payment element to collect billing details has the same problem as trying to confirm the payment directly without - we get errors asking us to send stripe info we don't have, like donor phone number etc:

image


[diff.txt](https://github.com/thebiggive/donate-frontend/files/12513362/diff.txt)

bdsl: We could configure the payment element to not ask for country and postcode, but again if we try to create the payment method as soon as its filled in we have to pass these, and the way we've arranged the form we wouldn't expect them to be filled in until *after* the payment element part:

image

The country is filled but just with a default value that won't generally be accurate.

image

bdsl: Yes we could get country from the header, although that seems odd when we already have country drop-down. We can't rely on the gift-aid step for postcode since that only asks for postcode if gift aid is turned on.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: I've seen you can go into details with this - so I've now set just postcode and country to 'never' and left all the other billing details on their default setting where Stripe asks for them if it wants to. In my tests it doesn't seem to want to.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: Wallets are now turned off.

NoelLH (approved):
Nice, changes look good to me

PR #1270, opened 2023-09-01
DON-783: Replace youtube with vimeo

bdsl:
Had to faff a bit with the thumbnail in the Vimeo settings for this video to get it looking right.

Before:

image

After:

image


dorota-joanna (dismissed):
LGTM

src/app/home/home.component.html
```diff
@@ -120,7 +120,7 @@
     <div style="padding: 50px 0 40px 0;">
       <biggive-video-feature
         default-text-colour="white"
-        video-url="https://www.youtube.com/embed/zMnzMXxRGPU"
+        video-url="https://player.vimeo.com/video/860172535?dnt=1"
```
bdsl, line 123: We don't need the do not track thing here, it gets automatically added by the component at https://github.com/thebiggive/components/blob/675104de579e83f73dfc0707cd598d15a33d6961/src/util/video.ts#L7

```suggestion
        video-url="https://player.vimeo.com/video/860172535"
```

PR #1273, opened 2023-09-04
DON-783: Only use matomo ecomerce tracking if we have marketing consent

bdsl:


src/app/conversionTracking.service.ts
```diff
@@ -2,21 +2,30 @@ import {MatomoTracker} from "ngx-matomo";
...
   }
 
+
   public convert(donation: Donation, campaign: Campaign) {
     this.trackConversionWithMatomo(donation, campaign);
   }
 
   private trackConversionWithMatomo(donation: Donation, campaign: Campaign) {
+    if (! this.marketingCookiesAllowed) {
+      return;
```
bdsl, line 26: Although we're not using the matomo eccomerce tracking feature in this case we can still count donations via matomo using the `thank_you_fully_loaded` event that we send from https://github.com/thebiggive/donate-frontend/blob/02727e2a1c9c8231c7edd5001d18d601e017348d/src/app/donation-complete/donation-complete.component.ts#L230

dorota-joanna (dismissed):
LGTM

PR #1279, opened 2023-09-06
Don 783 cookie three radios

bdsl:
Visually this is as shown at https://github.com/thebiggive/components/pull/294 . I've tested manually in firefox and confirmed that the cookie preferences are applied correctly to matomo and get site control.


dorota-joanna (approved):
LGTM

PR #1282, opened 2023-09-11
DON-822: Add unit test for paying with stripe with donation funds

bdsl:


src/app/donation-start/donation-start-form/donation-start-form.component.spec.ts
```diff
@@ -1,29 +1,102 @@
...
+    undefined as unknown as ChangeDetectorRef,
+    {
+      convert: () => {
+      }
+    } as unknown as ConversionTrackingService,
+    undefined as unknown as MatDialog,
+    donationService,
+    undefined as unknown as ElementRef<any>,
+    undefined as unknown as FormBuilder,
+    undefined as unknown as IdentityService,
```
NoelLH, line 47: Does this method of instantiation require all these @bdsl, or is it just a preference to make them explicitly `undefined`?

NoelLH (commented):
Mostly looks good, couple of Qs inline

src/app/donation-start/donation-start-form/donation-start-form.component.spec.ts
bdsl: I'm not sure I follow the question - the constructor of the component class requires all these arguments. 'as unknown as <type>' is just how we force the compiler to pretend that undefined is the required type. We know that for these particular tests those arguments are not used so we can just use undefined as a dummy value.

src/app/donation-start/donation-start-form/donation-start-form.component.spec.ts
```diff
@@ -386,6 +459,51 @@ describe('DonationStartNewPrimaryComponent', () => {
     expect(component.donationForm.controls.marketing!.get('optInTbgEmail')?.errors).toBeNull();
   });
 
+
+  it('Should allow paying with donation funds with no saved payment method', async () => {
+    var finaliseCashBalancePurchaseCalled = false;
+
+    const fakeDonationService = {
+      getDefaultCounty: () => 'gb',
```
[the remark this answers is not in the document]
bdsl: @NoelLH Thanks, I added documentation to the real function and adjusted in 14120eaa.

src/app/donation-start/donation-start-form/donation-start-form.component.spec.ts
NoelLH: Makes sense, I'd forgotten we had all these args in a constructor!

src/app/donation-start/donation-start-form/donation-start-form.component.spec.ts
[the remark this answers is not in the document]
NoelLH: Nice, thanks!

NoelLH (approved):

PR #1283, opened 2023-09-11
DON-870: Thank you page accessibility tweeks

bdsl:
…ation is complete

Ideally I'd like to just have the page title be read out and do this job, but I couldn't see a way to do that. Might be worth having an sr-only copy of the page title on all pages in the app?

### DIFF
--- a/src/app/donation-complete/donation-complete.component.ts
+++ b/src/app/donation-complete/donation-complete.component.ts
@@ -1,6 +1,6 @@
 import { HttpErrorResponse } from '@angular/common/http';
 import { Component, Input, OnInit, ViewChild } from '@angular/core';
-import { DomSanitizer, SafeHtml } from "@angular/platform-browser";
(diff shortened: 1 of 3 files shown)


src/app/donation-complete/donation-complete.component.ts
```diff
@@ -219,8 +219,8 @@ export class DonationCompleteComponent implements OnInit {
     this.campaignService.getOneById(donation.projectId).subscribe(campaign => {
       this.campaign = campaign;
       this.pageMeta.setCommon(
-        `${campaign.charity.name}`,
-        `Thanks for donating to the "${campaign.title}" campaign`,
+        `Thank you for donating to the "${campaign.title}" campaign`,
```
bdsl, line 222: I was hoping this might be auto read by the screen reader. It doesn't seem to be me for me. Still we might as well have "Thank you" in the title.

NoelLH (approved):

PR #1285, opened 2023-09-13
Stop passing `donateLinkId` around

NoelLH:
Groundwork for dropping this deprecated, redundant property from all apps.

And other minor refactoring.

### DIFF
--- a/src/app/campaign-details/campaign-details.component.spec.ts
+++ b/src/app/campaign-details/campaign-details.component.spec.ts
@@ -57,7 +57,6 @@ describe('CampaignDetailsComponent', () => {
       {
         id: 'testCharityId',
         name: 'Test Charity',
-        donateLinkId: 'SFIdOrLegacyId',
         optInStatement: 'Opt in statement.',
         website: 'https://www.testcharity.co.uk',
         regulatorNumber: '123456',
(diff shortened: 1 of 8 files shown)


bdsl (approved):
LGTM - nothing is reading the donateLinkId so assuming its not read dynamically getting rid of it is a nice simplification.

I think this PR isn't actually required to make anything work at runtime, as we don't have any production code that calls the Campaign constructor. We only do `new Campaign` in tests. Afaik the Campaign class isn't used at all at production runtime, its only used at compile time and in tests.

NoelLH:
> LGTM - nothing is reading the donateLinkId so assuming its not read dynamically getting rid of it is a nice simplification.
> 
> I think this PR isn't actually required to make anything work at runtime, as we don't have any production code that calls the Campaign constructor. We only do `new Campaign` in tests. Afaik the Campaign class isn't used at all at production runtime, its only used at compile time and in tests.

Yep you may be right – I'm not totally sure what happens when a campaign's Charity is cast to the model if it's missing an expected property. But I think tidying up makes sense either way.

bdsl:
> I'm not totally sure what happens when a campaign's Charity is cast to the model

I'm sure nothing happens. The mention of the Charity class is on the type level only, it gets erased when the typescript is converted to Javascript. The error would only happen at runtime if a programmer reads the Charity type and creates some code that relies on one of the properties. But that is a real risk if the class doesn't accurately describe the shape of the objects, so it makes sense to keep it up to date.

At another time it might be worth thinking about using either using Zod or another runtime validation library that checks they match, or manually piping the observable through a call to the class constructor.

PR #1286, opened 2023-09-13
Don-883: Reset donation form if donor goes back to first step

bdsl:
Going back to the first step indicates that they may be about to change the donation amount, and we have a bug that means changing the amount makes the confirm payment step at the end fail. This avoids that, at the (high but worth paying) cost of erasing what they've typed into the form. 

It's not a good solution by any means, but I've now spent a full day trying to find a better solution, and not found one. This is a lot better than nothing, and doesn't stop us looking for a better solution.

The one fix I managed to find for the original bug has the side effect of requiring the donor to type in their payment details a 3rd time (after the expected 2nd time that happens because they changed the amount). That's confusing and I think worse than the change here.

The only other option I can think of if we don't find a better fix is to revert to the old stripe card element instead of the payment element.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -9,7 +9,7 @@
       <mat-step
         formGroupName="amounts"
         [stepControl]="amountsGroup"
-        label="Your donation"
+        [label]="yourDonationStepLabel"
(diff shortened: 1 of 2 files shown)


bdsl:
I think this would be enough to let us revert https://github.com/thebiggive/regression-tests/pull/116

NoelLH (approved):

PR #1287, opened 2023-09-14
DON-882: Cancel donation before reloading stepper at first step

bdsl:
This should avoid the problem we're seeing in regression tests of being prompted to continue the existing donation, which would be a bit confusing and also makes the tests fail


dorota-joanna (approved):
LGTM

PR #1289, opened 2023-09-14
DON-884 – vary campaign highlights figures for campaigns which share funds

NoelLH:
Also refactor to make `CampaignInfoComponent` standalone and remove duplication of its functionality.

Because it became essentially a single-line wrapper around another component and was only used once, I also retired `DonationStartSecondaryComponent` to simplify and make things smaller.


NoelLH:
I think there's a tiny regression in the icon layout in the highlights card, marking draft for a moment while I have a look.

NoelLH:
> I think there's a tiny regression in the icon layout in the highlights card, marking draft for a moment while I have a look.

Should be sorted. I think something about the component nesting updates made the icons need centring explicitly – not totally clear why but it seems to work.

src/app/campaign-info/campaign-info.component.html
```diff
@@ -1,21 +1,20 @@
 <div>
-  <!-- Todo: componentise this -->
   <biggive-campaign-highlights
     space-below=0
-    primary-figure-label="Amount raised"
+    [primaryFigureLabel]="primaryFigureLabel"
```
bdsl, line 4: I think introduce the property in the class probably makes things harder to follow, since then to understand this line you need to search for everywhere that writes to the primaryFigureLabel, and find there's the value set on assignment and the other value set in ngOnInit - and unless you search you don't know that there aren't more.

I think better to keep the object simpler and do this logic in the template:

```suggestion
    [primaryFigureLabel]="this.campaign.parentUsesSharedFunds ? 'Overall total raised' : 'Amount raised'"
```

Same applies to secondary figure label.

src/app/campaign-info/campaign-info.component.ts
```diff
@@ -28,6 +35,9 @@ export class CampaignInfoComponent implements OnInit {
   campaignFinished: boolean;
   campaignRaised: string; // Formatted
   campaignTarget: string; // Formatted
+  donationCount: number;
```
bdsl, line 38: I think this could use a docblock. Had a go writing one, the fact that I have an 'or' in it makes me wonder if confusing to have this as a property as well. Maybe there's a better way to document that makes it more intuitive:

```suggestion
  /**
   * May be the count of donations to this campaign, or the to the parent campaign.
   */
  donationCount: number;
```

src/app/campaign-info/campaign-info.component.ts
```diff
@@ -42,8 +52,26 @@ export class CampaignInfoComponent implements OnInit {
...
     this.campaignOpen = CampaignService.isOpenForDonations(this.campaign);
     this.campaignFinished = CampaignService.isInPast(this.campaign);
-    this.campaignTarget = this.currencyPipe.transform(this.campaign.target, this.campaign.currencyCode, 'symbol', '1.0-0') as string;
-    this.campaignRaised = this.currencyPipe.transform(this.campaign.amountRaised, this.campaign.currencyCode, 'symbol', '1.0-0') as string;
+    this.campaignTarget = this.currencyPipe.transform(
+      this.campaign.parentUsesSharedFunds ? this.campaign.parentTarget : this.campaign.target,
+      this.campaign.currencyCode,
+      'symbol',
+      '1.0-0',
+    ) as string;
```
bdsl, line 58: do you know why we need the `as string` here? I'm not sure why currencyPipe,transform returns a nullable.

src/app/campaign-info/campaign-info.component.ts
NoelLH: Thanks @bdsl – I've pushed one which I think does more to explain the options, but let me know if you think it should be different.

src/app/campaign-info/campaign-info.component.ts
bdsl: I see the `as string` is in the old version of the code just as much though so this shouldn't be a blocker to the PR.

NoelLH: Not entirely sure @bdsl, this is an existing bit I just reformatted. I think maybe pipes give null output from null input, but because of the resolver and snapshot we know that our input isn't null if we got this far?

src/app/campaign.model.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @link https://app.swaggerhub.com/apis/Noel/TBG-Campaigns/#/Campaign
+ */
```
bdsl, line 3: hmm OK. I'd much rather have it all documented here with docblocks - without looking I don't know whether or not I have a way to edit that documentation. But given that it exists it's good to link to it.

NoelLH: Yeah, as I mentioned elsewhere today @bdsl I think the ideal would be to generate it from one source and inline language-specific outputs in all the apps that use it. But I really don't want to start doing this manually since the properties should mean the same thing in all places.

I think for now this gives a decent way to see the model's expected usage in detail. (If we're sticking with it, we probably need the same contract to live on a more affordable multi-user solution than SwaggerHub.)

src/app/campaign.model.ts
```diff
@@ -42,13 +46,17 @@ export class Campaign {
...
     public alternativeFundUse?: string,
     public campaignCount?: number,
     public championOptInStatement?: string,
     public championRef?: string,
     public feePercentage?: number,
     public logoUri?: string,
+    public parentAmountRaised?: number,
+    public parentDonationCount?: number,
     public parentRef?: string,
+    public parentTarget?: number,
```
bdsl, line 59: are these three new fields always non-null if and only if parentUsesSharedFunds is true? In which case we could make that clear in the type system by putting them together in optional subj-object?

Something like:

```diff
     public isMatched: boolean,
     public matchFundsRemaining: number,
     public matchFundsTotal: number,
-    public parentUsesSharedFunds: boolean,
     public problem: string,
     public quotes: Array<{person: string, quote: string}>,
     public ready: boolean,
@@ -53,6 +52,11 @@ export class Campaign {
     public championRef?: string,
     public feePercentage?: number,
     public logoUri?: string,
+    public sharedFundsParent?: {
+        amountRaised: number,
+        donationCount: number,
+        target: number,
+    },
     public parentAmountRaised?: number,
     public parentDonationCount?: number,
     public parentRef?: string,
```

src/app/campaign-info/campaign-info.component.ts
bdsl: Right - I hadn't kept track of the fact that its not new.

src/app/donation-start/donation-start-secondary/donation-start-secondary.component.html
bdsl, line 1: The secondary component has been dead since we deleted the old style donation stepper right? I must have missed it when I deleted the main part of the old stepper.

bdsl (dismissed):
Reviewed - no real blockers but I think if we could make the change I suggest at https://github.com/thebiggive/donate-frontend/pull/1289/files#r1326224857 it would make the component code quite a bit more intuitive. Obviously that change has to be made in SF as well as here.

src/app/donation-start/donation-start-secondary/donation-start-secondary.component.html
NoelLH: @bdsl I think we are using it prior to this PR, but I trimmed it down to the point that it was a very thin wrapper around the info component. So I thought I would take the next step and delete it.

src/app/campaign.model.ts
NoelLH: @bdsl I think they are supplied whenever applicable, since that's both faster on the SF side and lets us use the data in other ways later if we want. (e.g. a donate page could show both totals)

So we could do this and call it `parentCampaign` possibly, but as I mentioned elsewhere it feels like it's a limited benefit for this many properties.

src/app/campaign.model.ts
```diff
@@ -42,13 +46,17 @@ export class Campaign {
     public target: number,
```
bdsl, line 46: does target need to become nullable at this point if a campaign that uses shared funds doesn't have a target of its own? Or even better following on from https://github.com/thebiggive/donate-frontend/pull/1289/files#r1326224857 we could use a union type to say that a campaign either has its own target or it has parent campaign stuff.

NoelLH: I think that the reality in the data is that the children still have & surface a "target", which we can see being used on a [Staging campaign of this type](https://donate-staging.thebiggivetest.org.uk/global-2) now – it's just not very meaningful.

I haven't added conditionals that would slow down Salesforce so that will remain the case.

src/app/campaign.model.ts
bdsl: ah ok. I was meaning to go and have another look at that SF PR. It's less useful if we still need the boolean.

src/app/campaign.model.ts
bdsl: OK - maybe worth making it nullable here anyway so that SF is free to stop sending it in future in cases where it isn't meaningful, and to remind us working in frontend to not expect to always be able to (meaningfully) find a target on a campaign object?

NoelLH: Makes sense – done on 6f1b592c3939144b760be2a32bc2a788410fa183

PR #1290, opened 2023-09-15
DON-885 – simplify grid on shared fund meta-campaign

NoelLH:
Make shared fund / emergency campaigns' child campaign cards hide N/A figures and progressbars.

Update components to support this, and all other libs where there are minor version updates.

New grid in simplified mode on local


bdsl (approved):
LGTM

PR #1297, opened 2023-09-19
DON-614 – move e2e framework to Cypress

NoelLH:
Bootstrapped via https://www.npmjs.com/package/@cypress/schematic – details of subsequent steps in commit messages.

Component test bootstrapping skipped for now since that doesn't need replacing.

Just wrote one very basic "app boots" test, analogous to what we once had with Protractor, for now.

Since we are unlikely to frequently work on the tests at this level right now, I went with making the configuration non-watching and headless in all cases, rather than configuring CI differently. It seems to work without more advanced config both locally and in CI: 🎉 

Screenshot 2023-09-19 at 15 54 19


bdsl:
Not sure why [3ad4a07](https://github.com/thebiggive/donate-frontend/pull/1297/commits/3ad4a07385b97944a4b3faa547147f832df6936a) failed checks - I've reverted the revert of it and that passes. Also set CircleCI to store screenshots as artefacts in case a cyprus test fails in future, so it should be easier to tell why then.

bdsl (approved):
LGTM! Cyprus looks a lot nicer to use than protractor.

NoelLH:
> Not sure why [3ad4a07](https://github.com/thebiggive/donate-frontend/pull/1297/commits/3ad4a07385b97944a4b3faa547147f832df6936a) failed checks - I've reverted the revert of it and that passes. Also set CircleCI to store screenshots as artefacts in case a cyprus test fails in future, so it should be easier to tell why then.

Interesting – maybe the 4s default timeout is a little aggressive for a cold boot as things stand? We can keep an eye on it.

I think we also want to store artifacts from `cypress`, fixed in https://github.com/thebiggive/donate-frontend/pull/1300

PR #1301, opened 2023-09-20
DON-614: Intercept HTTP call to SF API in cypress tests

bdsl:
Makes the frontend a little more self-contained, although we lose some incidental testing of the SF API


dorota-joanna (approved):
LGTM

PR #1302, opened 2023-09-20
DON-893: Ask Matchbot to confirm donation payments instead of c…

bdsl:
…alling stripe directly

Not to be merged until matchbot has that capability of course


NoelLH (commented):
Looking good!

I think considering saved methods too would be great to do now if possible, then we'll be in a good place to think about removing `donation.clientSecret` completely as a default property that gets passed around – sending it from `/confirm` only where the client specifically needs it for a _next_action_.

I think that change will be in `StripeService.payWithMethod()` and look very similar to what you've done for new payment methods.

bdsl:
@NoelLH Thanks, done. That allowed me to remove the `clientSecret` property from the Donation interface.

NoelLH:
Nice, thanks @bdsl. I made a PR into this branch with a couple of small suggestions: https://github.com/thebiggive/donate-frontend/pull/1305

But I think it's pretty much good to go

NoelLH (approved):

PR #1303, opened 2023-09-20
Don't build a `//` URI where the real app doesn't

NoelLH:
It seems like due to Chrome behaviour(?) Cypress "eats" the extra slash, but it could confuse devs comparing it to how the real app builds this.


bdsl:
Thanks!

PR #1305, opened 2023-09-25
DON-893 tweaks

NoelLH:
A few small proposed tweaks to https://github.com/thebiggive/donate-frontend/pull/1302

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -772,13 +772,13 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
       throw new Error("Missing stripe elements");
     }
 
-    let confirmResult;
+    let paymentMethodResult;
     let paymentMethod;
     if (this.selectedSavedMethod) {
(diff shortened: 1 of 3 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -772,9 +772,14 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
       throw new Error("Missing stripe elements");
     }
 
-    const paymentMethod = this.selectedSavedMethod || (
-      await this.stripeService.confirmPaymentWithPaymentElement(this.donation, <StripeElements>this.stripeElements)
-    ).paymentMethod;
+    let paymentMethodResult;
+    let paymentMethod;
```
bdsl, line 776: when I hover these the type shows up as `any` - do you know why that doesn't fail the build since we have `"noImplicitAny": true,` in `tsconfig.app.json` ? Maybe it's OK for it to be `any` here because better types are inferred as soon as the assignments are done and that's very close.

NoelLH: Mm, yeah good question – not sure I'm afraid

PR #1314, opened 2023-09-26
DON-883 – keep Stripe.js in sync with the Donation

NoelLH:
And rationalise what we reset, leaving an already valid Payment Element
and its status tracking properties untouched where possible.

Plus related small tweaks & fixes.

### DIFF
--- a/src/app/stripe.service.ts
+++ b/src/app/stripe.service.ts
@@ -42,7 +42,7 @@ export class StripeService {
       ],
       mode: 'payment',
       currency: donation.currencyCode.toLowerCase(),
-      amount: this.getAmountInMinorUnit(donation),
+      amount: this.amountIncTipInMinorUnit(donation),
       setup_future_usage: 'on_session',
       on_behalf_of: campaign.charity.stripeAccountId,
       paymentMethodCreation: 'manual',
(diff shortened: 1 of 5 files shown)


src/app/stripe.service.ts
```diff
@@ -45,13 +42,17 @@ export class StripeService {
...
       currency: donation.currencyCode.toLowerCase(),
-      amount: amountInMinorUnit,
+      amount: this.getAmountInMinorUnit(donation),
       setup_future_usage: 'on_session',
       on_behalf_of: campaign.charity.stripeAccountId,
       paymentMethodCreation: 'manual',
     });
   }
 
+  updateAmount(elements: StripeElements, donation: Donation) {
```
bdsl, line 52: Nice find - looks good based on the docblock in StripeElements.update . I couldn't find docs on the stripe website, but I don't think we should need that.

bdsl (commented):
All LGTM other than where I've commented - wish I'd found the update function before!

src/app/stripe.service.ts
NoelLH: It was mentioned in a step inside the finalise on server doc page – which I don't think has a working permalink to the section frustratingly. So it might well be something that's important iff you confirm that way.

bdsl (approved):

PR #1316, opened 2023-09-27
DON-874: Advertise Arts For Impact from the 9th of October

bdsl:
As of now this keeps the homepage unchanged like so:

image

but by changing the date mentioned at line 63 to one in the past it appears like so:

image


dorota-joanna (commented):
LGTM

PR #1317, opened 2023-09-28
Don 819 ameliorate disabled buttons

bdsl:
WIP - we've set up Snack Bar to display errors for the first step. 

Todo: Similar for all other continue buttons where they can have errors -  steps 2 to 5

Consider moving the snackbar away from the bottom of the screen and making it more prominent.

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -55,7 +55,19 @@
           aria-live="polite"
           *ngIf="donationAmountField?.invalid && donationAmountField?.errors && (donationAmountField?.dirty || donationAmountField?.touched)"
         >
+        <div *ngIf="don819FlagEnabled">
           {{displayableAmountsStepErrors()}}
+        </div>
+
+        <div *ngIf="!don819FlagEnabled">
+          <div *ngIf="donationAmountField?.errors?.min">Sorry, the minimum donation is {{ currencySymbol }}1.</div>
+          <div *ngIf="donationAmountField?.errors?.max">Your donation must be {{ maximumDonationAmount  | exactCurrency:campaign.currencyCode }} or less to proceed.<span *ngIf="creditPenceToUse === 0"> You can make multiple matched donations of {{ maximumDonationAmount | exactCurrency:campaign.currencyCode }} if match funds are available.</span></div>
+          <div *ngIf="donationAmountField?.errors?.max">Your donation must be {{ maximumDonationAmount  | exactCurrency:campaign.currencyCode }} or less to proceed.<span *ngIf="creditPenceToUse === 0"> You can make multiple matched donations of {{ maximumDonationAmount | exactCurrency:campaign.currencyCode }} if match funds are available.</span></div>
+          <div *ngIf="donationAmountField?.errors?.pattern">Please enter a whole number of {{ currencySymbol }} without commas.</div>
+          <div *ngIf="donationAmountField?.errors?.pattern">Please enter a whole number of {{ currencySymbol }} without commas.</div>
+          <div *ngIf="donationAmountField?.errors?.required">Please enter how much you would like to donate.</div>
+          <div *ngIf="donationAmountField?.errors?.required">Please enter how much you would like to donate.</div>
+        </div>
         </div>
 
         <div *ngIf="creditPenceToUse === 0">
@@ -193,6 +205,7 @@
             class="continue b-w-100 b-rt-0"
             mat-raised-button
             color="primary"
+            [disabled]="!donationForm.controls['amounts']!.valid && !don819FlagEnabled"
             (click)="progressFromStepOne()"
           >Continue</button>
       </div>
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -55,7 +55,7 @@
           aria-live="polite"
           *ngIf="donationAmountField?.invalid && donationAmountField?.errors && (donationAmountField?.dirty || donationAmountField?.touched)"
         >
-        {{ displayableAmountsStepErrors( )}}
+        {{ displayableAmountsStepErrors() }}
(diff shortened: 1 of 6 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -196,8 +193,7 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
             class="continue b-w-100 b-rt-0"
             mat-raised-button
             color="primary"
-            [disabled]="! donationForm.controls['amounts']!.valid"
-            (click)="next()"
+            (click)="progressFromStepOne()"
```
bdsl, line 196: Maybe worth feature-flagging all these changes, so we can merge this without having to do all 5 steps at once and get them perfect first time?

Would just need to set up the flag somehow either based on the environment or a URL param or something, then put back the old logic for disabling the button in case the flag is off, like so:

```typescript
            [disabled]="! donationForm.controls['amounts']!.valid && ! don819FlagEnabled"
            (click)="progressFromStepOne()"
```

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -55,10 +55,7 @@
           aria-live="polite"
           *ngIf="donationAmountField?.invalid && donationAmountField?.errors && (donationAmountField?.dirty || donationAmountField?.touched)"
         >
-          <div *ngIf="donationAmountField?.errors?.min">Sorry, the minimum donation is {{ currencySymbol }}1.</div>
-          <div *ngIf="donationAmountField?.errors?.max">Your donation must be {{ maximumDonationAmount  | exactCurrency:campaign.currencyCode }} or less to proceed.<span *ngIf="creditPenceToUse === 0"> You can make multiple matched donations of {{ maximumDonationAmount | exactCurrency:campaign.currencyCode }} if match funds are available.</span></div>
-          <div *ngIf="donationAmountField?.errors?.pattern">Please enter a whole number of {{ currencySymbol }} without commas.</div>
-          <div *ngIf="donationAmountField?.errors?.required">Please enter how much you would like to donate.</div>
+        {{ displayableAmountsStepErrors( )}}
```
bdsl, line 58:
```suggestion
        {{ displayableAmountsStepErrors()}}
```

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -964,7 +968,57 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     }
   }
 
-  onUseSavedCardChange(event: MatCheckboxChange, paymentMethod: PaymentMethod) {
+  progressFromStepOne() {
+
```
bdsl, line 972:
```suggestion
  progressFromStepOne() {
```

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -964,7 +968,57 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+    }
+
+    this.next();
+  }
+
+    public displayableAmountsStepErrors = () => {
+        const errors = this.donationAmountField?.errors;
+
+        if (!errors) {
+            return '';
```
bdsl, line 991: not sure if we really want to return '' here - think we may have put this in as a shortcut to getting the type check to pass last week. Maybe throw an error instead? Maybe that's just changing the ? to ! on 988.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1171,11 +1225,13 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     const firstElInStepWithAngularError = stepJustDone.querySelector('.ng-invalid.ng-touched[formControlName]');
     if (firstElInStepWithAngularError && !this.closeAncestorsHaveDisplayNone(firstElInStepWithAngularError)) {
       this.scrollTo(firstElInStepWithAngularError);
+      console.log(firstElInStepWithAngularError);
```
bdsl, line 1228:
```suggestion
```

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1171,11 +1225,13 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
       return true;
     }
 
     const firstCustomError = stepJustDone.querySelector('.error');
     if (firstCustomError) {
+      console.log(firstCustomError);
```
bdsl, line 1234:
```suggestion
```

src/app/featureFlags.ts
```diff
@@ -6,6 +6,7 @@ export const flagsForEnvironment = (_: EnvironmentID) => {
     // Banner needs more work to look right and be useful before release - see ticket
     // DON-783
     cookieBannerEnabled: (environment.environmentId === 'development' || environment.environmentId === 'staging'),
+    don819FlagEnabled: (environment.environmentId === 'development' || environment.environmentId === 'staging'),
```
bdsl, line 9: Maybe have this only on in dev environment until we want to ask for a non-developer to start reviewing what we've done? I think if it's in staging we might need to make it controllable by the user e.g. via a URL param, so that we can still test the current system with disabled buttons.

```suggestion

    // avoiding use of disabled buttons on donate page where we can and providing more specific error messages instead when we can't continue to the next step. Feature flagged so we can release all steps at once.
    don819FlagEnabled: (environment.environmentId === 'development'),
```

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: ah see it's documented as null if there are no errors, so actually this seems fine.

bdsl (commented):
LGTM! I can't use the approve button as I'm listed as the author if this PR.

dorota-joanna (approved):
LGTM

PR #1321, opened 2023-10-02
DON-889: Add Skip to Content link

bdsl:
This button is visible only when it is focused - at other times it hidden off the edge of the screen.

I used the #focus-within CSS selector, which should work with all browsers except IE and Opera Mini.

image

### DIFF
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -1,6 +1,19 @@
+@import '../abstract';
+
 #cookie-banner {
   position: fixed;
   bottom: 3em;
   right: 3em;
   z-index: 10;
+}
+
+.skip-to-content-link {
+  z-index: 10;
+  position: absolute;
+  transform: translateY(-100%);
+  transition: transform 0.3s;
+}
+
+.skip-to-content-link:focus-within {
+  transform: translateY(0%);
 }
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,3 +1,17 @@
+  <div class="skip-to-content-link">
+    <biggive-button
+      space-above="5"
+      colour-scheme="primary"
+      label="Skip to content"
+      full-width="true"
+      size="medium"
+      rounded="false"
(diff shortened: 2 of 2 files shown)


src/app/app.component.scss
```diff
@@ -1,6 +1,19 @@
...
 }
+
+.skip-to-content-link {
+  z-index: 10;
+  position: absolute;
+  transform: translateY(-100%);
+  transition: transform 0.3s;
+}
+
+.skip-to-content-link:focus-within {
```
NoelLH, line 17: Does this need to be `:focus-within` because an inner shadow DOM containing the button might be what actually receives the focus? When I've implemented similar to this in the past I've pretty much followed [this](https://www.a11yproject.com/posts/how-to-hide-content/) and used simple `:focus`.

bdsl: I don't think it's an issue with the shadow dom, it's just because the div isn't focusable, its the button inside that is focusable.

bdsl: Maybe it is sort of a shadow dom thing, or just an encapsulation thing - if we wanted to apply a css rule on #focus we'd have to target the `a` element inside the button component. Focus-within seems to work well.

bdsl: Also could use a simple a tag instead of the button, but then it's hard to follow the consistent look and feel. I don't that's super important here though.

bdsl:
This is largely based on following https://css-tricks.com/how-to-create-a-skip-to-content-link/ - but I noticed that if I listen to the page in my Ubuntu screen reader without touching the keyboard the skip to button content isn't audible, and I would like it to be. 

So I tried copying https://www.a11yproject.com/posts/how-to-hide-content/ and doing something more similar to what Sky News has, but my Ubuntu screen reader still seems to detect that the button is invisible and therefore not read it out, despite reading out the one on Sky News.

NoelLH (approved):

PR #1324, opened 2023-10-03
DON-901 – retry ID service updates on failure

NoelLH:
Try up to 3 times with a fixed 2s gap. Updates should be harmless to double send and this should reduce the frequency of edge cases where Stripe is missing customer info.

### DIFF
--- a/src/app/identity.service.ts
+++ b/src/app/identity.service.ts
@@ -110,7 +110,7 @@ export class IdentityService {
       `${environment.identityApiPrefix}${this.peoplePath}/${person.id}`,
       person,
       this.getAuthHttpOptions(person),
-    ).pipe(retry(2), delay(2000));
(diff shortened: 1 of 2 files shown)


src/app/identity.service.ts
```diff
@@ -109,7 +110,7 @@ export class IdentityService {
       `${environment.identityApiPrefix}${this.peoplePath}/${person.id}`,
       person,
       this.getAuthHttpOptions(person),
-    );
+    ).pipe(retry(2), delay(2000));
```
bdsl, line 113: Not a blocker, but imho worth using this to make thousands easy to read.
```suggestion
    ).pipe(retry(2), delay(2_000));
```

bdsl (approved):
LGTM, I added one small suggestion.

PR #1328, opened 2023-10-04
Don 819 confirmation step show errors

bdsl:
Also showing an error for failure to create the donation with snackbar:

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -1025,7 +1025,7 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
       {
 
         // formula for duration from https://ux.stackexchange.com/a/85898/7211
-        duration: 50_000, // Math.min(Math.max(message.length * 50, 2_000), 7_000),
+        Math.min(Math.max(message.length * 50, 2_000), 7_000),
         panelClass: 'snack-bar',
       }
     );
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -1022,7 +1022,11 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     this.snackBar.open(
       message,
       undefined,
-      {duration: 5_000, panelClass: 'snack-bar'}
+      {
+        // formula for duration from https://ux.stackexchange.com/a/85898/7211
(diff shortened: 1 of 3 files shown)


dorota-joanna:
This snackbar looks really good, but setting `donateCreationError` doesn't trigger the snack-bar for me for some reason even though the error is visible right above the Continue button:
Screenshot from 2023-10-04 20-50-45

I might be doing something wrong..?
I also noticed that in case of the donation amount errors on step 1, when I correct the amount the snack-bar persists, maybe it would be clearer if it disappeared when the amount is adjusted?

dorota-joanna (commented):
This snackbar looks really good, but setting `donateCreationError` doesn't trigger the snack-bar for me for some reason even though the error is visible right above the Continue button:
Screenshot from 2023-10-04 20-50-45

I might be doing something wrong..?
I also noticed that in case of the donation amount errors on step 1, when I correct the amount the snack-bar persists, maybe it would be clearer if it disappeared when the amount is adjusted?

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -992,7 +1022,12 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     this.snackBar.open(
       message,
       undefined,
-      {duration: 5_000, panelClass: 'snack-bar'}
+      {
+
+        // formula for duration from https://ux.stackexchange.com/a/85898/7211
+        duration: 50_000, // Math.min(Math.max(message.length * 50, 2_000), 7_000),
```
bdsl, line 1028:
```suggestion
        Math.min(Math.max(message.length * 50, 2_000), 7_000),
```

bdsl:
> This snackbar looks really good, but setting donateCreationError doesn't trigger the snack-bar for me for some reason even though the error is visible right above the Continue button

I didn't do anything to make setting donateCreationError automatically trigger the snack-bar, I just added a call to the new `showDonationCreateError` function next to everything that sets `donateCreationError = true`.

bdsl:
> I also noticed that in case of the donation amount errors on step 1, when I correct the amount the snack-bar persists, maybe it would be clearer if it disappeared when the amount is adjusted?

It's supposed to always disappear in between 2 and 7 seconds, depending how long the text is. But I commented that out and changed it to 50 seconds and committed 50 by mistake. Fixed in https://github.com/thebiggive/donate-frontend/pull/1328/commits/51a92482fe62cc4e65ecc17ab74de385a6bcf9f4

dorota-joanna (approved):
LGTM

PR #1333, opened 2023-10-05
Don 819 disabled buttons payment step

dorota-joanna:


src/app/donation-start/donation-start-form/PaymentReadinessTracker.ts
```diff
@@ -0,0 +1,78 @@
...
+ * writing it just returns a boolean from readyToProgressFromPaymentStep, but my plan is to add a function to return
+ * a list of reasons *why* we are not ready for them to proceed that we can display using snackbar.
+ *
+ * Originally extracted from the DonationStartFormComponent
+ */
+export class PaymentReadinessTracker {
+  /**
+   * Property name taken from the component. The name doesn't really reflect the usage. Planning to remove this property.
+   */
+  private stripePaymentMethodReady: boolean = false;
```
bdsl, line 12: these properties are copies of the properties in DonationStartFormComponent. Every time one of the matching properties in DonationStartFormComponent was set previously there should now be a call to a function on this object that sets the property here.

Tomorrow I'll probably at least one of them, maybe more and replace with more specific properties that say exactly what the relevant state is, so we can generate not only the boolean answer but also the list of reasons for a no.

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -444,7 +444,7 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
             class="continue b-w-100 b-rt-0"
             mat-raised-button
             color="primary"
-            [disabled]="(!stripePaymentMethodReady && !selectedSavedMethod) || !paymentGroup.valid"
+            [disabled]="! readyToProgressFromPaymentStep"
```
bdsl, line 447: in the old version this line was confusing because `stripePaymentMethodReady` was set for several unrelated reasons in different scenarios. The logic is now I think much clearer in the implementation of `PaymentReadinessTracker.readyToProgressFromPaymentStep`.

Whether it was necessary to refactor step by step to get there is debatable - maybe we should have just written that from scratch. But this way we can be confident that we're preserving the behaviour.

src/app/donation-start/donation-start-container/donation-start-container.component.spec.ts
```diff
@@ -67,7 +67,7 @@ describe('DonationStartContainer', () => {
     component = fixture.componentInstance;
     fixture.detectChanges();
   });
-
+// this one test is failing when run in the webstorm test runner IDE
```
bdsl, line 70:
```suggestion

```

src/app/donation-start/donation-start-form/PaymentReadinessTracker.ts
NoelLH: @bdsl I'm not clear what's next here, I think there's a typo in
> I'll probably at least one of them

src/app/donation-start/donation-start-form/donation-start-form.component.html
NoelLH: Very nice! 🙌

NoelLH (dismissed):
Looks good! (Though I think ideally we'll do a Production release this AM before merging this)

src/app/donation-start/donation-start-form/PaymentReadinessTracker.ts
bdsl: should have said "probably remove" at least one of them. But I've already removed `stripePaymentMethodReady` after I wrote that comment.

PR #1334, opened 2023-10-06
DON-906 – fix some Campaign Info progressbars

NoelLH:
For shared fund child campaigns, make the progress % reflect the parent, like the related numbers next to the bar, where applicable.

### DIFF
--- a/src/app/campaign.service.ts
+++ b/src/app/campaign.service.ts
@@ -67,7 +67,22 @@ export class CampaignService {
     return dateToUse;
   }
 
-  static percentRaised(campaign: (Campaign | CampaignSummary)): number | undefined {
+  /**
+   * @param useParentIfApplicable Whether to get a percentage for the parent/meta-campaign if possible. This
+   *                              is possible when `campaign` is a detailed Campaign, e.g. on /campaign/..., and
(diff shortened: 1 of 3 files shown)


src/app/campaign.service.ts
```diff
@@ -67,7 +67,22 @@ export class CampaignService {
...
-  static percentRaised(campaign: (Campaign | CampaignSummary)): number | undefined {
+  /**
+   * @param useParentIfApplicable Whether to get a percentage for the parent/meta-campaign if possible. This
+   *                              is possible when `campaign` is a detailed Campaign, e.g. on /campaign/..., and
+   *                              it takes effect when the `parentUsesSharedFunds`.
+   */
+  static percentRaised(
+    campaign: (Campaign | CampaignSummary),
+    useParentIfApplicable = false,
+  ): number | undefined {
```
bdsl, line 78: I think this would be easier to understand if we have two separate functions instead of one function with the `useParentIfApplicable` flag. The flag value is always known statically so there's no need to test it at runtime.

NoelLH: Makes sense, let's amend in a subsequent PR as the fix is somewhat time sensitive

PR #1339, opened 2023-10-09
DON-819: Show errors blocking progress from payment step

bdsl:
Instead of a greyed out continue button at the payment step, we now have an active button that triggers display of all errors in the section the donor isn't ready to continue:

The errors are shown both in a toast message to get immediate attention, and above the button to allow the donor to review them as they edit the form:

image

### DIFF
--- a/src/app/donation-start/donation-start-form/PaymentReadinessTracker.ts
+++ b/src/app/donation-start/donation-start-form/PaymentReadinessTracker.ts
@@ -72,9 +72,6 @@ export class PaymentReadinessTracker {
     return [...this.humanReadableFormValidationErrors(), ...paymentErrors];
   }
 
-  /**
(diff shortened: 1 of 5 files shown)


src/app/donation-start/donation-start-form/PaymentReadinessTracker.ts
```diff
@@ -26,27 +26,104 @@ export class PaymentReadinessTracker {
...
+          "Please complete your payment method."
+      ];
+    }
+
+    return [...this.humanReadableFormValidationErrors(), ...paymentErrors];
+  }
+
+  /**
+   * todo - translate keys to human readable names, deal with errors other than 'reqauired'
+   */
```
dorota-joanna, line 77:
```suggestion
```

dorota-joanna (dismissed):
LGTM

PR #1344, opened 2023-10-13
DON-909: Show message for unexpected errors blocking stepper.next, in…

bdsl:
…stead of silently not nexting


dorota-joanna (approved):
LGTM

PR #1358, opened 2023-10-23
DON-898: Ask donation funds users for permission instead of assuming …

bdsl:
…we may email them marketing messages

Copy copied from the donation start form

image

### DIFF
--- a/src/app/transfer-funds/transfer-funds.component.html
+++ b/src/app/transfer-funds/transfer-funds.component.html
@@ -157,6 +157,39 @@
                         <button mat-button matStepperNext>Next</button>
                     </div>
                 </mat-step>
+              <mat-step
+                formGroupName="marketing"
+                [stepControl]="marketingGroup"
+              >
+                <ng-template matStepLabel>Receive Updates</ng-template>
+
+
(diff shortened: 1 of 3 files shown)


src/app/transfer-funds/transfer-funds.component.html
```diff
@@ -157,6 +157,39 @@ <h3 class="b-rh-1 b-bold">Transfer Donation Funds</h3>
...
                     </div>
                 </mat-step>
+              <mat-step
+                formGroupName="marketing"
+                [stepControl]="marketingGroup"
+              >
+                <ng-template matStepLabel>Receive Updates</ng-template>
+
+
+                <p class="binary-question-header b-bold b-rt-0" id="optInTbgEmail-label">Would you be happy to receive emails from the <span class="colour-primary">Big Give</span>?</p>
```
bdsl, line 167: We should probably delete the word `the`, both here and on the donation start form where I copied this from.

NoelLH: I agree – happy for it to be in this PR or another.

We might also want to think about removing the emphasis span around the org name here, as this form has only one comms question?

NoelLH (approved):

src/app/transfer-funds/transfer-funds.component.html
bdsl: Done in e4b0e336

PR #1359, opened 2023-10-24
DON-472 help donors landing on old campaign pages find new ones

NoelLH:
* [update the /charity/... campaign list](https://github.com/thebiggive/donate-frontend/commit/1b55c4e95792e7f94a1bcc06fc05c0fb51ceebe4)
    * Use the new design; retire the old campaign-card
    * Sort with newest (last ending) campaign first
* [reinstate the /campaign/... redirect](https://github.com/thebiggive/donate-frontend/commit/372bf0a7ac48acb9ab91e338167c50bdf703b7a2)
    * For definitely-closed campaigns, go back from /donate/... to info.
Also fix an error from undefined `this.donationStartForm` when the
template conditions mean none is rendered and there is no such
ViewChild
* [add an initial closed campaign CTA](https://github.com/thebiggive/donate-frontend/commit/82fb70656f427f0c9cf7c98518235051a1b7fce8)
    * For now, introduces a quick banner on *all* closed
campaigns saying that there *may* be others to explore


src/app/charity/charity.component.html
```diff
@@ -10,10 +10,30 @@ <h1 *ngIf="campaigns[0]" class="b-rh-1 b-bold">{{ campaigns[0].charity.name }} C
...
+            [moreInfoButtonUrl]="'/campaign/' + campaign.id"
+            [organisationName]="campaign.charity.name"
+            primary-figure-label="Match Funds Remaining"
+            [primaryFigureAmount]="campaign.matchFundsRemaining | currency: campaign.currencyCode : 'symbol' : '1.0-0'"
+            [progressBarCounter]="getPercentageRaised(campaign)"
+            secondary-figure-label="Total Funds Raised"
+            [secondaryFigureAmount]="campaign.amountRaised | currency: campaign.currencyCode : 'symbol' : '1.0-0'"
+            [isFutureCampaign]="isInFuture(campaign)"
+            [isPastCampaign]="isInPast(campaign)"
+            [datetime]="getRelevantDateAsStr(campaign)">
```
bdsl, line 34: Is this code just moved from a now deleted component? If so there's no need to review the details - if not @NoelLH can you let me know which parts are different? I'm not sure if there's any easy way to see that.

src/app/charity/charity.component.ts
```diff
@@ -13,17 +15,37 @@ export class CharityComponent implements OnInit {
...
       'Archive of Big Give match funded campaigns',
       null,
     );
   }
+
+  getPercentageRaised(childCampaign: CampaignSummary) {
+    CampaignService.percentRaisedOfIndividualCampaign(childCampaign);
+  }
+
+  getRelevantDateAsStr(campaign: CampaignSummary) {
```
bdsl, line 39: Definitely not a blocker but I think this is just complicated enough that it's worth making the type explicit instead of relying on inference.

```suggestion
  getRelevantDateAsStr(campaign: CampaignSummary): string | null {
```

src/app/charity/charity.component.ts
```diff
@@ -13,17 +15,37 @@ export class CharityComponent implements OnInit {
...
+    return date ? this.datePipe.transform(date, 'dd/MM/yyyy, HH:mm') : null;
+  }
+
+  isInFuture(campaign: CampaignSummary) {
+    return CampaignService.isInFuture(campaign);
+  }
+
+  isInPast(campaign: CampaignSummary) {
+    return CampaignService.isInPast(campaign);
+  }
```
bdsl, line 50: Again not a blocker but it looks like these can be written a bit more simply as:

```suggestion
  isInFuture = CampaignService.isInFuture;

  isInPast = CampaignService.isInPast;
```

bdsl (dismissed):
Assuming code replacing `app-campaign-card` is unedited therefore approving.

src/app/charity/charity.component.html
NoelLH: @bdsl Not exactly, the new version is adapted from the meta-campaign grid – this allows us to remove the old design card from Donate entirely as this was the last page using it. You can see the before and after visually in Jira.

src/app/charity/charity.component.ts
NoelLH: Seems reasonable – this helper was copied from the metacampaign page too, but I'm hesitant to abstract out one liners when there are only 2-3 instances

src/app/charity/charity.component.html
```diff
@@ -10,10 +10,30 @@ <h1 *ngIf="campaigns[0]" class="b-rh-1 b-bold">{{ campaigns[0].charity.name }} C
...
+            [campaignType]="campaign.isMatched ? 'Match Funded' : null"
+            [campaignTitle]="campaign.title"
+            [donateButtonUrl]="'/donate/' + campaign.id"
+            [moreInfoButtonUrl]="'/campaign/' + campaign.id"
+            [organisationName]="campaign.charity.name"
+            primary-figure-label="Match Funds Remaining"
+            [primaryFigureAmount]="campaign.matchFundsRemaining | currency: campaign.currencyCode : 'symbol' : '1.0-0'"
+            [progressBarCounter]="getPercentageRaised(campaign)"
+            secondary-figure-label="Total Funds Raised"
+            [secondaryFigureAmount]="campaign.amountRaised | currency: campaign.currencyCode : 'symbol' : '1.0-0'"
```
bdsl, line 31: I'm not sure what '1.0-0' is. Maybe worth adding a comment to explain and/oror extracting it to give it a name.

NoelLH: Done in https://github.com/thebiggive/donate-frontend/pull/1359/commits/20bb1dee6f9b55f6fb128857c6df46959e2a1dd6 and made everything in currency stats contexts use the new var

PR #1360, opened 2023-10-24
DON-902 – fix submenu not usable

NoelLH:
It was previously being lost under the reserved funds banner on desktop, making all sub-items unreachable


dorota-joanna (approved):
LGTM

PR #1361, opened 2023-10-24
DON-820 – avoid & better report errors paying with donation funds

NoelLH:
* [only offer reuse of donations with the active `paymentMethodType`](https://github.com/thebiggive/donate-frontend/pull/1361/commits/7a902dd55514ea292ac8c397adc12c26f19daa89)
[7a902dd](https://github.com/thebiggive/donate-frontend/pull/1361/commits/7a902dd55514ea292ac8c397adc12c26f19daa89)
    * This should avoid the donate form breaking in edge cases like donors
who started one donation recently, then later topped up with new funds
* [update error copy, reflecting the remaining way we know paying with donor funds can fail](https://github.com/thebiggive/donate-frontend/pull/1361/commits/1008caa2458d688125b634596e708bc08625e8a4)
[1008caa](https://github.com/thebiggive/donate-frontend/pull/1361/commits/1008caa2458d688125b634596e708bc08625e8a4)

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -412,14 +412,14 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
   })
 
   resumeDonationsIfPossible() {
-    this.donationService.getProbablyResumableDonation(this.campaignId)
+    this.donationService.getProbablyResumableDonation(this.campaignId, this.getPaymentMethodType())
       .subscribe((existingDonation: (Donation|undefined)) => {
         this.previousDonation = existingDonation;
 
         // The local check might not have the latest donation status in edge cases, so we need to check the copy
         // the Donations API returned still has a resumable status and wasn't completed or cancelled since being
         // saved locally.
-        if (!existingDonation || !this.donationService.isResumable(existingDonation)) {
+        if (!existingDonation || !this.donationService.isResumable(existingDonation, this.getPaymentMethodType())) {
           // No resumable donations
           return;
         }
@@ -652,7 +652,6 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
       });
     }
   }
-
 
   /**
    * According to stripe docs https://stripe.com/docs/js/element/events/on_change?type=paymentElement the change event has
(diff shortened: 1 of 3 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -775,24 +774,27 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+          const errorMsg = response.error?.error?.description;
+
           this.matomoTracker.trackEvent(
             'donate_error',
             'stripe_customer_balance_payment_error',
-            error.message ?? '[No message]',
+            errorMsg ?? '[No message]',
           );
-          this.stripeError = `Cash balance application failed: ${error.message}`;
+          this.stripeError = 'Your donation has not been processed as it seems you have insufficient funds. Please refresh the page to see your remaining balance.';
```
bdsl, line 795: We're not 100% sure at this point that the two things are related, are we? So maybe better in separate sentences?

```suggestion
          this.stripeError = 'Your donation has not been processed. It seems you have insufficient funds. Please refresh the page to see your remaining balance.';
```

src/app/donation.service.ts
```diff
@@ -71,24 +71,31 @@ export class DonationService {
...
    * Get a recent eligible-for-resuming donation to this same campaign/project, if any exists. We look
    * for candidates based on local status initially but return an Observable resulting from a re-GET
-   * so we include the latest server status. This is why this private method only knows the donation is
+   * so we include the latest server status. This is why this method only knows the donation is
    * 'probably' resumable.
    */
-  getProbablyResumableDonation(projectId: string): Observable<Donation | undefined> {
+  getProbablyResumableDonation(
+    projectId: string,
+    requiredPaymentMethodType: 'card' | 'customer_balance',
```
bdsl, line 90: Definitely not a blocker, but I'm not sure what the word `required` is doing for us here. IMHO `paymentMethodType` would be enough here and in `isResumable`.

bdsl (dismissed):
Added some comments inline, no blockers.

src/app/donation.service.ts
NoelLH: Yep, I suppose by the same logic as `projectId` it's superfluous. Updated.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
NoelLH: I think that we're fairly sure – I don't understand how the new syntax would be more accurate in any scenario we know about, so I'd favour keeping the requested copy from Jira for now.

bdsl (approved):

PR #1362, opened 2023-10-24
DON-596 – transfer funds tip handling

NoelLH:
* ask Identity to return any existing tip balances awaiting bank transfer funding
* if present in GBP, acknowledge any existing tip
* and don't prompt for another, including the Gift Aid form group

### DIFF
--- a/src/app/donation-start/donation-start-container/donation-start-container.component.ts
+++ b/src/app/donation-start/donation-start-container/donation-start-container.component.ts
@@ -87,7 +87,7 @@ export class DonationStartContainerComponent implements AfterViewInit, OnInit{
       return;
     }
 
-    this.identityService.get(id, jwt).subscribe(
+    this.identityService.get(id, jwt, false).subscribe(
       (person: Person) => {
       this.donor = person; // Should mean donations are attached to the Stripe Customer.
       this.personIsLoginReady = true;
--- a/src/app/donation-start/donation-start-container/donation-start-container.component.ts
+++ b/src/app/donation-start/donation-start-container/donation-start-container.component.ts
@@ -87,7 +87,7 @@ export class DonationStartContainerComponent implements AfterViewInit, OnInit{
       return;
     }
 
-    this.identityService.get(id, jwt, false).subscribe(
+    this.identityService.get(id, jwt).subscribe(
       (person: Person) => {
       this.donor = person; // Should mean donations are attached to the Stripe Customer.
       this.personIsLoginReady = true;
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -2093,7 +2093,7 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
 
             if (this.donor?.id) {
               const jwt = this.identityService.getJWT() as string;
-              this.identityService.get(this.donor?.id, jwt, false).subscribe((person: Person) => {
+              this.identityService.get(this.donor?.id, jwt).subscribe((person: Person) => {
                 if (! this.donor?.id) {
                   throw new Error("Person identifier went away");
                 }
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -2093,7 +2093,7 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
 
             if (this.donor?.id) {
               const jwt = this.identityService.getJWT() as string;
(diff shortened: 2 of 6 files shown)


src/app/identity.service.ts
```diff
@@ -81,19 +81,24 @@ export class IdentityService {
       person);
   }
 
-  get(id: string, jwt: string): Observable<Person> {
+  get(id: string, jwt: string, withTipBalances: boolean): Observable<Person> {
```
bdsl, line 84: It's sort of OK in the IDE, but when we read code that calls this in github we'd have to remember what the third param means - if we just see false or true that doesn't tell us much. So could we make it
```suggestion
  get(id: string, jwt: string, {withTipBalances}: {withTipBalances: boolean}): Observable<Person> {
```

The repetition of `withTipBalances` is a bit ugly, but then the calling code gets a lot clearer:
```diff
-    this.identityService.get(id, jwt, false).subscribe(
+    this.identityService.get(id, jwt, {withTipBalances: false}).subscribe(
```

src/app/person.model.ts
```diff
@@ -14,6 +14,12 @@ export interface Person {
      */
     cash_balance?: { [currencyCode: string]: number };
 
+    /**
+     * The total of donor fund form-created pending tips for the given Person, which will be fulfilled
```
bdsl, line 18:
```suggestion
     * The total of donor fund form-created pending tips for the given Person, counted in minor-currency units i.e. pence which will be fulfilled
```

bdsl (commented):
Added inline comment. Also wondering if we should turn off the new section of the form about permission to send them emails when we're not asking for a tip, since it isn't relevant.

src/app/identity.service.ts
bdsl: Or we could make false the default and then this PR doesn't need to change any of the callers except the one that passes true:
```suggestion
  get(id: string, jwt: string, {withTipBalances = false}: {withTipBalances?: boolean} = {}): Observable<Person> {
```

src/app/identity.service.ts
```diff
@@ -102,7 +107,7 @@ export class IdentityService {
       return of(null);
     }
 
-    return this.get(idAndJWT.id, idAndJWT.jwt);
+    return this.get(idAndJWT.id, idAndJWT.jwt, false);
```
NoelLH, line 110:
```suggestion
    return this.get(idAndJWT.id, idAndJWT.jwt);
```

src/app/donation-start/donation-start-container/donation-start-container.component.ts
```diff
@@ -87,7 +87,7 @@ export class DonationStartContainerComponent implements AfterViewInit, OnInit{
       return;
     }
 
-    this.identityService.get(id, jwt).subscribe(
+    this.identityService.get(id, jwt, false).subscribe(
```
NoelLH, line 90:
```suggestion
    this.identityService.get(id, jwt).subscribe(
```

src/app/transfer-funds/transfer-funds.component.ts
```diff
@@ -342,17 +342,24 @@ export class TransferFundsComponent implements AfterContentInit, OnInit {
 
   private loadAuthedPersonInfo(id: string, jwt: string) {
     this.isLoading = true;
-    this.identityService.get(id, jwt).subscribe((person: Person) => {
+    this.identityService.get(id, jwt, true).subscribe((person: Person) => {
```
NoelLH, line 345:
```suggestion
    this.identityService.get(id, jwt, {withTipBalances: true}).subscribe((person: Person) => {
```

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -2093,7 +2093,7 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
 
             if (this.donor?.id) {
               const jwt = this.identityService.getJWT() as string;
-              this.identityService.get(this.donor?.id, jwt).subscribe((person: Person) => {
+              this.identityService.get(this.donor?.id, jwt, false).subscribe((person: Person) => {
```
NoelLH, line 2096:
```suggestion
              this.identityService.get(this.donor?.id, jwt).subscribe((person: Person) => {
```

NoelLH:
> Added inline comment. Also wondering if we should turn off the new section of the form about permission to send them emails when we're not asking for a tip, since it isn't relevant.

Yep good point – that's hopefully done in https://github.com/thebiggive/donate-frontend/pull/1362/commits/ad5e8a325cd9601ae0c0b5cae7236e5ee6641061

bdsl (approved):

PR #1367, opened 2023-10-30
DON-596 - fix logic when a tip is pending

NoelLH:
* set tip numbers to 0 explicitly as an existing tip's detected
* add an extra check to prevent creating another tip if form data somehow gets out of sync

### DIFF
--- a/src/app/transfer-funds/transfer-funds.component.html
+++ b/src/app/transfer-funds/transfer-funds.component.html
@@ -23,7 +23,7 @@
                         Must be between {{ minimumCreditAmount | exactCurrency:currency }} and {{ maximumCreditAmount | exactCurrency:currency }} inclusive.
                     </p>
 
-                    <div *ngIf="pendingTipBalance > 0">
+                    <div *ngIf="donorHasPendingTipBalance">
                         <p>Thanks for tipping Big Give {{ (pendingTipBalance / 100) | exactCurrency:"GBP" }} </p>
                     </div>
 
--- a/src/app/transfer-funds/transfer-funds.component.html
+++ b/src/app/transfer-funds/transfer-funds.component.html
@@ -23,8 +23,8 @@
                         Must be between {{ minimumCreditAmount | exactCurrency:currency }} and {{ maximumCreditAmount | exactCurrency:currency }} inclusive.
                     </p>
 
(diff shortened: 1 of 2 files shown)


src/app/transfer-funds/transfer-funds.component.ts
```diff
@@ -340,6 +341,14 @@ export class TransferFundsComponent implements AfterContentInit, OnInit {
     this.identityService.clearJWT();
   }
 
+  /**
+   * Amount in existing committed tips to be fulfilled, in minor units (i.e. pence),
+   * currently just for GBP / UK bank transfers. 0 if donor's not yet loaded.
+   */
+  get pendingTipBalance() {
```
bdsl, line 348: I feel like probably we should have explicit type here because otherwise you have to look a few references deep to find the type of 'gbp'. Could be persuade otherwise though.
```suggestion
  get pendingTipBalance(): number {
```

src/app/transfer-funds/transfer-funds.component.html
```diff
@@ -23,8 +23,8 @@ <h3 class="b-rh-1 b-bold">Transfer Donation Funds</h3>
                         Must be between {{ minimumCreditAmount | exactCurrency:currency }} and {{ maximumCreditAmount | exactCurrency:currency }} inclusive.
                     </p>
 
-                    <div *ngIf="(donor.pending_tip_balance?.gbp || 0) > 0">
-                        <p>Thanks for tipping Big Give {{ ((donor.pending_tip_balance?.gbp || 0) / 100) | exactCurrency:"GBP" }} </p>
+                    <div *ngIf="pendingTipBalance > 0">
```
bdsl, line 26: this expression `pendingTipBalance > 0` is used three times - probably worth making it into a function like `donorHasPendingTip: boolean` or something?

NoelLH: Done in https://github.com/thebiggive/donate-frontend/pull/1367/commits/5e0cb12d96f390e4dbd7cccaf576d07a84fd7b6a

bdsl (approved):
Not 100% sure where the bug was in the old code, but new version looks good so no harm in merging and re-testing.

PR #1371, opened 2023-11-01
DON-876: Donation start page a18y improvements

bdsl:
Removing the empty panel doesn't change the appearance of the page.

Not sure why this panel was there, but it's empty and has no code comment to explain its purpose and is confusing when using keyboard navigation and screen reader, so removing

Adjusted confirmation text at the end of the page looks like this:

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -152,10 +152,6 @@
               Back to percentage tip
             </span>
 
-          <mat-expansion-panel hideToggle>
-            <mat-expansion-panel-header>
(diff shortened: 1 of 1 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -560,16 +556,16 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
         label="Confirm"
         class="c-make-your-donation"
       >
+<!--        here -->
```
bdsl, line 559:
```suggestion
```

dorota-joanna (approved):
LGTM

PR #1372, opened 2023-11-02
DON-877: Save my details dialog screen reader adjustments:

bdsl:
- Stop auto-focusing on the 'cancel' button so it doesn't have to be read first
- Set entire dialog content as 'aria-live' so it is automatically read out when the dialog is opened
- Change link copy from "Learn more" to "Learn more about stripe"


dorota-joanna (approved):
LGTM

PR #1373, opened 2023-11-02
DON-904 - tweak new My Account timeout

NoelLH:
Only allow setting it client-side, and clear
it when appropriate.

Also fix a typo in a similar bit of metacampaign
cleanup

### DIFF
--- a/src/app/meta-campaign/meta-campaign.component.ts
+++ b/src/app/meta-campaign/meta-campaign.component.ts
@@ -119,7 +119,7 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
 
   ngOnDestroy() {
     if (isPlatformBrowser(this.platformId) && this.tickerUpdateTimer) {
-      clearInterval(this.tickerUpdateTimer);
+      clearTimeout(this.tickerUpdateTimer);
     }
 
     if (this.routeChangeListener) {
--- a/src/app/my-account/my-account.component.ts
+++ b/src/app/my-account/my-account.component.ts
@@ -25,7 +25,7 @@ export class MyAccountComponent implements OnDestroy, OnInit {
   registerSucessMessage: string | undefined;
   protected readonly faExclamationTriangle = faExclamationTriangle;
 
-  private savedCardsTimer: number;
+  private savedCardsTimer: ReturnType<typeof setTimeout>; // https://stackoverflow.com/a/56239226
 
   constructor(
     private pageMeta: PageMetaService,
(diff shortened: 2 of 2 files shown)


src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -119,7 +119,7 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
 
   ngOnDestroy() {
     if (isPlatformBrowser(this.platformId) && this.tickerUpdateTimer) {
-      clearInterval(this.tickerUpdateTimer);
+      clearTimeout(this.tickerUpdateTimer);
```
bdsl, line 122: Looks good, I see this was set by setTimeout not setInterval so I guess clearInterval was a mistake before.

src/app/my-account/my-account.component.ts
```diff
@@ -17,19 +17,22 @@ import {HttpErrorResponse} from "@angular/common/http";
...
-export class MyAccountComponent implements OnInit {
+export class MyAccountComponent implements OnDestroy, OnInit {
   public person: Person;
 
   public paymentMethods: PaymentMethod[]|undefined = undefined;
   registerErrorDescription: string | undefined;
   registerSucessMessage: string | undefined;
   protected readonly faExclamationTriangle = faExclamationTriangle;
 
+  private savedCardsTimer: ReturnType<typeof setTimeout>; // https://stackoverflow.com/a/56239226
```
bdsl, line 28: Annoying that the type is different in Node and browsers, but this looks like a good solution. And comment is definitely needed - thanks.

src/app/my-account/my-account.component.ts
```diff
@@ -51,7 +54,15 @@ export class MyAccountComponent implements OnInit {
       }
     });
 
-    setTimeout(this.checkForPaymentCardsIfNotLoaded, 5_000);
+    if (isPlatformBrowser(this.platformId)) {
```
bdsl, line 57: not a blocker, but might be worth moving this inside the `else` block above, so it doesn't run unless there is a logged in person and after we know who they are.

src/app/my-account/my-account.component.ts
bdsl, line 28:
```suggestion
  private savedCardsTimer: undefined | ReturnType<typeof setTimeout>; // https://stackoverflow.com/a/56239226
```

Since we only conditionally set this in OnInit maybe worth making undefined part of the type, or initialising to null here.

bdsl (approved):

PR #1376, opened 2023-11-07
Don 918 update homepage

bdsl:
To test, with a simulated current date, visit a URL such as http://localhost:4200/?simulatedDate=2023-11-28T12:00:00%2B00:00

The number at the end after `%2B` is the timezone offset from UTC - so keep it at zero if you just want to test for London in winter.

image

image

image

image

### DIFF
--- a/src/app/home/home.component.ts
+++ b/src/app/home/home.component.ts
@@ -84,6 +84,15 @@ export class HomeComponent implements OnInit {
       'https://images-production.thebiggive.org.uk/0011r00002IMRknAAH/CCampaign%20Banner/db3faeb1-d20d-4747-bb80-1ae9286336a3.jpg',
     );
     this.stats = this.route.snapshot.data.stats;
+
+
+    if (environment.environmentId !== 'production') {
+      const queryParams = this.route.snapshot.queryParams;
+
+      if (queryParams.simulatedDate) {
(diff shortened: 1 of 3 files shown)


src/app/home/home.component.ts
```diff
@@ -38,19 +21,60 @@ export class HomeComponent implements OnInit {
...
+        text: 'See Results',
+        href: new URL('/christmas-challenge-2023', environment.donateGlobalUriPrefix)
       }
     },
-    globalCard,
     {
+      appearAt: 'asap',
+      disappearAt: GMFApplicationCloseDate,
+      headerText: "Applications for Green Match Fund are now open!",
+      bodyText: "Apply by January 16 2024",
```
NoelLH, line 67: Should we standardise on a date format for these? Some copy has the number with suffix followed by the month

NoelLH (approved):
Added a small question about dates in copy, but the logic all looks good to me

src/app/home/home.component.ts
bdsl: Yes we probably should standardise, either to have the suffix or not have it. I don't have a strong feeling which way, I think I'd rather we leave it out everywhere. @BethBigGive what do you think?

bdsl: Merged so will move question to ticket.

PR #1382, opened 2023-11-20
DON-925 Add redirect in browser from hompeage to CC23 page during hours of CC (+ 12 hours in advance)

bdsl:
The redirect can be tested in any non-prod envirionment by passing a simulated date in a query param, e.g

http://localhost:4200?simulatedDate=2023-11-28T00%3A00%3A00%2B00%3A00

If the noredirect param is also passed then the redirect will be cancelled, as if the user had clicked the logo link as generated by components during the CC period:

http://localhost:4200/?simulatedDate=2023-11-28T00:00:00%2B00:00&noredirect


src/app/home/home.component.ts
```diff
@@ -125,6 +127,24 @@ export class HomeComponent implements OnInit {
...
+    // start the redirect 12 hours in advance of CC open:
+    const startRedirectingToCCAt = new Date('2023-10-28T00:00:00+00:00');
+
+    // end the redirect exactly at the time CC closes.
+    if (
+      !queryParams.hasOwnProperty('noredirect') &&
+      this.currentTime >= startRedirectingToCCAt &&
+      this.currentTime < CCCloseDate &&
+      isPlatformBrowser(this.platformId)) {
+      this.router.navigate(
```
bdsl, line 140: we planned to do a 302 rather than a 301 redirect, but as this is a purely in-browser redirect the distinction isn't relevant. I'm fairly sure that browsers will treat this redirect as temporary.

NoelLH: Right – it's not an HTTP redirect at all so browsers will add it to location history and would never repeat it without running the same JS again.

src/app/home/home.component.ts
```diff
@@ -125,6 +127,24 @@ export class HomeComponent implements OnInit {
...
+        ['/christmas-challenge-2023'],
+        {
+          replaceUrl: true, // As we are redirecting imediatly it would be confusing to leave a page the user hasn't seen in their history.
```
NoelLH, line 143:
```suggestion
          replaceUrl: true, // As we are redirecting immediately it would be confusing to leave a page the user hasn't seen in their history.
```

NoelLH (approved):

PR #1385, opened 2023-11-21
DON-929 – improve menu + login state handling

NoelLH:
Hopefully also fixes JWT removal on logout not consistently taking effect.

Also reduces some extraneous Identity API calls and probably some unnecessary Subscriptions.


dorota-joanna (approved):
LGTM, tested locally and got logged out successfully on donate page and transfer-funds page redirects to home after logout

PR #1387, opened 2023-11-21
DON-924: Fix missing indication of active filters on explore & metaca…

bdsl:
…mpaign pages

I think the filters now work right with all actions including adding and removing filters and hard-refreshing the page, or equivalently copying the address from the address bar and messaging it to another person. But it might be worth another dev having a look before we put it ready for QA.

Screenshot doesn't show anything different to what we have now, but the big difference is that refreshing the page does not change the appearance, whereas in prod it makes the filter removal buttons disappear.

image


NoelLH (approved):
Tested quite a few permutations on local in Firefox (just in dev mode, not SSR). Can't find any issues running that way 🔥

PR #1388, opened 2023-11-21
DON-923 – prepare for CC23 A/B test

NoelLH:


src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -65,11 +65,10 @@
...
             <span *ngIf="!alternateCopy">
-              Big Give doesn't charge platform fees. That means we're only able to continue offering our services thanks to
-              the generosity of donors who leave a tip.
+              Big Give is a registered charity (1136547), and since 2008, thanks to the generosity of people like you who leave a tip, we can continue providing a match funding platform for the charities and causes you care about.
             </span>
             <span *ngIf="alternateCopy">
-              Big Give is a registered charity (1136547), and since 2008, our aim has been to help thousands of charities double the difference of their donations. Thanks to the generosity of people like you who leave a tip, we can continue providing a match funding platform for the charities and causes you care about.
+              Big Give is a registered charity (1136547) and since 2008 has been helping thousands of charities double their donations. Tips left by generous people like you go into our Anchor Match Fund to provide further match funding for our charity partners as well as helping us deliver our charitable mission. Doubling the difference.
```
bdsl, line 71: slight nitpick, but should we match the two sides of the 'as well as' more closely in verb form? We've got the bare verb 'go' on one side and the gerund/participle 'helping' on the other.

bdsl (approved):
LGTM. I'm just wondering if we should hold it back from `develop` until after we've got the next deploy to prod done, so that QA for this doesn't hold up the deploy which will already be a bigger batch of work than we would like to release all at once.

NoelLH:
> LGTM. I'm just wondering if we should hold it back from `develop` until after we've got the next deploy to prod done, so that QA for this doesn't hold up the deploy which will already be a bigger batch of work than we would like to release all at once.

Yep, I think that makes sense. It might also happen naturally before we get the feedback to finalise our test plan.

PR #1389, opened 2023-11-22
DON-925: Hide content of home page when user is about to be redirecte…

bdsl:
…d away from it.

Showing content before a redirect is unhelpful.

It might be better to do this in the app-routing.ts file, but I'm not sure we have easy access to queryParams there, I didn't see a way to access them.


dorota-joanna (commented):
LGTM

PR #1393, opened 2023-11-24
DON-925 – redirect to CC23 on the server side, too

NoelLH:


src/app/home/home.component.ts
```diff
@@ -110,6 +112,7 @@ export class HomeComponent implements OnInit {
     private route: ActivatedRoute,
     private router: Router,
     @Inject(PLATFORM_ID) private platformId: Object,
+    @Optional() @Inject(RESPONSE) private response: any,
```
bdsl, line 115:
```suggestion
    @Optional() @Inject(RESPONSE) private response: Response,
```

Requires importing Response, I think this is the right one:
```
import { Response } from 'express';
```

bdsl (approved):
I'd replace the `any` type, although not so important since this is temp code. Otherwise LGTM, looking forward to seeing how it works in staging.

bdsl (approved):

PR #1394, opened 2023-11-24
BG2-2521: Generate alarms on deployment failure (plus for now on succ…

bdsl:
…essful deployment to staging to test if the alarm works)


.circleci/config.yml
```diff
@@ -240,8 +240,10 @@ workflows:
           container-image-name-updates: 'container=donate,tag=regression-${CIRCLE_SHA1}'
           post-steps:
             - slack/notify:
+                channel: deployments, regression-alarms
                 event: fail
                 template: basic_fail_1
+                mentions: @here
```
bdsl, line 246: ah actually thought I'd deleted this line. But maybe worth leaving it in temporarily to see what it does.

.circleci/config.yml
```diff
@@ -293,8 +296,11 @@ workflows:
                     ]
                   }
             - slack/notify:
+                # Temporarily sending alarms to test alarm sending setup.
```
NoelLH, line 299: Isn't it the build that has failed semi-regularly, rather than the deploy / service update?

bdsl: Ah yes you're right. But we should alarm if either fail - and merging this will be enough to tell us whether generating an alarm like this works or not.

bdsl: Added notification for the build in https://github.com/thebiggive/donate-frontend/pull/1394/commits/5332e0b539fdb1b9572a0c8ad7d8201f3d456d69 . I don't know why Circle doesn't seem to have an option to add a post-step for the workflow as a whole, rather than just each individual job.

.circleci/config.yml
```diff
@@ -314,6 +340,16 @@ workflows:
             - ecs-deploys
           requires:
             - deploy-production-static
+          post-steps:
+            - jira/notify:
```
NoelLH, line 344: Do we need an extra Jira notify here? Why's it `staging`?

.circleci/config.yml
```diff
@@ -314,6 +340,16 @@ workflows:
...
+                job_type: deployment
+                environment_type: staging
```
bdsl, line 346: No we don't. Sorry that must have been bad copy and pasting.

```suggestion
```

.circleci/config.yml
bdsl: @NoelLH no we don't, sorry I must have added that by mistake. Deleted.

dorota-joanna (approved):
LGTM

NoelLH (approved):
There are still 2 Jira additions which I assume we don't need, but I think they're in harmless and temporary places & don't have incorrect `environment_type`s

bdsl:
> There are still 2 Jira additions which I assume we don't need, but I think they're in harmless and temporary places & don't have incorrect `environment_type`s

Right, I'll clear those up in the next PR along with taking out the test alarms.

PR #1401, opened 2023-11-27
DON-752: Create login page

bdsl:
image

Login page is at `/login`

Includes function to redirect the user to any other page on the site if `?r=<redirectTarget>` is added to the URL. My plan is to use a router guard to redirect attempts to access pages like `my-account` to `/login?r=my-account`

### DIFF
--- a/src/app/app-routing.ts
+++ b/src/app/app-routing.ts
@@ -1,4 +1,4 @@
-import { Routes } from '@angular/router';
+import {Router, Routes} from '@angular/router';
 
 import { CampaignListResolver } from './campaign-list.resolver';
 import { CampaignResolver } from './campaign.resolver';
@@ -6,6 +6,20 @@ import { CharityCampaignsResolver } from './charity-campaigns.resolver';
 import {campaignStatsResolver} from "./campaign-stats-resolver";
 import {LoginComponent} from "./login/login.component";
 import {environment} from "../environments/environment";
+import {inject} from "@angular/core";
+import {IdentityService} from "./identity.service";
+
+const redirectToHomeIfLoggedIn = () => {
+  const identityService = inject(IdentityService);
+  const router = inject(Router);
+
+  const isLoggedIn = identityService.probablyHaveLoggedInPerson();
+  if (! isLoggedIn) {
+    return true;
+  } else {
+    return router.parseUrl('/');
+  }
+};
 
 const routes: Routes = [
   {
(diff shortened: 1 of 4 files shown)


src/app/app-routing.ts
```diff
@@ -1,9 +1,25 @@
...
+
+const redirectToHomeIfLoggedIn = () => {
+  const identityService = inject(IdentityService);
+  const router = inject(Router);
+
+  const isLoggedIn = identityService.probablyHaveLoggedInPerson();
+  if (! isLoggedIn) {
+    return true;
+  } else {
+    return router.parseUrl('/');
```
bdsl, line 20: todo - inject the Activated route and redirect the same way that the login component does on successful login, i.e. not always to the home page.

src/app/app-routing.ts
```diff
@@ -122,6 +138,12 @@ const routes: Routes = [
     loadChildren: () => import('./my-account/my-account.module')
       .then(c => c.MyAccountModule),
   },
+  {
+    path: 'my-account',
+    pathMatch: 'full',
+    loadChildren: () => import('./my-account/my-account.module')
+      .then(c => c.MyAccountModule),
+  },
```
bdsl, line 146:
```suggestion
```

src/app/login/login.component.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  ngOnInit() {
+    this.loginForm = this.formBuilder.group({
+      emailAddress: [null, [
+        Validators.required,
+        Validators.pattern(EMAIL_REGEXP),
+      ]],
+      password: [null, [
+        Validators.required,
+        Validators.minLength(10),
```
bdsl, line 52: not sure it's worth having a minLength validation on the password field here. It's here because I copied this from the login modal. Min length is good when setting a new password, but I think it might be more confusing than helpful when trying to authenticate with an existing password.

The only password validation rule as far as the user is concerned here is that the password has to match what's set on the account. Yes we know that that implies it will be at least 10 characters, but that feels a bit arbitrary to enforce here.

dorota-joanna: I think you're right, I don't remember being notified about password length on any login page, but maybe it's good to check length at the form level before checking against db, as we know it can't be right and won't make unnecessary db calls?

dorota-joanna (approved):
LGTM

src/app/login/login.component.ts
bdsl: I think if we do that for performance reasons then we don't need to tell the user about it, as it just makes things more complicated for them - and if we do it it makes more sense to do it in the Identity server instead of in frontend. I'll leave it here for now but likely change in a later PR.

PR #1402, opened 2023-11-28
DON-933: Fix logout not working (?)

bdsl:
…ting

Perhaps navigation is interferring with cookie deletion somehow, although the delete function isn't awaitable.


bdsl:
Testing this on my local it doesn't seem to help, logout still fails some of the time.

NoelLH (approved):

bdsl:
`npm run ci` returned 0 on my local, so will merge this despite CircleCI having an incident and not doing a build. We might still have to wait for Circle before we can deploy.

PR #1404, opened 2023-11-29
DON-936 – tweak main amount input label

NoelLH:
Improve the situation on small mobile breakpoints for charity names which are long-ish but don't exceed
20 characters


NoelLH:
This is the new behaviour at breakpoints under 450px width:

Screenshot 2023-11-29 at 11 42 10

dorota-joanna (approved):
LGTM

PR #1405, opened 2023-11-29
DON-937 – captcha fix & edge case recovery

NoelLH:


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1545,7 +1548,18 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
       return false;
     }
 
-    this.idCaptcha.reset();
+    try {
+      this.idCaptcha.reset();
+    } catch (e) {
```
bdsl, line 1553: this is implicitly `any`, but I guess in practice we can be reasonably sure it will be an instance of Error. Catch has to be either `any` or `unknown`, the latter makes things potentially safer but more verbose. Not a blocker.

bdsl (approved):
LGTM

PR #1406, opened 2023-11-29
DON-938 – fire captcha again when we can detect that it was bypassed

NoelLH:
Bypass can happen by pressing escape or clicking outside a puzzle. The best way to test this is by [setting a bot-like User Agent](https://stackoverflow.com/a/50955485/2803757) or using TestingBot to get more puzzles.

Should avoid situations where the donor gets to 'Payment details' but the Stripe.js element has not been set up.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -520,7 +520,6 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
         if (!this.idCaptchaCode) {
           clickEvent.preventDefault();
 
-          this.stepChangeBlockedByCaptcha = true;
           try {
             this.idCaptcha.reset();
           } catch (e) {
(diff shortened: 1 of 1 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -507,6 +507,32 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+          this.stepChangeBlockedByCaptcha = true;
+          try {
+            this.idCaptcha.reset();
+          } catch (e) {
+            this.matomoTracker.trackEvent('identity_error', 'step_change_captcha_reset_failed', e.message);
+            this.reset(); // Includes page refresh atm – annoying but less so than filling out even more broken fields.
+            return;
+          }
+
+          this.idCaptcha.execute();
```
bdsl, line 532: Setting `stepChangeBlockedByCaptcha` will make the stepper jump to the next step on captcha return. But we don't want the next step here, do we? We want whichever step they clicked on. Do we need to store that step id somewhere to use it in the captchaReturn?

dorota-joanna (approved):
LGTM

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: Happy to look at the docker stuff tomorrow with you if you like. Not sure what the solution is.

PR #1408, opened 2023-11-30
DON-938 – UX & further edge case improvements

NoelLH:
* if donor has managed to get to the Payment details step, try to jump to the previous step
* if that happens, show a toast explaining why
* if they are somehow on Payment Details without a captcha code, also explain why the fields are unavailable next to where they should load on the page

Not tested end to end yet, but screenshots from local dev wip:

Screenshot 2023-11-30 at 11 17 47

Screenshot 2023-11-30 at 11 17 39


dorota-joanna (commented):
LGTM

PR #1410, opened 2023-11-30
DON-941 – surface suggestion to call bank on `invalid_amount` decline

NoelLH:


bdsl (dismissed):
Looks good, I'm not sure we need to include the check for > 500 though - the `invalid_amount` code is enough to tell us it's something to do with the amount, although we don't know that it's a large donation, it might just be more than the donor has available.

Maybe if we lose the >500 check the copy could be something like "The payment was declined. You might need to contact your bank before making a donation of this amount." that doesn't call it a large donation?

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1414,6 +1414,12 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
       customMessage = true;
     }
 
+    if (error.code === 'card_declined' && error.decline_code === 'invalid_amount' && this.donation && this.donation.donationAmount > 500) {
```
NoelLH, line 1417:
```suggestion
    if (error.code === 'card_declined' && error.decline_code === 'invalid_amount') {
```

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1414,6 +1414,12 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+      // We've seen e.g. HSBC in Nov '23 decline large donations with this code.
+      friendlyError = 'The payment was declined. You might need to contact your bank before making a large donation.';
```
NoelLH, line 1419:
```suggestion
      friendlyError = 'The payment was declined. You might need to contact your bank before making a donation of this amount.';
```

PR #1411, opened 2023-11-30
DON-940 – supply extra Stripe `payment_method.billing_details`

NoelLH:
With Payment Stripe.js element. Should help with donor query support and maybe with fraud signals.


src/app/donation.service.ts
```diff
@@ -337,9 +337,12 @@ export class DonationService {
...
         billing_details: {
           address: {
-          country: updatedMethodDetails.countryCode,
-          postal_code: updatedMethodDetails.postalCode
-        }}
+            country: updatedMethodDetails.countryCode,
+            postal_code: updatedMethodDetails.postalCode
+          },
+          email: person.email_address,
+          name: `${person.first_name} ${person.last_name}`,
```
bdsl, line 344: Looks good - I see whatever we send in billing_details get passed through matchbot as-is and then sent on to stripe https://github.com/thebiggive/matchbot/blob/57087038c93ae010b38d82f94c7cb201771e82e9/src/Application/Actions/UpdatePaymentMethod.php#L74 https://stripe.com/docs/api/payment_methods/update

But I see now that updatePaymentMethod is only called from the My Account page. It isn't called from the donation page.

bdsl: So I don't think this PR is changing what you want to change.

bdsl (approved):

PR #1414, opened 2023-12-01
DON-937 – make enter key call `next()` & proceed 1 step when section valid

NoelLH:
Instead of trying to submit the whole form, which will almost always fail.

I suggest reviewing with whitespace diff off – I fixed some confusing indents where I was reading.


bdsl (approved):
I haven't tested, but code looks good.

PR #1415, opened 2023-12-01
DON-938 – rationalise various captcha + step management

NoelLH:
* fix the condition for showing last-resort 'no captcha' error message in 'Payment details' step – our goal is still to make it impossible to see this through earlier step change checks & restrictions
* fix attempts to create undefined amount donations if amount field's empty
* jump back to 'Your donation' step only when we need to collect a [new] amount
* fix when we show a "toast" when dismissed captchas mean donor has got to a point
that can't work;
and jump back a step then too to avoid showing unusable fields
* invoke the captcha from the same common place in code, with the same associated
state changes, every time
* pick up on stepper header clicks more reliably
* check for `this.donor` when assessing captcha readiness, since if we have a valid
donor already we don't need to issue new captcha puzzles

### DIFF
--- a/src/app/donation-complete/donation-complete.component.ts
+++ b/src/app/donation-complete/donation-complete.component.ts
@@ -36,7 +36,6 @@ export class DonationCompleteComponent implements OnInit {
   loggedIn = false;
   minPasswordLength: number;
   noAccess = false;
-  offerToSetPassword = false;
   encodedPrefilledText: string;
   recaptchaIdSiteKey = environment.recaptchaIdentitySiteKey;
   registerError?: string;
@@ -206,7 +205,6 @@ export class DonationCompleteComponent implements OnInit {
               next: person => {
                 this.patchedCorePersonInfo = true;
                 this.person = person;
-                this.offerToSetPassword = !person.has_password;
               },
               error: (error: HttpErrorResponse) => {
                 // For now we probably don't really need to inform donors if we didn't patch their Person data, and just won't ask them to
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -982,7 +982,7 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     }
 
     this.idCaptchaCode = captchaResponse;
-    if (!this.donation) {
+    if (!this.donation && this.donationAmount > 0) {
       this.createDonationAndMaybePerson();
     }
   }
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -709,6 +709,7 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     // We need to check the current index in the stepper because we've seen this fire as soon
     // as the step *before* 'Payment details' loads and initialises the Stripe Payment element.
     if (!this.donation || !this.stripePaymentMethodReady || !this.stripePaymentElement || !this.stripeElements) {
+      // Gift Aid step is only shown (at step index 1) when campaign is in GBP.
       const paymentStepIndex = this.donation?.currencyCode === 'GBP' ? 2 : 1;
       if (this.stepper.selectedIndex > paymentStepIndex) {
         this.jumpToStep('Payment details');
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -268,7 +268,7 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
 
   ngOnDestroy() {
     if (this.donation) {
-      this.clearDonation(this.donation, false);
+      this.clearDonation(this.donation, false, false);
     }
 
     this.destroyStripeElements();
@@ -433,7 +433,8 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
             );
 
             if (this.donation) {
-              this.clearDonation(this.donation, true);
+              // We already know the requested amount, so no need to jump back.
+              this.clearDonation(this.donation, true, false);
             }
             this.createDonationAndMaybePerson();
           });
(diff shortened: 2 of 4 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1266,6 +1305,20 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     }
   }
 
+  private markYourDonationStepIncomplete() {
+    const step = this.stepper.steps.get(0);
+    if (step) {
```
bdsl, line 1310: Step should always be truth shouldn't it? Can we throw an error if it isn't?

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -677,8 +706,13 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     // Jump back if we get an out of band message back that the card is *not* valid/ready.
     // Don't jump forward when the card *is* valid, as the donor might have been
     // intending to edit something else in the `payment` step; let them click Next.
+    // We need to check the current index in the stepper because we've seen this fire as soon
+    // as the step *before* 'Payment details' loads and initialises the Stripe Payment element.
     if (!this.donation || !this.stripePaymentMethodReady || !this.stripePaymentElement || !this.stripeElements) {
-      this.jumpToStep('Payment details');
+      const paymentStepIndex = this.donation?.currencyCode === 'GBP' ? 2 : 1;
```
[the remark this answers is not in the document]
NoelLH: edb718801bb4038bdac1d29a5641b04f1df952a0

src/app/donation-start/donation-start-form/donation-start-form.component.ts
NoelLH: It should, yes. Do you think this works? https://github.com/thebiggive/donate-frontend/pull/1415/commits/3b1fce2667541d52a73186ae42d451080292454b

src/app/donation-start/donation-start-form/donation-start-form.component.ts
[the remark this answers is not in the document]
bdsl: Thanks, that makes sense.

src/app/donation-complete/donation-complete.component.ts
```diff
@@ -36,7 +36,6 @@ export class DonationCompleteComponent implements OnInit {
   loggedIn = false;
   minPasswordLength: number;
   noAccess = false;
-  offerToSetPassword = false;
```
bdsl, line 39: I see this is only written never read on develop, makes sense to delete.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1740,17 +1811,22 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+  /**
+   * Also resets captcha & relevant donation persistence state mgmt,
+   * and returns to step 1 so required input can be collected again.
+   * We DON'T reset `this.donor`, so there should be no need for a new captcha code.
+   *
+   * @param jumpToStart If the caller is already setting up a known-value donation alongside the clear,
+   *                        this should be false. In other cases we need to know the new amount so it
+   *                        should usually be true, if the page is not being unloaded.
+   */
+  private clearDonation(donation: Donation, clearAllRecord: boolean, jumpToStart: boolean) {
```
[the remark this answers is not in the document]
NoelLH: Good shout, done in 0a4814e282034fc596abe9d59db2919f51c118db

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -571,6 +579,22 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
   }
 
   async stepChanged(event: StepperSelectionEvent) {
+    if (event.selectedIndex > 0 && !this.donor) {
+      if (event.selectedIndex > 1) {
```
bdsl, line 583: I was trying to understand why selected index 1 is treated as a special case, but I didn't quite get there. Could we add a comment?

Also if 1 does need to be a special case it might make the code clear to deal with the case of `event.selectedIndex === 1` first.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1740,17 +1816,22 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+   * Also resets captcha & relevant donation persistence state mgmt,
+   * and returns to step 1 so required input can be collected again.
+   * We DON'T reset `this.donor`, so there should be no need for a new captcha code.
+   *
+   * @param jumpToStart If the caller is already setting up a known-value donation alongside the clear,
+   *                        this should be false. In other cases we need to know the new amount so it
+   *                        should usually be true, if the page is not being unloaded.
+   */
+  private clearDonation(donation: Donation, {clearAllRecord, jumpToStart}: {clearAllRecord: boolean; jumpToStart: boolean}) {
     if (clearAllRecord) { // i.e. don't keep donation around for /thanks/... or reuse.
```
bdsl, line 1829:
```suggestion
   * @param clearAllRecord Don't keep donation around for /thanks/... or reuse.
   * @param jumpToStart If the caller is already setting up a known-value donation alongside the clear,
   *                        this should be false. In other cases we need to know the new amount so it
   *                        should usually be true, if the page is not being unloaded.
   */
  private clearDonation(donation: Donation, clearAllRecord: boolean, jumpToStart: boolean) {
    if (clearAllRecord) {
```

NoelLH: I think this suggestion is reversing the destructure bit. Don't think I can partially apply it so will try to mimic locally

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -571,6 +579,22 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
   }
 
   async stepChanged(event: StepperSelectionEvent) {
+    if (event.selectedIndex > 0 && !this.donor) {
```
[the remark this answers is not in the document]
NoelLH: Yep I think so!

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: ah yeah sorry that was a mistake, I just meant to move the comment about the meaning of `clearAllRecord` into the docblock

PR #1417, opened 2023-12-01
DON-937 – update invalid form messaging

NoelLH:
Give it a more explicable donor toast & log event, now we know the remaining reason that former "code B1" can happen.


bdsl (approved):
LGTM

PR #1419, opened 2023-12-01
DON-937 – submit donation iff the final button's clicked

NoelLH:


src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -625,8 +624,10 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
             </tr>
           </table>
 
-          <button *ngIf="!submitting"
-            type="submit"
+          <button
+            *ngIf="!submitting"
+            (click)="submit()"
+            (keyup.enter)="submit()"
```
NoelLH, line 630: In macOS Firefox, without the key binding I was able to select the button like others and saw it in a visually focused state, but had to press space to activate it. I think supporting enter too is possibly helpful for occasional keyboard focus users.

dorota-joanna (approved):
LGTM

PR #1420, opened 2023-12-01
MAT-339: Display generic error message in case of 500 from matchbot o…

bdsl:
…n donation confirm

Currently we show an error message if Stripe throws a CardError, but if it returns some other exception we don't display any error.

I'm not 100% sure if we should encorage the donor to retry at this point, just in case there's some unexpected error that happens after the donation is confirmed and they end up double donating. Stripe won't allow double donating from one payment intent, but potentially the donor could go back and make a new intent.

Need to think about the wording of the message.


NoelLH (dismissed):

dorota-joanna (approved):
LGTM

PR #1422, opened 2023-12-03
DON-943 – refresh balance on My Account

NoelLH:
After donating with customer_balance in the same app load. For now, add calls only very conservatively when we know the balance probably needs an update, to avoid extra load on Identity that hasn't been load tested.

### DIFF
--- a/src/app/identity.service.ts
+++ b/src/app/identity.service.ts
@@ -30,6 +30,8 @@ export class IdentityService {
   // Tracks and changes login status; shared between e.g. outer app menu and specific pages.
   loginStatusChanged = new EventEmitter<boolean>();
 
+  cashBalanceStale = false;
+
   constructor(
     private http: HttpClient,
     private matomoTracker: MatomoTracker,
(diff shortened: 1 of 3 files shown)


src/app/identity.service.ts
```diff
@@ -83,6 +85,8 @@ export class IdentityService {
   }
 
   get(id: string, jwt: string, {withTipBalances = false}: {withTipBalances?: boolean} = {}): Observable<Person> {
+    this.cashBalanceStale = false;
+
```
bdsl, line 89: should we not wait until the person is returned from the server and then set the stale flag to false? I haven't tested but I think that would be:

```typescript
  get(id: string, jwt: string, {withTipBalances = false}: {withTipBalances?: boolean} = {}): Observable<Person> {
     return this.http.get<Person>(
      `${environment.identityApiPrefix}${this.peoplePath}/${id}`,
      {
        headers: new HttpHeaders({ 'X-Tbg-Auth': jwt }),
        params: {
          withTipBalances: withTipBalances ? 'true' : 'false',
        },
      },
    ).pipe(
      tap(() => {this.cashBalanceStale = false})
     );
  }
```

PR #1423, opened 2023-12-04
DON-934, DON-935 – update components

NoelLH:
To 202312041141.0.0, to get totalizer / ticker enhancements


NoelLH:
Seeing some problems on mobile actually, closing for now

PR #1429, opened 2023-12-05
Show debug info on donation form on dev environment only

bdsl:
This isn't especially pretty, but may be useful for quite a lot of tasks developing the donation form. If it gets in the way feel free to rip it out.

image


dorota-joanna (approved):
LGTM

PR #1431, opened 2023-12-06
DON-752: Login page styling

bdsl:
Copying styling and layout from "My Account" and the donation start form.

Quite a bit of copy and paste here, not sure yet if it's time to refactor to remove duplication now or if it's better to wait for one more copy and then refactor.

At the moment this looks like so:

image

edit - after some more work the page looks like this:

image


dorota-joanna (approved):
LGTM

PR #1434, opened 2023-12-08
DON-752: Improve login page error display

bdsl:
image

image


bdsl:
Thanks for the verbal approval @dorota-joanna !

PR #1436, opened 2023-12-12
DON-752: Refactor: try to privatise IdentityService.saveJWT

bdsl:
I tried to make saveJWT a private function but that led to test errors - not sure exactly what was going wrong in that commit. See https://app.circleci.com/pipelines/github/thebiggive/donate-frontend/4857/workflows/cd11916a-8540-4abd-bf3f-d47932774f73/jobs/13879

But this goes part of the way, adding the call to saveJWT inside IdentityService.login so that callers to that don't need to also call saveJWT.

### DIFF
--- a/src/app/identity.service.ts
+++ b/src/app/identity.service.ts
@@ -5,7 +5,7 @@ import {CookieService} from 'ngx-cookie-service';
 import {MatomoTracker} from 'ngx-matomo';
 import {StorageService} from 'ngx-webstorage-service';
 import {Observable, of} from 'rxjs';
-import {delay, retry} from 'rxjs/operators';
+import {delay, retry, tap} from 'rxjs/operators';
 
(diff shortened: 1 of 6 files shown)


src/app/identity.service.ts
```diff
@@ -41,11 +41,16 @@ export class IdentityService {
...
+    return this.http.post<{ id: string, jwt: string }>(
       `${environment.identityApiPrefix}${this.loginPath}`,
       credentials,
-    );
+    ).pipe(tap({
+      next: (response)  => {
+        this.saveJWT(response.id, response.jwt);
+        this.loginStatusChanged.emit(true);
+      }
+    }));
```
dorota-joanna, line 53: @bdsl, nice use of pipe + tap to return a function!

dorota-joanna (approved):
LGTM

PR #1439, opened 2023-12-18
DON-948: Don't display logged in status on donate page if donor has n…

bdsl:
…ot set a password

This is needed for consistency with the rest of the site, which doesn't treat you as really "logged in" unless you have set a password for your account.

Before:

image

After:

image


dorota-joanna (approved):
LGTM

bdsl:
Closing for now - I think we need to have more discussion on the ticket about what we actually want to do here.

PR #1441, opened 2024-01-02
DON-934, DON-935 – check in latest components

NoelLH:
Making a draft PR with the latest version for now.

This isn't working properly for me – it seems like the cloning approach has issues when in Angular. The correct elements are logged and supposedly being copied but `#sleeve_2` (for example) has ticker items with blank content in their slots.


NoelLH:
Cloning still isn't right, so the latest version simplifies things and always uses exactly 1 copy, so as not to make things worse on some breakpoints by waiting for copies that don't exist. We can track making it optimal on COM-97.

PR #1442, opened 2024-01-03
DON-919: Update homepage cards - 2024 application season

bdsl:
Screenshot as it looks now (same as in prod already):

image

Screenshot with simulated future date of Jan 10th:

http://localhost:4200/?simulatedDate=2024-01-10

image

And then on Feb 1st it will show

image

And with a simulated date after Feb 9th it looks a bit empty but probably we'll have put something else in before that happens:

image

### DIFF
--- a/src/app/home/home.component.ts
+++ b/src/app/home/home.component.ts
@@ -10,7 +10,7 @@ import {environment} from "../../environments/environment";
 
 const CCCloseDate = new Date('2023-12-05T12:00:00+00:00');
 const Jan8th = new Date('2024-01-08T12:00:00+00:00');
-const Feb9th = new Date('2024-02-09T12:00:00+00:00');
+const Feb9th = new Date('2024-02-09T23:59:59+00:00');
 const ArtsForImpactApplicationCloseDate = new Date('2023-12-16T00:00:00+00:00');
 const GMFApplicationCloseDate = new Date('2024-01-17T00:00:00+00:00');
 
--- a/src/app/home/home.component.ts
+++ b/src/app/home/home.component.ts
@@ -8,8 +8,9 @@ import {PageMetaService} from '../page-meta.service';
 import {HighlightCard} from "./HighlightCard";
 import {environment} from "../../environments/environment";
 
-const CCOpenDate = new Date('2023-11-28T12:00:00+00:00');
-const CCCloseDate = new Date('2023-12-05T12:00:00+00:00')
+const CCCloseDate = new Date('2023-12-05T12:00:00+00:00');
+const Jan8th = new Date('2024-01-08T12:00:00+00:00');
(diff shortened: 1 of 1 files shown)


src/app/home/home.component.ts
```diff
@@ -8,8 +8,9 @@ import {PageMetaService} from '../page-meta.service';
 import {HighlightCard} from "./HighlightCard";
 import {environment} from "../../environments/environment";
 
-const CCOpenDate = new Date('2023-11-28T12:00:00+00:00');
-const CCCloseDate = new Date('2023-12-05T12:00:00+00:00')
+const CCCloseDate = new Date('2023-12-05T12:00:00+00:00');
+const Jan8th = new Date('2024-01-08T12:00:00+00:00');
+const Feb9th = new Date('2024-02-09T12:00:00+00:00');
```
bdsl, line 13: I'm not sure what times of day we want to use - I've put 12 noon in because that's what we've often done before, but would be good to confirm. @BethBigGive ?

src/app/home/home.component.ts
```diff
@@ -35,32 +36,32 @@ export class HomeComponent implements OnInit {
...
   private highlightCards: readonly HighlightCard[] = [
     {
-      appearAt: 'asap',
-      disappearAt: CCOpenDate,
-      headerText: 'Christmas Challenge 2023 is starting soon!',
-      bodyText: 'Save the date 28th November - 5th December 2023',
-      iconColor: "brand-cc-red",
-      backgroundImageUrl: new URL('/assets/images/card-background-cc-lights.jpg', environment.donateGlobalUriPrefix),
+      appearAt: Jan8th,
+      disappearAt: Feb9th,
```
bdsl, line 40: I assumed this and the one below have to disappear on Feb9th - is that right @BethBigGive and if so do we want to replace them with anything?

src/app/home/home.component.ts
```diff
@@ -95,8 +96,20 @@ export class HomeComponent implements OnInit {
...
+      bodyText: 'You donate.\nWe double it.',
+      backgroundImageUrl: new URL('/assets/images/blue-texture.jpg', environment.donateGlobalUriPrefix),
+      iconColor: 'primary',
+      button: {
+        text: "Explore now",
+        href: new URL('/explore/', environment.donateGlobalUriPrefix)
+      }
+    },
+    {
+      appearAt: Feb9th,
```
bdsl, line 111: if the two new cards disappear on Feb9th then I suppose there will be space to show this one again.

src/app/home/home.component.ts
dorota-joanna, line 13:
```suggestion
const Feb9th = new Date('2024-02-09T23:59:59+00:00');
```

dorota-joanna (approved):
LGTM

PR #1446, opened 2024-01-09
DON-948: Treat user as not logged in on donate page if they do not ha…

bdsl:
…ve a password

Passwordless donor accounts are necassary for viewing thanks pages, but sort of confusing. They go away after the session anyway, so we don't want to let them block donors logging in to a full passworded account.

The donate page will now always show either the login button or the "you are logged in" message, and it only shows the latter in the case that the donor has set a password (either in the current session or previously)

No other changes.

image


NoelLH (approved):
Looks good!

bdsl:
Thanks @NoelLH ! I think I'll wait until we've got a frontend deploy out before merging this - I don't want to create any reason to delay the next deploy any more.

PR #1448, opened 2024-01-10
DON-752: Add initial version of registration page

bdsl:
Blend of code copied from existing login component and from the registration modal.

image


src/app/register/register.component.html
```diff
@@ -0,0 +1,92 @@
...
+            <label slot="label" for="emailAddress">Email address *</label>
+            <input
+              slot="input"
+              matInput type="email" id="emailAddress" formControlName="emailAddress" autocapitalize="off">
+          </biggive-text-input>
+          <biggive-text-input spaceBelow="4">
+            <label slot="label" for="password">Password *</label>
+            <input
+              slot="input"
+              matInput type="password" id="password" formControlName="password">
```
bdsl, line 58: We could add a "repeat password" input, but I think it's probably not necessary, and we don't have one on the existing registration modal. See https://ux.stackexchange.com/questions/484/is-a-repeat-password-field-necessary-in-a-signup-page

Although we're not requiring the email address to be confirmed before the new account is used, so if someone's email doesn't work and they mistype their password they will have no way to recover the account.

src/app/register/register.component.scss
```diff
@@ -0,0 +1,131 @@
+@use '@angular/material' as mat;
+@import '../../abstract';
+
+// large parts of this file are copied from either my-account.component.scss for the general layout,
+// or donation-start-form.component.scss for the text input styling.
+// We might want to refactor to remove duplication, but I'm not sure where the common code would fit best,
+// and I think it probably makes sense to wait until we have a third copy and then factor out common parts.
```
bdsl, line 7: this is now the third copy so I think the refactor needs to be done, probably before merging this.

src/app/register/register.component.html
```diff
@@ -0,0 +1,92 @@
...
+        #captcha
+        size="invisible"
+        errorMode="handled"
+        (resolved)="captchaReturn($event)"
+        (errored)="captchaError()"
+        siteKey="{{ recaptchaIdSiteKey }}"
+      ></re-captcha>
+
+      <p>
+        Registering as a donor with Big Give will help you donate to multiple campaigns, and will you the option
```
bdsl, line 25:
```suggestion
        Registering as a donor with Big Give will help you donate to multiple campaigns, and will give you the option
```

dorota-joanna (approved):
LGTM

PR #1451, opened 2024-01-11
DON-752: Redirect user via login page if attempt to see donation fund…

bdsl:
…s form without first logging in

This means the registration modal will not be used once the login page is launched


dorota-joanna (approved):
LGTM

PR #1453, opened 2024-01-12
DON-752: Dedicate donor registration page specifically to donation fu…

bdsl:
…nds users

As far as we know there's no reason anyone would want to explictly register on our site other than to use the donation funds feature. Nothing else requires registration except saving a payment method, and the way to do that is to make a donation and then set a password on the thanks page. Copy from ticket comments.

image

image


dorota-joanna (approved):
LGTM

PR #1454, opened 2024-01-12
DON-752: Adjust You Account page layout

bdsl:
Per comments on ticket.

I don't quite like the placement of "Log out", as it doesn't vertically align with anything. I think eventually we might want to remove it from here and put it in the main menu instead, but that should probably be considered out of scope of the current ticket.

image

It's not shown in the screenshot above because I deleted my card from my local, but the card edit and delete buttons are changed from buttons to white underlined text links. We should check how they work with screen readers.

### DIFF
--- a/src/app/my-account/my-account.component.html
+++ b/src/app/my-account/my-account.component.html
@@ -34,12 +34,36 @@
       </table>
     </biggive-page-section>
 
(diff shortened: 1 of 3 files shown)


dorota-joanna (approved):
LGTM

PR #1457, opened 2024-01-15
DON-735: Adjust layout for single highlight cards on home page

bdsl:
We can handle 1,2, or 3 highlight cards on the homepage. 2 or more use the grid component as in the past, a single card is treated as a special case.

1 card:

image

2 cards:

image

3 cards:

image

6 cards is not supported, but in case it happened this is how it would look:

image


dorota-joanna (approved):
LGTM

PR #1460, opened 2024-01-18
DON-954-thanks-copy-tweak

bdsl:


bdsl:
First version of tweaked copy looks like this. I think it's a bit too big:

image

src/app/donation-complete/donation-complete.component.html
```diff
@@ -28,7 +28,15 @@
       <div class="thank-you">
         <h2>Thank you!</h2>
 
-        <p class="thank-you-text">Your generous donation has been made successfully</p>
+        <p class="thank-you-text">
+          Your generous donation of {{ donation.donationAmount | exactCurrency:donation.currencyCode }} has been made.
+          That’s worth {{ totalValue | exactCurrency:donation.currencyCode }} to {{ donation.charityName }}.
```
bdsl, line 33: changed `,` to `.` at end here. I think `.` is probably better anyway, and it's simpler as the next words have to be conditional on how the donation was paid.

src/app/donation-complete/donation-complete.component.html
```diff
@@ -28,7 +28,15 @@
...
+
+          <span *ngIf="donation.pspMethodType === 'customer_balance'">
+            You can check your remaining funds you have by going to <a href="/{{myAccountPath}}">My Account</a>
+            or when you go to make another donation.
```
bdsl, line 37: There could be some confusion here because "my account" only shows the balance if it's non-zero. We could think about showing the balance for any logged in donor now that the funds feature is well tested by real donors, or do something a bit more complicated to show zero to donation funds users with zero balance but nothing to other donors who have zero because they never used the funds feature.

PR #1463, opened 2024-01-18
Fix DON-946: Stop stripe blocking retrying donation

bdsl:
For most errors we ask people to fix them in place before they progress to the next step, but that doesn't work for stripe errors - in that case they need to continue through the form and press the confirm button again.

This reverses some of the work done in DON-493, which I think no longer makes sense now we are confirming payments on the server side only.

Now if you use get a card decline the error looks like this, but it doesn't block you progress through the form and pressing confirm again. You can keep trying as many times as you like until the donation works, which it may do if you add funds to your bank account at the same time:

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,7 @@ orbs:
 
 jobs:
   test:
+    resource_class: large
     docker:
       - image: cimg/node:16.18-browsers
         auth:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -10,6 +10,7 @@ orbs:
 
 jobs:
   test:
+    resource_class: large
```
bdsl, line 13: the large resource is included in our CircleCI plan, and should be faster than the default medium.

dorota-joanna (approved):
LGTM

PR #1464, opened 2024-01-19
DON-954: Adjust donation page copy

bdsl:
New thanks copy looks like this:

image


bdsl:
For a small donation made with card, not logged in, we see this:

image

bdsl:
After setting a password and choosing not to log in it shows :

image

But then refreshing the page changes it to 
image

which is not right, we shouldn't be asking them to set a password again. I think that means we need to find somewhere to persist the info that the donor has a password already that will survive a refresh and be available when not logged in.

src/app/donation-thanks/donation-thanks.component.ts
```diff
@@ -195,9 +238,7 @@ export class DonationCompleteComponent implements OnInit {
 
         // Try to patch the person only if they're not already a finalised donor account,
         // e.g. they could have set a password then reloaded this page.
-        if (this.identityService.isTokenForFinalisedUser(idAndJWT.jwt)) {
-          this.registrationComplete = true;
```
bdsl, line 199: deleted this line because we don't want to tell the person they are newly registered in case they registered six weeks ago.

dorota-joanna (approved):
LGTM

bdsl:
Thanks for testing @dorota-joanna ! We found a couple of problems, testing first with a donation that uses up part of the donor's funds, and then with a donation that uses all of it.

In the first case the funds remaining figure was out of date - it showed the amount from before making the donation, not after. We need to make it refresh or something.

And in the second case it didn't show either the message about funds remaining or no funds remaining. Possibly because stripe doesn't show zero funds as a figure.

PR #1466, opened 2024-01-22
Revert "Revert "DON-954: Adjust donation page copy""

bdsl:
Reverts thebiggive/donate-frontend#1465

Needs a few bits fixing before merge see https://github.com/thebiggive/donate-frontend/pull/1464#issuecomment-1904057518


dorota-joanna (approved):
LGTM, the donation funds amount is correctly updated on thank you page, including when 0 funds remaining

PR #1472, opened 2024-01-26
DON-735: Use appropriate colours, images & links for homepage cards from s…

bdsl:
…alesforce API

image

### DIFF
--- a/src/app/home/HighlightCard.ts
+++ b/src/app/home/HighlightCard.ts
@@ -25,3 +25,59 @@ export type HighlightCard = {
       href: URL,
   },
 }
+
+export type campaignFamilyName = 'christmasChallenge'|'summerGive'|'greenMatchFund'|'womenGirls'|'mentalHealthFund'|'artsforImpact';
+
+
+export type SfApiHighlightCard = Omit<HighlightCard, 'backgroundImageUrl'|'iconColor'|'button'> & {
+  campaignFamily: campaignFamilyName
+  button: {
+    text: string,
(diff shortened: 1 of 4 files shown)


dorota-joanna (approved):
LGTM

bdsl:
I'll also add in code here to replace the website origins from the production-appropriate ones that SF is returning to the specific thing we want in case we're in the local dev, staging or regression environment.

src/app/home/HighlightCard.ts
```diff
@@ -25,3 +25,59 @@ export type HighlightCard = {
...
+    christmasChallenge: 'brand-cc-red',
+    summerGive: 'brand-c4c-orange',
+    greenMatchFund: 'brand-gmf-green',
+    womenGirls: 'brand-wgmf-purple',
+    mentalHealthFund: 'brand-mhf-turquoise',
+    artsforImpact: 'brand-afa-pink',
+  };
+
+  const campaignFamilyBackgroundImages: Record<campaignFamilyName, URL> = {
+    christmasChallenge: new URL('/assets/images/card-background-cc-lights.jpg', donateGlobalUriPrefix),
```
bdsl, line 63: I would have liked to have either a test or a compile time check to make sure that these image files exist, and aren't typo'd or deleted when still required. But I couldn't find a way to do that, since the tests run in the browser and won't have access to the filesystem. I don't think it's essential.

dorota-joanna (approved):
LGTM

PR #1478, opened 2024-02-02
DON-930: Use correct total figure for metacampaign ticker

bdsl:
image

See related PR in Salesforce for more discussion.

### DIFF
--- a/src/app/fund.model.ts
+++ b/src/app/fund.model.ts
@@ -1,17 +1,17 @@
 /**
  * Only Champion Fundings are used in the app thus far, not Pledges.
  */
-export class Fund {
-  constructor(
+export type Fund = {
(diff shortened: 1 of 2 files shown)


src/app/fund.model.ts
```diff
@@ -1,17 +1,17 @@
 /**
  * Only Champion Fundings are used in the app thus far, not Pledges.
  */
-export class Fund {
-  constructor(
+export type Fund = {
```
bdsl, line 4: we were never calling the constructor, never actually making any object that has this as its prototype. We only used the class as a compile time type. Hence changing it to a type to better reflect reality.

dorota-joanna (approved):
LGTM

PR #1479, opened 2024-02-05
DON-576: Show list of zero cards case of error in SF homepage highlig…

bdsl:
…ht card API

Previously we would have thrown an error from the resolver which would have blocked rendering the rest of the homepage - as we found on Friday when there was a bug on the SF side.

### DIFF
--- a/src/app/highlight-cards-resolver.ts
+++ b/src/app/highlight-cards-resolver.ts
@@ -3,12 +3,20 @@ import {ActivatedRouteSnapshot, ResolveFn} from '@angular/router';
 import {environment} from "../environments/environment";
 import {inject} from "@angular/core";
 import {CampaignService} from "./campaign.service";
+import {catchError} from "rxjs/operators";
+import {of} from "rxjs";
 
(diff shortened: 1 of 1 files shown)


src/app/highlight-cards-resolver.ts
```diff
@@ -3,12 +3,20 @@ import {ActivatedRouteSnapshot, ResolveFn} from '@angular/router';
...
 export const highlightCardsResolver: ResolveFn<readonly HighlightCard[] | null> = (route: ActivatedRouteSnapshot) => {
   const campaignService = inject(CampaignService);
 
   if (route.queryParams.hasOwnProperty('highlightAPIEnabled') && environment.environmentId !== 'production') {
-    return campaignService.getHomePageHighlightCards();
+    return campaignService.getHomePageHighlightCards().pipe(
+      // If the HighlightCards API has any error we still want to show the rest of the homepage, so we catch the error
+      catchError(error => {
+        console.error("Error fetching hompepage highlight cards", error);
+        return of([]);
```
bdsl, line 17: If we wanted instead of an empty list we of highlight cards we could put a card here apologising for the error and saying we'll hope to fix it soon. Not sure if that's better.

dorota-joanna (approved):
LGTM

PR #1481, opened 2024-02-07
DON-955: update k2m closing date

dorota-joanna:


bdsl:
LGTM. So on the 10th the homepage will look like this:

http://localhost:4200/?simulatedDate=2024-02-10T00:00:00%2B00:00

image

bdsl (approved):

PR #1482, opened 2024-02-08
Don 576 charity carousel amends

bdsl:
- Turn of autoplay
- Add previous and next buttons

Screenshot from 2024-02-08 11-39-23
Screenshot from 2024-02-08 11-39-31
Screenshot from 2024-02-08 11-39-42
buttons

I wanted to find a way to control the slider from TS code, so I tried following the instructions at https://swiperjs.com/element#parameters-as-props but I couldn't get that to work. Seems like we don't need it for now though, I've done everything in the HTML code.

If we like this then as part of the other carousel ticket I'll probably extract it out into a component.

### DIFF
--- a/src/app/home/home.component.html
+++ b/src/app/home/home.component.html
@@ -91,6 +91,8 @@
       autoplay="false"
       loop="true"
       centered-slides="false"
+      navigation-next-el="#carousel-button-next"
+      navigation-prev-el="#carousel-button-prev"
     >
       <swiper-slide><img src="/assets/images/logo-the-passage-2023.png" alt="The Passage - Everyone deserves a place to call home" width="300"></swiper-slide>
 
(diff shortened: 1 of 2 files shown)


src/app/home/home.component.html
```diff
@@ -110,6 +112,19 @@
 
       <swiper-slide><img src="/assets/images/logo-the-brain-tumour-charity.png" alt="The Brain Tumour Charity" width="300"></swiper-slide>
 
+      <div slot="container-end" class="charity-carousel-navigation">
```
bdsl, line 115: code below copied from https://github.com/thebiggive/components/blob/3bc64cc5201b881264c50ca15e345b7e9c1f6b52/src/components/biggive-carousel/biggive-carousel.tsx#L105

src/app/home/home.component.scss
```diff
@@ -49,3 +49,44 @@ swiper-slide {
     padding: 0;
   }
 }
+
+.charity-carousel-navigation {
```
bdsl, line 53: code below copied and adapted from https://github.com/thebiggive/components/blob/a1f5f9542fbec587c61e18aec8c3a0c230329954/src/components/biggive-carousel/biggive-carousel.scss#L21

dorota-joanna (approved):
LGTM

PR #1488, opened 2024-02-12
DON-576: Charity carousel amends

bdsl:
Rebuilt carousel button SVGs to include circles with shadows, so we can control the position of the arrow within the circle easily.

image


dorota-joanna (approved):
LGTM

PR #1489, opened 2024-02-13
DON-576: Make buttons for carousel bigger and further apart and more …

bdsl:
…buttony

Before:

image

After:

image


dorota-joanna (approved):
LGTM

PR #1493, opened 2024-02-20
BG2-2554: Change validation on all donate form fields from required t…

bdsl:
…o requiredNotBlank

This will help people avoid setting e.g. just ' ' as their home address when home address is a required field.

Doing the same thing for other fields like name because we might as well
- I don't think we want to accept just a blank space as a name since its too easy to do by mistake.


dorota-joanna (approved):
LGTM

PR #1497, opened 2024-02-22
WPD-153 (etc): Update components version

bdsl:
Includes changes to biggive-form-field-select made for use in WP, and dev dependancy updates from dependabot. Shouldn't have any impact here.

I've checked that the country select still works in the donate form and my selected country code gets saved on the new donation in the matchbot DB.


dorota-joanna (approved):
LGTM

PR #1500, opened 2024-03-12
Update libs

NoelLH:
General `npm update`


bdsl (approved):
Haven't reviewed the diff of course, but I approve of the principle. At some point it would be worth making a plan to do this regularly.

PR #1504, opened 2024-03-20
DON-950: Upgrade to Typescript 5.4.2

bdsl:
At least 5.2 is required for Angular 17


bdsl:
Seems like we have to upgrade TS and Angular at the same time.

PR #1506, opened 2024-03-25
DON-950: Replace ngx-matomo library with ngx-matomo-client

bdsl:
Intending to remove ngx-matomo because it's not compatible yet with Angular 17

See https://github.com/Arnaud73/ngx-matomo/issues/107 and notes on ticket.


dorota-joanna (approved):
LGTM

bdsl:
Thanks @dorota-joanna . I think I'll wait until after https://github.com/thebiggive/donate-frontend/pull/1510 is deployed before merging this to develop, so that I can do some testing of this with Matomo between deploying to staging and deploying to prod.

PR #1519, opened 2024-04-16
DON-962 – add one more Matomo event, to facilitate new goal with funnel

NoelLH:


src/environments/environment.interface.ts
```diff
@@ -32,7 +32,7 @@ export interface Environment {
   matomoAbTest?: {
     name: string,
     variantName: string, // The one that's not 'original'.
-    startDate: string,
+    startDate: string, // In 'YYYY/MM/DD HH:MM:SS UTC' format – can usually copy from 'Embed code' in Matomo UI view of test.
```
bdsl, line 35: I'll make push a tiny PR after this to type this using `type MatomoDate = `${number}/${number}/${number} ${number}:${number}:${number} UTC` .

bdsl (approved):

PR #1524, opened 2024-04-24
Don-965 cookies preferences

bdsl:
image

image


src/app/app.component.ts
```diff
@@ -154,6 +178,13 @@ export class AppComponent implements AfterViewInit, OnDestroy, OnInit {
     this.cookiePreferenceService.agreeToAll();
   }
 
+  @HostListener('preferenceModalClosed', ['$event'])
+  onCookieBannerPreferenceModalClosed(_event: CustomEvent) {
+    if (this.showingDedicatedCookiePreferencesPage) {
+      this.router.navigateByUrl('/')
```
bdsl, line 184: I think navigating to the homepage is good enough, navigating back is a lot more complicated, since we'd only want to do that if it was back within the BG website.

There's no reason for users to expect to be taken back to where they were, I think we can trust them to have only clicked the preferences centre link when they weren't in the middle of another task that they didn't mind stopping, and they know where to find the browser button if they want it. IMHO navigating back here would add user confusion if anything.

NoelLH: I don't really agree that it's confusing to return to what you had before upon closing what looks like a modal. But don't think it's a big deal either way.

NoelLH (approved):

src/app/app.component.ts
bdsl: Maybe not. Ideally since we navigated away this would not be a modal at all, not have a close button, user would have to choose if and when to navigate away.

PR #1525, opened 2024-04-25
DON-982 – Limit Explore to Active campaigns

NoelLH:
Needs a bit more discussion about our ideal Explore order before we make a change I think, especially as a more important sorting change is in progress. Draft for now.


bdsl:
Closing as we decided not to do this for now.

PR #1535, opened 2024-05-07
Drop latest tip A/B test 'A' variant

NoelLH:


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -269,14 +267,6 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
 
     this.tipControlStyle = (queryParams?.tipControl?.toLowerCase() === 'slider')
       ? 'slider' : 'dropdown';
-
-    if (! environment.production) {
-      this.manuallySelectedABTestVariant = queryParams?.selectABTestVariant;
```
NoelLH, line 274: Slightly in two minds about dropping this vs. keeping to reduce the diff for preparing the next test. The component variable which it changes had a name specific to the latest test though, and it seems a bit odd to keep the `if(...)` block but remove the substantive action. Open to suggestions.

bdsl: I think dropping it makes sense, if you think we might want to re-use the code and could miss it we could add a comment here mentioning that some code was deleted to tell us to go look for it in git history. But I'm not sure if this is the best place for the comment. I feel like we should be able to find it in git history anyway via ticket IDs etc.

bdsl (approved):

PR #1537, opened 2024-05-09
DON-950 – update to Angular 17 + misc housekeeping

NoelLH:
I think this mostly works now. When I test a built Docker container on port 4000 it:
* doesn't load in a visually perfect way – some things jump as it hands over to the client side, but I think this is something we've realised is already happening with the status quo Angular 16 situation as live now
* does load resources without any CORS errors with my latest tweaks, and following the corrected readme instructions

See individual commit messages for semi-related or unconnected housekeeping I noticed while working on the upgrade.


src/app/app.module.ts
```diff
@@ -68,17 +62,23 @@ const matomoTrackers = environment.matomoSiteId ? [
...
     { provide: MAT_CHECKBOX_DEFAULT_OPTIONS, useValue: { color: 'primary' } },
     { provide: MAT_RADIO_DEFAULT_OPTIONS, useValue: { color: 'primary' } },
-    { provide: RECAPTCHA_NONCE, useValue: environment.recaptchaNonce },
     {
-      provide: RECAPTCHA_BASE_URL,
-      useValue: 'https://recaptcha.net/recaptcha/api.js'  // using this URL instead of default google.com means we avoid google.com cookies.
-    }
+      provide: RECAPTCHA_LOADER_OPTIONS,
+      useValue: {
+        onBeforeLoad(_url: any) {
```
bdsl, line 74: From looking at https://www.npmjs.com/package/ng-recaptcha I think ng-recaptcha will pass a URL object here. Even if we don't use it I'd prefer to have the real type (or even `unknown` ) rather than `any`. 
```suggestion
        onBeforeLoad(_url: URL) {
```

bdsl (commented):
Added one suggestion inline, otherwise all LGTM. Don't feel you need to review again about whatever you do with that line.

PR #1544, opened 2024-05-09
Fix setting of recaptha URL that was breaking donation form

bdsl:


src/app/app.module.ts
```diff
@@ -71,9 +71,10 @@ const matomoBaseUri = 'https://biggive.matomo.cloud';
     {
       provide: RECAPTCHA_LOADER_OPTIONS,
       useValue: {
-        onBeforeLoad(_url: URL) {
+        // see ng-recaptcha/README.md
+        onBeforeLoad(_url: URL): {url: URL, nonce: string} {
           return {
-            baseUrl: 'https://recaptcha.net/recaptcha/api.js', // using this URL instead of default google.com means we avoid google.com cookies.
+            url: new URL("https://www.recaptcha.net/recaptcha/api.js"), // using this URL instead of default google.com means we avoid google.com cookies.
```
bdsl, line 77: the library destructures this object assigning url to baseUrl - which means that if url isn't set here it gives a misleading error message about `baseUrl` being undefined.

dorota-joanna (approved):
LGTM

src/app/app.module.ts
NoelLH: Oops, thanks! I even wondered why I couldn't see a `baseUrl` example on the latest readme, probably should have finished that thought..

PR #1545, opened 2024-05-09
Try to include the narrowest recaptcha CSP we expect to work

NoelLH:


bdsl:
Looks good but I'd hold off merging until after we have a green regtest run.

bdsl (approved):

PR #1550, opened 2024-05-15
Fix DON-976: Charity name obscures amount field

bdsl:
Just pulling in Components changes from Noel

PR to main rather than develop to allow for speedy fixing


dorota-joanna (approved):
LGTM

bdsl:
Closed as we found another way to fix the urgent issue, and this didn't fully fix it.

PR #1552, opened 2024-05-17
Try opting in to the new build system

NoelLH:
The easiest way from https://angular.io/guide/esbuild#for-existing-applications first


NoelLH:
Not a good start. Made DON-977 to look again down the line.

PR #1563, opened 2024-05-24
DON-971: Fix non-camel case sort field appearing in explore / metacam…

bdsl:
…paign URL


NoelLH (approved):
Nice, much clearer better typed!

bdsl:
Thanks @NoelLH - I think it's still a bit more complex than it needs to be, but this was the best I could do without going into components to change the events that sends.

PR #1564, opened 2024-06-05
DON-981: Reload page after login on donation page

bdsl:
I didn't fully get to the bottom of how the bug was happening, but I did manage to reproduce it reliably.

It seemed like the error message that's displayed "Please enter your billing postcode" may be unrelated to the actual block to progress.

I found that we were calling the stepper.next method at https://github.com/thebiggive/donate-frontend/blob/fc4a02a8469edf57c927d56c8a778dce002e22bf/src/app/donation-start/donation-start-form/donation-start-form.component.ts#L1084 but for some reason the stepper didn't visible advance. I don't know what's going inside the stepper or angular that means that doesn't happen, but refreshing the page fixes it. I tried wrapping the function in ngZone.run in case somehow we were outside a context where Angular could detect changes, but that didn't fix anything.

So solution I've chosen here is to just reload the page as soon as the user logs in. That fixes it, and I think still provides a good user experience. Logging in is always slightly slow, since it requires the user finding or remembering their password, and the backend server to verify it. I think users recognise that authentication is a cross-cutting concern (if not in those words) and so are used to the entire app potentially changing and reloading when they log in or log out. They already waited for the auth server to return so waiting very slightly longer for the page to reload hopefully won't feel annoying.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -2274,6 +2274,7 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
 
   public loadPerson(person: Person, id: string, jwt: string) {
     this.donor = person; // Should mean donations are attached to the Stripe Customer.
+    this.paymentStepErrors = "";
 
     // Only tokens for Identity users with a password have enough access to load payment methods, use credit
     // balances and access personal data beyond the anonymous new Customer basics.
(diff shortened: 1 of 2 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -2274,6 +2274,7 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
 
   public loadPerson(person: Person, id: string, jwt: string) {
     this.donor = person; // Should mean donations are attached to the Stripe Customer.
+    this.paymentStepErrors = "";
```
bdsl, line 2277: this is not necessary along with the change in the container, but I think is worth keeping just in case.

NoelLH (approved):

PR #1566, opened 2024-06-13
Revert "DON-981: Reload page after login on donation page"

NoelLH:
Reverts thebiggive/donate-frontend#1564

This didn't seem to fix the original problem in testing, and I have been trying my alternative fix approach without it as I branched off older `develop`. That seems to work now on https://github.com/thebiggive/donate-frontend/pull/1565.


dorota-joanna (approved):
LGTM

PR #1569, opened 2024-06-20
DON-981: Force recreating donation when logging in to account with funds

bdsl:
At the moment we have a bug where finalising the donation can be blocked when starting a pending donation and then logging in to an account that has donation funds, because we don't allow converting from a card donation to a funds donation.


NoelLH (approved):
Looks good. `prepareDonationCredits()` happens earlier than person info pre-fill so the reset shouldn't lose much potentially useful info that was on the form before. In theory we could helpfully pre-fill the amount box with the last donation amount but I don't think it's important.

bdsl:
@NoelLH Yes, and prefilling the amount is slightly complicated because we might not want to do that if it's more than the donor has available. Also in some cases maybe the donor put in £25k because we told them that's the max that we accept , but now they're logged in to an account with more funds than that they can donate up to £200k.

PR #1570, opened 2024-06-20
DON-713: Default to including only matched campaigns on explore page

bdsl:
If readers want to see unmatched campaigns they will need to click to dismiss the "Match Funded" pillbox. That change will be persisted in the address bar with a change from /explore to /explore?onlyMatching=false


bdsl:
Realised this isn't quite what we want to do.

PR #1572, opened 2024-06-21
Don 713 highlight cards on explore page

bdsl:
image

There are currently no highlight cards on our staging environment, so I added some hard-coded ones to test this. Kept that part out of the commit of course.

I've coded it to only show the highlight cards on the explore page if there are no more than 6 explore cards, and if there is no search term - I thought probably if someones doing a search then we want to just let them focus on that search. It might be worth also setting to only show the highlight cards if the are no filters turned on.

Also if there are no campaigns found from a search or selected filters (or if we just have no campaigns at all to show) we currently show "We can't find any campaigns matching this search but there are lots more to choose from.". I guess we don't want that at the same time as the highlight cards, but as coded so far this will show both together in that case:

image

### DIFF
--- a/src/app/explore/explore.component.spec.ts
+++ b/src/app/explore/explore.component.spec.ts
@@ -43,6 +43,9 @@ describe('ExploreComponent', () => {
       ],
       providers: [
         { provide: ActivatedRoute, useValue: {
+          snapshot: {
+            data: {}
+          },
           queryParams: of({}), // Let `loadQueryParamsAndRun()` subscribe without crashing.
         } },
       ],
--- a/src/app/highlight-cards/highlight-cards.component.ts
+++ b/src/app/highlight-cards/highlight-cards.component.ts
@@ -1,0 +1,16 @@
+import {Component, Input} from '@angular/core';
+import {HighlightCard} from "../highlight-cards/HighlightCard";
+import {ComponentsModule} from "@biggive/components-angular";
+
+@Component({
+  selector: 'app-highlight-cards',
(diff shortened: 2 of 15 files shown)


src/app/highlight-cards/highlight-cards.component.css
NoelLH, line 1: Can delete

src/app/highlight-cards/highlight-cards.component.ts
```diff
@@ -0,0 +1,16 @@
...
+import {ComponentsModule} from "@biggive/components-angular";
+
+@Component({
+  selector: 'app-highlight-cards',
+  standalone: true,
+  imports: [
+    ComponentsModule
+  ],
+  templateUrl: './highlight-cards.component.html',
+  styleUrl: './highlight-cards.component.css'
```
NoelLH, line 12:
```suggestion
```

src/app/explore/explore.component.spec.ts
```diff
@@ -41,8 +41,12 @@ describe('ExploreComponent', () => {
           },
         ]),
       ],
+      //         { provide: ActivatedRoute, useValue: { snapshot: { queryParams: {}, data: { campaigns: [],  highlights: []}} } },
```
NoelLH, line 44:
```suggestion
```

NoelLH (approved):

NoelLH:
I think having both messages when search term's applied is OK for now, it won't happen super often

PR #1575, opened 2024-06-24
DON-979: Use no-cookie version of youtube embed

bdsl:
This does seem to have the downside that the embed won't work on dev workstations, since we are not loading the site with HTTPs there, and fixing that doesn't seem that straightforward. But still should give some privacy improvement, and should work ok in staging etc. Looks like it's not trivial to get HTTPs working in dev: https://stackoverflow.com/questions/39210467/get-angular-cli-to-ng-serve-over-https

image

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -86,7 +86,7 @@ export function app(): express.Express {
           'js.stripe.com',
           'recaptcha.net',
           'www.gstatic.com',
-          // Both Vimeo's iframe embeds seem to need script access to not error with our current embed approach.
+          // Vimeo's iframe embed seems to need script access to not error with our current embed approach.
           'https://player.vimeo.com',
         ],
       },
--- a/server.ts
+++ b/server.ts
@@ -63,7 +63,8 @@ export function app(): express.Express {
           'player.vimeo.com',
           'recaptcha.net',
           'www.recaptcha.net',
-          'www.youtube.com',
+        ],
+        "frame-src": [
           'www.youtube-nocookie.com',
         ],
         'img-src': [
--- a/server.ts
+++ b/server.ts
@@ -64,6 +64,7 @@ export function app(): express.Express {
           'recaptcha.net',
           'www.recaptcha.net',
           'www.youtube.com',
+          'www.youtube-nocookie.com.com',
         ],
         'img-src': [
           `'self'`,
@@ -88,6 +89,7 @@ export function app(): express.Express {
           // Both video services' iframe embeds seem to need script access to not error with our current embed approach.
           'https://player.vimeo.com',
           'https://www.youtube.com/',
+          'https://www.youtube-nocookie.com',
         ],
       },
     },
--- a/server.ts
+++ b/server.ts
@@ -64,7 +64,7 @@ export function app(): express.Express {
           'recaptcha.net',
           'www.recaptcha.net',
           'www.youtube.com',
-          'www.youtube-nocookie.com.com',
+          'www.youtube-nocookie.com',
         ],
         'img-src': [
           `'self'`,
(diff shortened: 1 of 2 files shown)


server.ts
```diff
@@ -64,6 +64,7 @@ export function app(): express.Express {
           'recaptcha.net',
           'www.recaptcha.net',
           'www.youtube.com',
+          'www.youtube-nocookie.com.com',
```
NoelLH, line 67:
```suggestion
          'www.youtube-nocookie.com',
```

src/app/campaign-details/campaign-details.component.ts
```diff
@@ -105,7 +105,7 @@ export class CampaignDetailsComponent implements OnInit, OnDestroy {
 
     // As per https://angular.io/guide/security#bypass-security-apis constructing `SafeResourceUrl`s with these appends should be safe.
     if (campaign.video && campaign.video.provider === 'youtube') {
-      this.videoEmbedUrl = this.sanitizer.bypassSecurityTrustResourceUrl(`https://www.youtube.com/embed/${campaign.video.key}`);
+      this.videoEmbedUrl = this.sanitizer.bypassSecurityTrustResourceUrl(`https://www.youtube-nocookie.com.com/embed/${campaign.video.key}`);
```
NoelLH, line 108:
```suggestion
      this.videoEmbedUrl = this.sanitizer.bypassSecurityTrustResourceUrl(`https://www.youtube-nocookie.com/embed/${campaign.video.key}`);
```

server.ts
```diff
@@ -88,6 +89,7 @@ export function app(): express.Express {
           // Both video services' iframe embeds seem to need script access to not error with our current embed approach.
           'https://player.vimeo.com',
           'https://www.youtube.com/',
+          'https://www.youtube-nocookie.com',
```
NoelLH, line 92: Do we know that script access is needed for our own app? From https://makersaid.com/embed-youtube-without-cookies/ and the principle of the frame being more isolated for privacy reasons, I was hoping not.

server.ts
NoelLH, line 67: This might be a good opportunity to split out `frame-src`, with the caveat that we would need to comb through our other "defaults" and work out which are in fact used for frames, fonts, something else or redundant. A little fiddly but we'd end up with a slightly clearer security situation if we do it.

NoelLH (commented):
I'm hoping the double `.com`s explain the loading issue, SSL inside non-SSL when CSP is not relevant (dev mode) should certainly be allowed.

server.ts
bdsl: Fixed in f2af1a0d

server.ts
bdsl: No, apparently the script access isn't needed, I just tested on local. Removed in 681a61c0

server.ts
bdsl: I've moved youtube into frames-src in 6e8e3bf9 . Although confusingly the page seems seems to work fine and load the iframe on my local even if I comment out all youtube mentions entirely in server.ts

server.ts
```diff
@@ -85,9 +87,8 @@ export function app(): express.Express {
           'js.stripe.com',
           'recaptcha.net',
           'www.gstatic.com',
-          // Both video services' iframe embeds seem to need script access to not error with our current embed approach.
+          // Both Vimeo's iframe embeds seem to need script access to not error with our current embed approach.
```
NoelLH, line 90:
```suggestion
          // Vimeo's iframe embed seems to need script access to not error with our current embed approach.
```

NoelLH (approved):

PR #1583, opened 2024-07-01
Update big give components to version with built stencil/core 4.19.0

bdsl:
Expected to cause 15s slow initial load based on results from friday, but I want to deploy this change by itself to staging/regression to be sure whether it does cause the issue or if it must have been something else.


dorota-joanna (approved):
LGTM, matches the latest version of components

PR #1586, opened 2024-07-03
MAT-364: Auto retry donation if blocked by matchbot rate limiting

bdsl:


bdsl:
> Looks good so far.
> 
> Probably worth also clearing this in an `ngOnDestroy()` just to be certain that navigation away doesn't leave any retry efforts happening in the background?

Right, will do.

bdsl:
> Probably worth also clearing this in an `ngOnDestroy()` just to be certain that navigation away doesn't leave any retry efforts happening in the background?

Done in efac4951

NoelLH (approved):

PR #1587, opened 2024-07-04
DON-713: Remove obsolete `onlyMatching` param in search query sent to…

bdsl:
… SF.

SF now ignores this.


NoelLH (approved):
Makes sense. Made a new OpenAPI version https://app.swaggerhub.com/apis/Noel/TBG-Campaigns/1.0.32 with the deletion

PR #1590, opened 2024-07-08
DON-976: Allow long charity names to show over donation amount input

bdsl:
This isn't perfect, because the label is sometimes wider than we would like, so occludes a bit more of the border behind it on the right hand side than the left.  But I think it's probably good enough and it's not worth spending lots of time on trying to possibly make something better. Example of the label being a bit wider than we'd like:

image

To make it easier to review what we've got here I've made several copies of this donation amount input on the same form showing different length copy. The charity name here is also fictionalised. Of course these changes are not going to be checked into git. Video below to show how this looks over a range of screen widths:

[Screencast from 2024-07-08 10-37-01.webm](https://github.com/thebiggive/donate-frontend/assets/159481/9380e5a2-125f-4b4a-ae47-a5271d882495)

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -45,27 +45,22 @@
             </div>
           </div>
           <div class="donation-input donation-input-main">
+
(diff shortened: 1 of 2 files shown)


dorota-joanna (approved):
LGTM

PR #1593, opened 2024-07-22
Don 983 past donations

bdsl:
image

### DIFF
--- a/src/app/my-donations/my-donations.component.html
+++ b/src/app/my-donations/my-donations.component.html
@@ -1,0 +1,1 @@
+<p>my-donations works!</p>
--- a/src/app/my-donations/my-donations.component.html
+++ b/src/app/my-donations/my-donations.component.html
@@ -20,6 +20,7 @@
               <th scope="col">Date</th>
               <th scope="col">Charity</th>
               <th scope="col">Amount</th>
+              <th scope="col">Method</th>
               <th scope="col">Reference</th>
             </tr>
 
@@ -28,6 +29,7 @@
                 <td>{{ donation.createdTime | date: 'mediumDate' }}</td>
                 <td><a href="/campaign/{{donation.projectId}}">{{ donation.charityName }}</a></td>
                 <td>{{ donation.donationAmount | exactCurrency:donation.currencyCode }}</td>
+                <td>{{this.displayMethodType(donation)}}</td>
                 <td>{{ donation.transactionId }}</td>
               </tr>
             }
--- a/src/app/my-donations/my-donations.component.html
+++ b/src/app/my-donations/my-donations.component.html
@@ -47,7 +47,7 @@
     </biggive-page-section>
     <biggive-page-section>
       <p>
-        For full details of donations please see emails sent from hello&commat;biggive.org. We are unable to display any
+        For full details of donations please see emails sent from hello&#64;biggive.org. We are unable to display any
         donations you may have made without logging in.
       </p>
     </biggive-page-section>
--- a/src/app/my-donations/my-donations.component.html
+++ b/src/app/my-donations/my-donations.component.html
@@ -4,10 +4,7 @@
     <biggive-page-section>
       <div class="heading-wrapper">
         <biggive-heading
-          space-above=0
-          space-below=1
           colour="tertiary"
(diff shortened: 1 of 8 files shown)


bdsl:
See related in matchbot https://github.com/thebiggive/matchbot/pull/941 . This depends on the matchbot change so in theory that should go out before this, but it doesn't matter here if they're done in the wrong order since this PR doesn't release the new page.

bdsl:
Screenshot is slightly out of date - I've now also added a Method column, to show if each donation was made with a card or with donation funds.

bdsl:
I realised I forgot to design this for mobile on Monday, and it doesn't really work on mobile layout now. Might be reasonable to block merging until that's done, although since its feature flagged I think probably better not to block if that's the only issue and then I fix the mobile layout as my next task. 

image

image

src/app/my-donations/my-donations.component.scss
```diff
@@ -0,0 +1,37 @@
...
+  background-color: $colour-primary;
+  width: 100%;
+
+  @media #{$breakpoint-lg} {
+    width: 80%; // this seems to be needed, not sure why the identical rule in auth_common.scss isn't working.
+  }
+
+}
+
+// code above here mostly copied from my-account.component.scss
```
NoelLH, line 32: Maybe a good time to add a mixin instead?

src/app/my-donations/my-donations.component.ts
```diff
@@ -0,0 +1,59 @@
...
+import {Donation} from "../donation.model";
+import {AsyncPipe, DatePipe} from "@angular/common";
+import {ExactCurrencyPipe} from "../exact-currency.pipe";
+import {MatProgressSpinner} from "@angular/material/progress-spinner";
+
+@Component({
+  selector: 'app-my-donations',
+  standalone: true,
+  imports: [
+    ComponentsModule,
```
NoelLH, line 18: Maybe worth swapping for `...allChildComponentImports`? I forget the details but I think things like router link behaviour were subtly different when some pages were missing that import – probably good to do it the same throughout the app.

src/app/my-donations/my-donations.component.html
```diff
@@ -0,0 +1,55 @@
...
+        }
+      } @else {
+        <div>
+          <mat-spinner color="accent" [diameter]="40" aria-label="Loading donations"></mat-spinner>
+        </div>
+      }
+    </biggive-page-section>
+    <biggive-page-section>
+      <p>
+        For full details of donations please see emails sent from hello&commat;biggive.org. We are unable to display any
```
NoelLH, line 50: Slightly surprised to see with Angular's new template feature they made it compulsory to escape _all_ `@`s, but they did!

Only thought is that the other place this was handled in the codebase we used `&#64;`. Probably better to pick one style for more reliable finding & changing in future.

NoelLH (commented):
LGTM so far, notwithstanding the note about mobile layout. Some small things inline.

src/app/my-donations/my-donations.component.html
bdsl: Thanks, changed in63acf99d . I went with `&#64;` since that's what angular generated when we told it to convert templates to new syntax.

Having to escape @ now is very slightly annoying, but worth it for the new syntax. They use @ for directives now so it can't be used just to represent an @ sign.

src/app/my-donations/my-donations.component.ts
bdsl: Right, done in d1de1cd9 . I had just used the angular command to create the my-donations component, didn't remember about allChildComponentImports

src/app/my-donations/my-donations.component.scss
bdsl: I think that's pretty much what I was trying to do with auth-common: 

https://github.com/thebiggive/donate-frontend/blob/a6ed8d9c90ba94a1b1b658271a08eccf89658b03/src/assets/scss/auth-common.scss#L1 

but not sure why I still ended up needing to duplicate some stuff. I've done what I can for now to reduce duplication in 3d6016b4

NoelLH (approved):

PR #1595, opened 2024-07-25
DON-983: Remove loading spinner from donation history page

bdsl:
No reason not to load all this in advance of rendering the page.


dorota-joanna (approved):
LGTM

PR #1601, opened 2024-08-01
Don 985: Add friendly captcha on registration form (dev env only)

bdsl:
image

### DIFF
--- a/src/app/featureFlags.ts
+++ b/src/app/featureFlags.ts
@@ -4,6 +4,9 @@ import {EnvironmentID} from "../environments/environment.interface";
 const flagsForEnvironment = (environmentId: EnvironmentID) => {
   return {
     myDonationsEnabled: environmentId !== 'production',
+
+    // Friendly Captcha only in dev as we have to add support to identity service before enabling in staging.
+    friendlyCaptchaEnabled: environmentId === 'development',
   } as const;
 }
 
--- a/src/app/register/register.component.html
+++ b/src/app/register/register.component.html
@@ -83,6 +83,15 @@
         </div>
       </form>
       <div class="actions">
+        @if (flags.friendlyCaptchaEnabled) {
+          <!-- Sitekey below is shared between environments and necessarily exposed in FE, so should be fine to hard-code -->
+          <div
+            class="frc-captcha"
+            data-sitekey="FCMIOJ2ARSHLBGAJ"
+            data-puzzle-endpoint="https://api.friendlycaptcha.com/api/v1/puzzle, https://eu-api.friendlycaptcha.eu/api/v1/puzzle"
+            #frccaptcha
+          ></div>
+        }
         <div id="register-button">
           <div aria-live="polite">
             @if (processing) {
--- a/src/app/register/register.component.ts
+++ b/src/app/register/register.component.ts
@@ -74,6 +74,11 @@ export class RegisterComponent implements OnInit, OnDestroy, AfterViewInit {
     if (! flags.friendlyCaptchaEnabled) {
       return;
     }
+
+    if (! isPlatformBrowser(this.platformId)) {
+      return;
+    }
+
     const widget = new WidgetInstance(this.friendlyCaptcha.nativeElement, {
       doneCallback: (solution) => {
         this.friendlyCaptchaSolution = solution;
--- a/src/app/register/register.component.ts
+++ b/src/app/register/register.component.ts
@@ -115,15 +115,13 @@ export class RegisterComponent implements OnInit, OnDestroy, AfterViewInit {
     }
 
     this.processing = true;
-    if (! this.friendlyCaptchaSolution) {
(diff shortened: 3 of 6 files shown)


src/app/register/register.component.html
```diff
@@ -83,6 +83,15 @@
         </div>
       </form>
       <div class="actions">
+        @if (flags.friendlyCaptchaEnabled) {
+          <!-- Sitekey below is shared between environments and necessarily exposed in FE, so should be fine to hard-code -->
+          <div
+            class="frc-captcha"
+            data-sitekey="FCMIOJ2ARSHLBGAJ"
```
bdsl, line 90: I think the key is intended to be shared between envs - I couldn't actually find anything in the [friendly captcha docs](https://docs.friendlycaptcha.com/#/)  to confirm that or say if we should be setting each env up as a separate site.

src/app/featureFlags.ts
```diff
@@ -4,6 +4,9 @@ import {EnvironmentID} from "../environments/environment.interface";
 const flagsForEnvironment = (environmentId: EnvironmentID) => {
   return {
     myDonationsEnabled: environmentId !== 'production',
+
+    // Friendly Captcha only in dev as we have to add support to identity service before enabling in staging.
+    friendlyCaptchaEnabled: environmentId === 'development',
```
bdsl, line 9: (in dev the captcha result appears not to be verified currently so sending the response from FriendlyCaptcha instead of recapcha doesn't block registration)

NoelLH: I think that's only the case if you changed your local env to have `RECAPTCHA_BYPASS=1` @bdsl. I don't think we can assume it's true for everyone with that env.

src/app/register/register.component.ts
```diff
@@ -91,8 +115,15 @@ export class RegisterComponent implements OnInit, OnDestroy {
     }
 
     this.processing = true;
-    this.captcha.reset();
-    this.captcha.execute();
+    if (! this.friendlyCaptchaSolution) {
```
NoelLH, line 118: Don't we want to avoid reCAPTCHA running any time Friendly Captcha is enabled, not just when it already completed?

src/app/register/register.component.ts
```diff
@@ -63,6 +70,23 @@ export class RegisterComponent implements OnInit, OnDestroy {
     });
   }
 
+  ngAfterViewInit() {
+    if (! flags.friendlyCaptchaEnabled) {
+      return;
+    }
+    const widget = new WidgetInstance(this.friendlyCaptcha.nativeElement, {
```
NoelLH, line 77: I think this may run in a server side render too, and doesn't need to?

src/app/register/register.component.html
```diff
@@ -83,6 +83,15 @@
...
+            data-puzzle-endpoint="https://api.friendlycaptcha.com/api/v1/puzzle, https://eu-api.friendlycaptcha.eu/api/v1/puzzle"
+            #frccaptcha
```
NoelLH, line 92: Is there an established use of `frc` we're trying to match? This name & CSS class seems slightly confusing if the 'c' also stands for 'captcha'. `#friendlyCaptcha` clearer?

NoelLH (commented):
Mostly LGTM, few things inline

src/app/register/register.component.html
bdsl: This is pasted from the [Full example in Angular](https://docs.friendlycaptcha.com/#/widget_api?id=full-example-in-angular) in the docs. Can change it but especially since the ticket is to prototype rather than definitely use I was prioritising doing things the quick way.

src/app/register/register.component.ts
bdsl: Agreed - will change.

src/app/register/register.component.ts
bdsl: Probably yes. I was imagining while we are prototyping Friendly Captcha we'd have both on the page and accept either solution, but I don't think its necessary.

src/app/featureFlags.ts
bdsl: I don't have `RECAPTCHA_BYPASS=1` but I do have the key from here https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do set that allows all captchas to pass, same as we have in .env.example:

https://github.com/thebiggive/identity/blob/fc966a2fe4fe91351a7e6a708b11104195688ba7/.env.example#L11

src/app/register/register.component.ts
bdsl: Fixed in 55c5f374

src/app/register/register.component.ts
bdsl: Changed in 396453a57

NoelLH (approved):

PR #1602, opened 2024-08-02
DON-985: Add Friendly Captcha to donation start form, as alternative …

bdsl:
…to recaptcha

Recaptcha remains in place in envs where friendly captcha isn't used yet, currently everwhere except dev

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.scss
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.scss
@@ -516,3 +516,9 @@ mat-expansion-panel-header h3 {
   margin-top: 30px;
   position: absolute;
 }
+
+div.frc-captcha {
+  margin: auto;
+  background: none;
+  border: 2px solid $colour-primary;
+}
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -3,7 +3,7 @@ import {DatePipe, getCurrencySymbol, isPlatformBrowser} from '@angular/common';
 import {HttpErrorResponse} from '@angular/common/http';
 import {
   AfterContentChecked,
-  AfterContentInit,
+  AfterContentInit, AfterViewInit,
   ChangeDetectorRef,
   Component,
   ElementRef,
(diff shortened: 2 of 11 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.scss
```diff
@@ -516,3 +516,9 @@ mat-expansion-panel-header h3 {
   margin-top: 30px;
   position: absolute;
 }
+
+div.frc-captcha {
+  margin: auto;
+  background: none;
+  border: 2px solid $colour-primary;
+}
```
bdsl, line 524: this is slightly different to the one on the login page, on that I didn't apply any stiles so it follows the default white background with no border. I think no background and blue border looks better though, so I could change that too - probably will add the change to this PR.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -405,6 +412,34 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+      return;
+    }
+    const widget = new WidgetInstance(friendlyCaptcha.nativeElement, {
+      doneCallback: (solution) => {
+        this.idCaptchaCode = solution;
+        console.log('DONE: ', solution);
+      },
+      errorCallback: (b) => {
+        console.log('FAILED', b);
+      },
```
bdsl, line 437: noticed these console logs I pasted in from the friendly captcha docs. Think I'll delete. Not sure if/when we should expect an error, I guess we should add an apologetic error message under the captcha in case of error.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1115,6 +1150,10 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
       return;
     }
 
+    if (!this.donation && (this.idCaptchaCode || this.donor) && this.donationAmount > 0) {
```
bdsl, line 1153: what if captcha isn't ready? Maybe user was too quick to try to progress.

dorota-joanna (dismissed):
LGTM

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: Dealt with that in a159fcb337e102dc6ee4cd0608bffbb824f6adc7

dorota-joanna (approved):
LGTM

PR #1603, opened 2024-08-05
DON-985: Remove unused registration modal form

bdsl:
We now require login before the donation funds form is viewed, so the registration modal is never used.

Removinng registration modal will reduce the work needed to switch captcha provider

See https://github.com/thebiggive/donate-frontend/blob/66ad6aad5d383946ce7dbf2bd2bc189763e7a5af/src/app/app-routing.ts#L77


dorota-joanna (approved):
LGTM

PR #1612, opened 2024-08-13
DON-985: Change "Create my Donation Funds account" button copy to "Su…

bdsl:
…bmit Donation Funds transfer form"

The old copy doesn't really work because:

* For new users the account was created on the previous page when they clicked Register

* For returning users the account may have been created months or years ago, and they may be just topping up their donation funds.

Before:

image

After:

image

### DIFF
--- a/src/app/transfer-funds/transfer-funds.component.html
+++ b/src/app/transfer-funds/transfer-funds.component.html
@@ -209,12 +209,12 @@
         }
         <mat-step>
           <ng-template matStepLabel>Submission</ng-template>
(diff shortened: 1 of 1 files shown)


bdsl:
This changes doesn't require an updated to https://biggive.org/donation-funds/ - it actually makes the form more closely fit what that instructions page describes: "Once you have clicked ‘Submit’ at the end of the page".

dorota-joanna (approved):
LGTM

NoelLH (approved):

PR #1616, opened 2024-08-16
DON-1067 – focus correctable inputs when scrolling to them

NoelLH:


bdsl (dismissed):
LGTM. Not a blocker but as usual I'm looking for a way to avoid the focus element. In this case I think we could call .focus unconditionally - should be harmless if the element is non-focusuable as MDN says .focus "sets focus on the specified element, if it can be focused. ".

Or maybe replace `this.scrollTo(firstElInStepWithAngularError, true);` with `this.scrollTo(firstElInStepWithAngularError); firstElInStepWithAngularError.focus();` which is more verbose but not really any more complex.

NoelLH:
I think I would rather not give browsers, screen readers etc. the option to decide what to do about an explicit focus call on an element that we expect to be not an input. It feels like the kind of thing where implementations may vary in the real world.

Marginal but think I agree about avoiding the extra arg since only one call was saying `true` to focusing. Replaced that with some doc block guidance on `scrollTo()` in https://github.com/thebiggive/donate-frontend/pull/1616/commits/619337decc303aa9684a16a0310f4b09185085db

PR #1617, opened 2024-08-16
Don 983 restyle "my donations"

bdsl:
image


src/app/my-donations/my-donations.component.html
```diff
@@ -16,11 +16,23 @@
...
+              <biggive-container-card
+                card-colour="tertiary"
+                background-colour="primary"
+              >
                 <div class="donation">
-                  <a href="/campaign/{{donation.projectId}}">{{ donation.charityName }}</a> - {{ donation.createdTime | date: 'mediumDate' }} <br/>
-                  {{ donation.donationAmount | exactCurrency:donation.currencyCode }} ({{this.displayMethodType(donation)}}) <br/>
-                  Reference: {{ donation.transactionId }}
+                  <p>{{this.displayMethodType(donation)}}</p>
+                  <h3><a href="/campaign/{{donation.projectId}}">{{ donation.charityName }}</a></h3>
```
bdsl, line 25: Currently only displaying the charity name not the campaign name because the data sent from matchbot doesn't give the campaign name. But it should be very easy to add it to what matchbot sends, and I think that's worth doing. Not necessarily in this PR since I don't want to make the frontend deploy have to wait behind a matchbot deploy.

src/app/my-donations/my-donations.component.ts
```diff
@@ -44,4 +44,14 @@ export class MyDonationsComponent implements OnInit{
       case "customer_balance": return "Donation Funds payment"
     }
   }
+
+  protected totalCharged(donation: Donation): number
+  {
+    // this duplicates logic in DonationThanksComponent.setDonation and also in matchbot's function
+    // \MatchBot\Domain\Donation::getAmountFractionalIncTip . I think it's probably worth removing the duplication by
+    // putting all the logic in matchbot, and also running it just once and saving the result to the database so
+    // we know we're always displaying the total using the logic as it existed at the time the donation was made.
```
bdsl, line 53: would be good to get thoughts on the above

src/app/my-donations/my-donations.component.ts
```diff
@@ -44,4 +44,14 @@ export class MyDonationsComponent implements OnInit{
       case "customer_balance": return "Donation Funds payment"
     }
   }
+
+  protected totalCharged(donation: Donation): number
+  {
+    // this duplicates logic in DonationThanksComponent.setDonation and also in matchbot's function
+    // \MatchBot\Domain\Donation::getAmountFractionalIncTip . I think it's probably worth removing the duplication by
+    // putting all the logic in matchbot, and also running it just once and saving the result to the database so
```
bdsl, line 52:
```suggestion
    // putting all the logic in matchbot, and also running it just once and saving the result to the database (after it comes back from Stripe) so
```

src/app/my-donations/my-donations.component.ts
NoelLH: I'm not sure this would save us most of the existing frontend calls; it's good to be able to recalculate tip previews immediately as options are changed for example, rather than waiting for an API call. (If we move to a slider we also don't want to create a situation where fast changes mean we are choosing between accurately showing the predicted charge vs. avoiding hammering MatchBot excessively.)

It may well be a good idea to store the amount in a handy denormalised DB field once a charge is "finalised" though. (Scare quotes because we probably want to treat it as mutable, to account for tip refunds.) The two places Donate reads completed donations can use that.

src/app/my-donations/my-donations.component.scss
```diff
@@ -25,15 +25,19 @@ div.donation-container {
...
 div.donation {
-  border-radius: 5px;
-  background: white;
-  margin-bottom: 10px;
-  width: 190px;
-  color: $colour-primary;
-  padding-left: 10px;
-  overflow-wrap: break-word;
-  flex-basis: 27%;
+  width: 225px;
```
NoelLH, line 28: Not sure I get why we're going fixed width here, or why this particular width. Can we use `<biggive-grid/>` or possibly DIY `flex`, to stay closer to layouts that are already battle tested and consistently used in the rest of the apps?

NoelLH (dismissed):
Not sure about how we're doing the columns/outer layout now, otherwise all LGTM

src/app/my-donations/my-donations.component.ts
bdsl: @NoelLH Right, we'd might still have to do the calculation in frontend for incomplete donations at [DonationStartFormComponent.donationAndExtrasAmount](https://github.com/thebiggive/donate-frontend/blob/ca689271cd30da031a6f62a92075582cb77d1316/src/app/donation-start/donation-start-form/donation-start-form.component.ts#L1013). Although even there I think we could do it in matchbot easily enough because it's only shown in step 5 of the form, so it doesn't doesn't have to be recalculated every second when the options in step 1 are changed. 

But I'd missed that one on Friday, I only saw the duplication between here, the thanks page, and the matchbot logic currently used for actually  taking money. So we now have code for the same calculation in four places:

1. `\MatchBot\Domain\Donation::getAmountFractionalIncTip` (used for taking the donor's money)
2. Frontend `DonationStartFormComponent.donationAndExtrasAmount` (used for telling the donor what we will take)
3. Frontend `DonationThanksComponent.setDonation` (used for telling the donor what we took immediately after
4. Frontend `MyDonationsComponent.totalCharged` (used for telling the donor what we took on an old donation)

I think almost certainly we should combine 1, 3 and 4 by adding record of the amount to the db. Probably we should combine 2 as well. And for robustness it might be worth adding a check to matchbot to make it reject any attempt to confirm a donation (or turn on autoConfirmFromCashBalance) if the total to be charged as known to matchbot doesn't match the number frontend has. 

And agreed if we add it as a db field it would need to be mutable, it would be updated when `\MatchBot\Domain\Donation::setTipAmount` is called, as it is in the `if ($isTipRefund) {` section of [\MatchBot\Application\Actions\Hooks\StripePaymentsUpdate](https://github.com/thebiggive/matchbot/blob/315714bb5445db289d3f77515e199ad22856e78a/src/Application/Actions/Hooks/StripePaymentsUpdate.php#L45) . 

I think I would then also extract these two lines into a function and put it in the Donation model class:

```
            $donation->setPartialRefundDate($refundDate);
            $donation->setTipAmount('0.00');
```

bdsl: I don't think there's anything to do about it on this PR, but I think as we have the same logic now in four places and it is very important that they all agree I would DRY it up immediately under the banner of DON-983 after merging this.

PR #1619, opened 2024-08-20
DON-983: Use total paid figure from Matchbot for thanks and my donati…

bdsl:
…ons page

Avoids repeating the logic so many times to sum three parts of the donation and calculate total paid.

Do not merge before https://github.com/thebiggive/matchbot/pull/971 is in production


dorota-joanna (approved):
LGTM

PR #1620, opened 2024-08-21
Don 983 no feecover

bdsl:
As discussed feeCover is a feature that we introduced a few years ago but haven't ever used and are not likely to use. Now removing to simplify existing code and correct implementation of new features.

Best reviewed with "hide whitespace" turned on, as there are some indentation changes.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -338,8 +338,6 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
           getCurrencyMaxValidator(),
           Validators.pattern('^[£$]?[0-9]+?(\\.00)?$'),
         ]],
-        coverFee: [false],
-        feeCoverAmount: [null],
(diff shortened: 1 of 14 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1642,7 +1636,6 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
       currencyCode: this.campaign.currencyCode || 'GBP',
       donationAmount: this.donationAmount,
       donationMatched: this.campaign.isMatched,
-      feeCoverAmount: sanitiseCurrency(this.amountsGroup.value.feeCoverAmount),
```
bdsl, line 1645: this already defaults to zero in matchbot https://github.com/thebiggive/matchbot/blob/73149a162e3db3bf942953b15ea670a59daabbef/src/Application/HttpModels/DonationCreate.php#L36 so we don't need to include it here.

dorota-joanna (approved):
LGTM

PR #1623, opened 2024-08-23
DON-1025: Restyle regular giving page to match donation history page

bdsl:
image

### DIFF
--- a/src/app/my-regular-giving/my-regular-giving.component.html
+++ b/src/app/my-regular-giving/my-regular-giving.component.html
@@ -15,16 +15,27 @@
         @if (mandates.length > 0) {
           <biggive-grid columnCount="2">
             @for (mandate of mandates; track mandate.id) {
-              <biggive-basic-card
-                [icon]="false"
-                [textColour]="'primary'"
-                [mainImageUrl]="''"
-                [mainTitle]="mandate.charityName"
-                subtitle="{{(mandate.amount.amountInPence / 100) | exactCurrency:mandate.amount.currency }}, {{ordinal(mandate.schedule.dayOfMonth)}} of each month, {{mandate.giftAid ? 'with Gift Aid, ' : ''}} since {{ mandate.schedule.activeFrom | date: 'mediumDate'}}"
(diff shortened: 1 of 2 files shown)


src/app/my-regular-giving/my-regular-giving.component.html
```diff
@@ -15,16 +15,27 @@
...
+              <biggive-container-card
+                card-colour="white"
+                background-colour="primary"
               >
-              </biggive-basic-card>
+                <div class="mandate">
+                  <h3><a href="{{'/campaign/' + mandate.campaignId}}">{{ mandate.charityName }}</a></h3>
+                <ul>
+                  <li>Amount: {{(mandate.amount.amountInPence / 100) | exactCurrency:mandate.amount.currency }}</li>
+                  @if (mandate.giftAid) {
```
bdsl, line 26: @NoelLH We talked about moving the GA decision info off the mandate and onto the donor's person record, but I guess if we do that its still useful to do the multiplication by 1.25 for display. More reason to move that calculation into matchbot.

src/app/my-regular-giving/my-regular-giving.component.html
```diff
@@ -15,16 +15,27 @@
...
-                [buttonAlign]="'center'"
+              <biggive-container-card
+                card-colour="white"
+                background-colour="primary"
               >
-              </biggive-basic-card>
+                <div class="mandate">
+                  <h3><a href="{{'/campaign/' + mandate.campaignId}}">{{ mandate.charityName }}</a></h3>
+                <ul>
+                  <li>Amount: {{(mandate.amount.amountInPence / 100) | exactCurrency:mandate.amount.currency }}</li>
```
bdsl, line 25: @DomTBG asked for 2 digits after decimal point, but I'm not sure if we do want that. This is only showing a decimal point if it's not a whole number of pounds (which on this line it probably never will be).  This matches the style we use in [many other places across the site](https://github.com/search?q=repo%3Athebiggive%2Fdonate-frontend%20exactCurrency&type=code), and I don't know if there's any reason to deviate from it.

NoelLH (approved):

PR #1625, opened 2024-08-27
DON-1025: Show next expected payment date on regular giving mandate

bdsl:
image

### DIFF
--- a/src/app/my-regular-giving/my-regular-giving.component.html
+++ b/src/app/my-regular-giving/my-regular-giving.component.html
@@ -30,7 +30,7 @@
                     <li>Total with gift aid {{(1.25 * mandate.amount.amountInPence / 100) | exactCurrency:mandate.amount.currency }}</li>
                   }
                   <li>Payment collection day: {{mandate.schedule.dayOfMonth}} </li>
-<!--                  <li>Next expected payment date: {{ 'todo fill this in with date to be calculated in matchbot'}}</li>-->
+                  <li>Next expected payment date: {{ mandate.schedule.expectedNextPaymentDate | date: 'mediumDate'}}</li>
                   <li>Active from: {{ mandate.schedule.activeFrom | date: 'mediumDate'}}</li>
                 </ul>
                 </div>
--- a/src/app/mandate.model.ts
+++ b/src/app/mandate.model.ts
@@ -6,7 +6,8 @@ export interface Mandate {
   "schedule": {
     "type": "monthly",
     "dayOfMonth": number,
-    "activeFrom": string
+    "activeFrom": string,
+    "expectedNextPaymentDate": string
   },
   giftAid: boolean,
   "amount": {


NoelLH (dismissed):

src/app/my-regular-giving/my-regular-giving.component.html
```diff
@@ -30,7 +30,7 @@ <h3><a href="{{'/campaign/' + mandate.campaignId}}">{{ mandate.charityName }}</a
                     <li>Total with gift aid {{(1.25 * mandate.amount.amountInPence / 100) | exactCurrency:mandate.amount.currency }}</li>
                   }
                   <li>Payment collection day: {{mandate.schedule.dayOfMonth}} </li>
-<!--                  <li>Next expected payment date: {{ 'todo fill this in with date to be calculated in matchbot'}}</li>-->
+                  <li>Next expected payment date: {{ mandate.schedule.expectedNextPaymentDate | date: 'mediumDate'}}</li>
```
bdsl, line 33: @NoelLH As we discussed I've adjusted matchbot so it now returns a full ATOM format timestamp as `expectedNextPaymentDate` including the TZ offset. But then it's not clear now to make sure angular doesn't convert that to a different TZ for display.

I thought for a minute I could tell it to use the UK TZ bu adding something like `{provide: DATE_PIPE_DEFAULT_OPTIONS, useValue: {timezone: 'Europe/London'}},` to `NgModule.providers` in `app.module.ts`, but it isn't possible to pass a region based TZ name (open Angular issue [#48279](https://github.com/angular/angular/issues/48279)). 

I don't think there's any very easy way to get the timezone offset for London without using a library, e.g. the tiny one file library at https://www.npmjs.com/package/get-timezone-offset . We could use that library to get an offset to pass to `DATE_PIPE_DEFAULT_OPTIONS` . Still feels like we might be doing too much work when in principle we want everything to be UK so there is no need for conversion between different time zones for purpose of displaying this date. If we stick with using full timestamp objects we could also make them something like 9am or noon instead of midnight so that the date will always be the same in at least London and GMT. If we do noon it will be the same in almost all TZ's.

Thoughts on the best way ahead?

bdsl: Especially if/when this ECMA script proposal gets accepted I'd still rather switch to using only a date, i.e. what apparently JS will call a [Temporal.PlainDate](https://tc39.es/proposal-temporal/docs/index.html#Temporal-PlainDate), brick/datetime and Java's JodaTime calls a LocalDate, and will serialise for JSON etc as e.g. `2006-08-24`. That feels especially convenient when we know we never want to do any TZ conversion and we always want to display the same number that was input.

bdsl: Can also do the formatting to a displayable string in PHP where it seems easier to make sure everything is done in London TZ. Or leave it as it is and rely on almost all the donors having UK TZ set on their machines, and if they have anything else being unlikely to mind the date being displayed in their TZ instead of the UK one.

NoelLH: I feel like our best & simplest end point will probably be to use e.g. 6am anyway. That could be both our standard authorisation/capture time for payments and what we use to show donors a date, making this a little simpler as a side effect. (Because we have a datetime, when we support other countries we could also easily have a policy of capturing at e.g. 6am local.)

I don't have a super strong view on whether we should also do the work to be a bit more timezone-safe w.r.t. browser locales. To do so, I would probably go with the package you linked or an even shorter version of its `Intl` logic that does just what we need. But also, it looks like Angular could be on the way to using Intl, so we could also keep an eye on that and aim to switch to IANA if they get there before we move our focus back to other countries. I've watched the two relevant issues.

PR #1639, opened 2024-09-10
DON-896: Don't show Link.com payment method option

bdsl:


bdsl:
This led to error:

"Payment processing failed: Payment details were collected through Stripe Elements using payment_method_types and cannot be confirmed through the API configured with automatic payment methods. "

PR #1642, opened 2024-09-11
Don 1071 restyle stripe element

bdsl:
Continuing work started by @NoelLH 

Comparing before and after - before (staging) on left, after (my local) on right. The content is slightly different as I'm logged into different accounts, as that's a much quicker way to see both side by side. Change here is purely about styling. Probably needs a bit more work.

image

image

image

image

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -429,7 +429,6 @@
                       <p>Payment fields require the "captcha" puzzle to be solved before we can show them safely.</p>
                     </div>
                   }
-                  <!-- todo this should only show to non-passworded donors -->
                   <div>
                     @if (showCardReuseMessage) {
                       Your card information will be held for re-use only if you set a password on the next page. It is held securely only with Stripe.
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -429,6 +429,7 @@
                       <p>Payment fields require the "captcha" puzzle to be solved before we can show them safely.</p>
                     </div>
                   }
+                  <!-- todo this should only show to non-passworded donors -->
                   <div>
                     @if (showCardReuseMessage) {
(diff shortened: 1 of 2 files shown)


NoelLH (commented):
Nicer, looking like a really good improvement already.

I'd maybe make a few small visual changes:
1. I think we'd do well to give active/focused items a colour distinct from the outer borders, so that it's more immediately obvious which things you are changing and which are constant. This is a lot like we are thinking about on COM-99, and for similar reasons I would probably go for our brand's coral: it's got decent contrast against a light background, while not being the primary blue.
2. If we can also make the active tab's icon use that colour that might also help differentiate it, but I can't immediately see how to do that – might be that only built-in theme highlight colours support it as it's hard for Stripe to figure out all logos' contrasts. (If that's the case I would consider going back to default blue active highlight, as it's better to have a not perfectly on-brand experience than a less accessible one.)
3. I think I would make the standard border a slightly lighter grey, to make things feel a little less "boxy", as it is so close to another square outer border. (If we can't find a good accessible balance, we could possibly consider removing the outer container/border for the Element and turning the inner border purple, though I think we might then also need to switch to the accordion methods layout to make the choice harder to miss.)

bdsl:
@NoelLH Thanks - all those changes made in [7fa1f94](https://github.com/thebiggive/donate-frontend/pull/1642/commits/7fa1f94d94f048480b8db3b3491eff3ea750074c) . Looks like so now:

image

bdsl:
Removing the outer border entirely and making each input inside look like its been placed independently on our page is an interesting option, but I think we won't ever be able to get it to match entirely with our cut off corner and and "eybrow" text, so having the box around it probably is still necessary to give some sort of explanation of that.

bdsl:
Also adjusted style of the picker items for saved cards.

image

(shadow hover state for selected item: )

image

bdsl:
[2b137fd](https://github.com/thebiggive/donate-frontend/pull/1642/commits/2b137fdcd37302dfef9031f78e6e947b7540a2c2) sets selected tab outline coral:

image

bdsl:
And also put a coral outline on the highlighted card as it appears after you click "see more":

image

NoelLH (approved):
Looks really good!

PR #1647, opened 2024-09-13
DON-896: Restore previous behavour, autofill & hide billing postcode …

bdsl:
…& country if saved payment method selected

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -401,22 +401,22 @@
                     }
                   </div>
                 </div>
-                <div>
-                  <biggive-form-field-select
-                    [prompt]="'Billing country'"
-                    [options]="countryOptionsObject"
-                    [selectedValue]="selectedCountryCode"
-                    [backgroundColour]="'grey'"
-                    [selectionChanged]="setSelectedCountry"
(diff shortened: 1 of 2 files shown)


bdsl:
I'm not sure we're actually getting much if any value from the billing postcode field, but easier to keep it and restore the previous behaviour than get the confidence to remove it.

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -401,22 +401,22 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
...
-                    [prompt]="'Billing country'"
-                    [options]="countryOptionsObject"
-                    [selectedValue]="selectedCountryCode"
-                    [backgroundColour]="'grey'"
-                    [selectionChanged]="setSelectedCountry"
-                    [spaceBelow]="3"
-                    [selectedOptionColour]="'inherit'"
-                    >
-                  </biggive-form-field-select>
+                <div [style.display]="isSavedPaymentMethodSelected ? 'none' : 'block'">
```
bdsl, line 404: yesterday this was within a div with the stripe payment element that was hidden if the donor selected a saved card (or had donation funds)

NoelLH (approved):

PR #1663, opened 2024-09-20
Increase type coverage

bdsl:
I used `npx type-coverage -p tsconfig.app.json` to measure type covearge and npx type-coverage -p tsconfig.app.json --detail to find specific instances of `any` types.

Before:

	npx type-coverage -p tsconfig.app.json
	13598 / 14175 95.92%
	type-coverage success.

After:
	npx type-coverage -p tsconfig.app.json
	13682 / 14181 96.48%
	type-coverage success.

So the proportion of any types is reduced by 14%

### DIFF
--- a/src/app/meta-campaign/meta-campaign.component.ts
+++ b/src/app/meta-campaign/meta-campaign.component.ts
@@ -62,7 +62,7 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
   public tickerMainMessage: string;
   public title: string; // Includes fund info if applicable.
 
-  private autoScrollTimer: any; // State update setTimeout reference, for client side scroll to previous position.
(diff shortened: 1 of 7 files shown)


src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -394,7 +394,7 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
       // use a too-low Y position when lots of cards were shown and we didn't have a delay, both with `scrollToAnchor()`
       // and manual calculation + `scrollToPosition()`.
 
-      this.autoScrollTimer = setTimeout(() => {
+      this.autoScrollTimer = window.setTimeout(() => {
```
bdsl, line 397: using `window` tells TS that we're only interested in the function from the Browser api, not the one from node that returns a different type.

PR #1669, opened 2024-09-24
DON-1068: Adjust gift aid declaration - stop collecting declarations …

bdsl:
…for past & future donations

image


dorota-joanna (approved):
LGTM

PR #1671, opened 2024-09-25
DON-1068: Use different GA copy for tip or no tip

bdsl:
Adds some complexity as if the donor declares that they want to claim GA without a tip we now need to forget that declaration if they go back in the form and add a tip

image

image


dorota-joanna (approved):
LGTM

PR #1672, opened 2024-09-25
BG2-2739: A/B test adding test for zero tip

bdsl:
Implemented based on pattern of code removed in
https://github.com/thebiggive/donate-frontend/pull/1535

I noticed an existing problem with the "£# donation to support Big Give"
- we were displaying the end of the sentence without the begining if the tip amount form field was invalid (e.g. blank because not yet filled in)

So the logic I have implemented is:

If tip amount is not filled in with a valid value, show nothing

If running A/B test and in variant B and tip amount is filled in with zero show "Even £1 makes a big difference to Big Give"

If not running A/B test or not in variant B show
show "Even £1 makes a big difference to Big Give show "£0 donation to support Big Give" or "£5 donation to support Big Give" etc.

To test variant B in a non-production environment, add ?selectABTestVariant=B to the donate page address bar and refresh

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -116,7 +116,9 @@
                                 We transform your tips into match funds. Since 2022, charities have received over £200,000 in match funding from Big Give.
                               </li>
                             </ul>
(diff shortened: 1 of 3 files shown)


dorota-joanna (approved):
LGTM

PR #1676, opened 2024-09-27
DON-1073: Add check for consistent node version across ECR and SE

bdsl:
Our existing tests run in the browser. This new test is really static anysis using a test framework, so it needs access to the filesystem which isn't possible in the browser. Hence running it separately, using the node built-in test framework.

### DIFF
--- a/src/app/node-version.spec.mjs
+++ b/src/app/node-version.spec.mjs
@@ -1,0 +1,29 @@
+/**
+ * When we build the site for deployment to AWS S3 and AWS ECR/ECS we need to use a consistent node version,
+ * partly for compatibility with our code, but also to make sure that all generated js chunk checksums match up
+ * so that the two builds will work together as one app and not fall over with 404 errors as we saw in DON-1073
+ */
+
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { fileURLToPath } from "node:url";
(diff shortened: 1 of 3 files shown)


src/app/node-version.spec.mjs
```diff
@@ -0,0 +1,29 @@
...
+import { fileURLToPath } from "node:url";
+import { dirname, resolve } from 'path'
+import {readFileSync} from "node:fs";
+const expectedNodeVersion = '20.13.1';
+
+const modulePath = dirname(fileURLToPath(import.meta.url));
+
+test('should only have the expected node version mentioned in circle config for S3 build', async () => {
+  const circleConfig = readFileSync(resolve(modulePath, '../../.circleci/config.yml')).toString();
+  const nodeMentions = [...circleConfig.matchAll(/node[0-9\\.]+/g)];
```
NoelLH, line 18:
```suggestion
  const nodeMentions = [...circleConfig.matchAll(/node:[0-9\\.]+/g)];
```

Might be confused but I think that only the old cache key had `node` and numbers without a colon separator?

bdsl: @NoelLH you're right. Fixed in d062c5d2 . I also added an assert that there's at least one match in each file.

NoelLH (approved):

PR #1677, opened 2024-09-30
DON-1072: Switch login form to friendly captcha

bdsl:
image

Do not merge until after https://github.com/thebiggive/identity/pull/157 is in prod to support friendly captcha on login in identity service

### DIFF
--- a/src/app/login/login.component.scss
+++ b/src/app/login/login.component.scss
@@ -3,3 +3,12 @@
 #banner {
   background-image: url(../../assets/images/grey-header.jpg);
 }
+
+div.frc-captcha {
+  margin-top: 2em;
+  margin-right: auto;
+  margin-left: auto;
+  margin-bottom: auto;
+  background: none;
+  border: 2px solid $colour-primary;
+}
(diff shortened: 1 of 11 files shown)


src/app/login/login.component.scss
```diff
@@ -3,3 +3,12 @@
 #banner {
   background-image: url(../../assets/images/grey-header.jpg);
 }
+
+div.frc-captcha {
```
bdsl, line 7: we now have style for div.frc-captcha defined in three (here on login, donation-start-form and on register )places so should probably refactor either now or on the next one to remove the duplication and increase consistency. Not sure where its best to put it as a shared thing unless we make a new angular component

NoelLH: Sounds like it would be OK as a SCSS mixin in `src/assets/scss/_mixins.scss`, as it's not so big that repeats in the built CSS are a big problem.

NoelLH (approved):

src/app/login/login.component.scss
bdsl: OK - even though currently that file has nothing but actual `@mixin`s? I've got a few more re-captcha tags to remove/replace, so I can do that in the next PR along with those.

NoelLH: @bdsl Yes, I mean adding a `@mixin` there so that the is just a 1 line import in each place we need it, and one place the concrete rules live.

PR #1681, opened 2024-10-02
DON-1073: WIP use buildx script for staging as well as ECR for better…

bdsl:
… alignment


bdsl:
Closing in favour of #1682 which is much simpler and also seems to actually work.

PR #1682, opened 2024-10-03
DON-1073 – push S3 static files from the ECR image instead of building twice

NoelLH:


bdsl (dismissed):
LGTM - presumably if this works then we can do similar for `deploy-production-static` and `deploy-regression-static` which will be a nice bit of reduction in the complexity of this config file.

bdsl:
and looks like this also cuts a minute off the time it'll take us to deploy to staging, compared to the current way of running the build twice.

bdsl:
might be worth also either deleting [node-version.spec.mjs](https://github.com/thebiggive/donate-frontend/blob/d8f14a696e02440c7e79499c8cc2cbf9aa6c8b89/src/app/node-version.spec.mjs#L3) here or changing the comments inside it. When this is merged the reason for creating that test stops existing. There is still some value in making sure the node version we build and run tests on matches the version we deploy to, but its pretty small since the tests themselves don't actually run in node they run in a browser.

NoelLH:
> might be worth also either deleting [node-version.spec.mjs](https://github.com/thebiggive/donate-frontend/blob/d8f14a696e02440c7e79499c8cc2cbf9aa6c8b89/src/app/node-version.spec.mjs#L3) here or changing the comments inside it. When this is merged the reason for creating that test stops existing. There is still some value in making sure the node version we build and run tests on matches the version we deploy to, but its pretty small since the tests themselves don't actually run in node they run in a browser.

Yep, I think let's get it as simple as possible for now. Removed in 1d8b7773eae42b882603e97f927c26e7e8e0b30b

bdsl (approved):

PR #1683, opened 2024-10-03
Com 99 focus state input text & select

bdsl:
~The focus state on the drop-down is made visible just by updating the component version to one that has that feature built in, possible because the component handles creating the actual select element.~

~For the text input it doesn't seem possible to handle it the same way, because the actual input element is not part of the web component but is displayed within it via a slot instead. We need the input to be from a slot so that it can be part of forms, both here and in wordpress. I suppose in principle we could make the text input work more like the select, with the input built in and a change handler function passed in, but that would be quite a bit more work across this repo, wordpress, and the components library.~

~The change I've made here for the text input is only applied to the donation-start form, but for consistency we should probably apply it to all uses of `biggive-text-input` input across the site. That's currently `donation-start-form`, `register` and `login` within donate, plus `contact-biggive` on the Wordpress site. We can't share the css with WP but probably worth sharing it between the three parts of donate plus future pages. Other parts of donate, e.g. `login-modal` use text inputs but not our custom `biggive-text-input` so this style wouldn't apply. Maybe we'd want to unify that design as well though.~

~I'm not sure why the screencast below shows the focus indication line under the text input as greyish - when I actually view the page directly it's coral, matching the other stuff. Must be an artifact of the screen recording.~

-----------

Now with @NoelLH 's help this is just a component version update - changes made in the components library mean that both the select component and the text input component now include functionality of indicating their focus by changing the border from blue to coral. 

This affects all pages on this site (and soon in WP) that use these components - so for `biggive-form-field-select` that's currently just the donation start page here in donate-frontend and the contact form in wordpress. For `biggive-text-input` it's the donation page, register, and login here and contact form in wordpress.

[Screencast from 2024-10-04 12-33-43.webm](https://github.com/user-attachments/assets/bd995616-7645-4fb2-8ed9-7acf487baf50)


dorota-joanna (approved):
LGTM

PR #1686, opened 2024-10-07
DON-998: Create mostly page to hold regular giving mandate creation form

bdsl:
URI is `/regular-giving/<campaignID>` e.g. `/regular-giving/a05ds000000x4tRAAQ`

Feature flagged to not be available in production.

Features so far:
- The page can only be viewed by a logged in user. Attempt to access without logging in will trigger a redirect through the login form and back to the page

- Heading shows the name of the campaign and charity

Still required:

Everything else! e.g.
- Return 404 or display error if campaign is not set up for regular giving (see BG2-2711)
- Display campaign details and links in a nice way (probably with a sidebar similar to on donation page)
- Create actual form for donor to fill in details of what they want to give and confirm. Also the form needs to reflect back to them some of their logged in account details so they can make sure they're logged in with the details they want to use for this regular giving mandate

image

### DIFF
--- a/src/app/app-routing.ts
+++ b/src/app/app-routing.ts
@@ -1,4 +1,4 @@
-import {ActivatedRouteSnapshot, Router, Routes} from '@angular/router';
+import {ActivatedRouteSnapshot, CanActivateFn, Router, Routes} from '@angular/router';
 
 import {CampaignListResolver} from './campaign-list.resolver';
 import {CampaignResolver} from './campaign.resolver';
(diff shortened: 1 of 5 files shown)


src/app/app-routing.ts
```diff
@@ -51,7 +52,8 @@ const requireLogin = (activatedRoute: ActivatedRouteSnapshot) => {
   // on successful login the login page redirects back to My Account by default.
   // If we need to redirect to any other pages in future, we can take an ActivatedRouteSnapshot param here
   // and pass it to the login page as an 'r' query param.
-  const redirectPath = activatedRoute?.routeConfig?.path;
+
+  const redirectPath = routerStateSnapshot.url
```
bdsl, line 56: activate route path didn't include the actual params, e.g the campaign ID in this case. Hasn't been an issue up to now because none of the routes that have used this requireLogin guard have had path parameters.

bdsl:
I have somewhat dived into building this without much of a plan for how it should look, but we do know already what functionality needs to be supported, and I thought getting the start of something in place might help to make the things a bit more concrete. I've copied code liberally from our existing donation start form.

But once this is merged it's probably worth taking a step back and drawing out a sketch of how the the entire page should look and getting that agreed, and maybe refactoring this to reduce the duplication of code with the donation page, before spending lots more time implementing all the little parts.

dorota-joanna (approved):
LGTM

PR #1690, opened 2024-10-08
DON-998: Show logged in donor details on regular giving page

bdsl:
image

### DIFF
--- a/src/app/regular-giving/regular-giving.component.html
+++ b/src/app/regular-giving/regular-giving.component.html
@@ -72,50 +72,16 @@
                 account if necassary, or contact Big Give.
               </p>
 
-              <biggive-text-input>
-                <label
-                  class="label-with-limited-space"
-                  slot="label"
-                  for="donationAmount"
-                >
-                  First Name
-                </label>
-                <input
-                  readonly="readonly"
-                  value="{{donor.first_name}}"
-                  slot="input"
-                />
-              </biggive-text-input>
-
-              <biggive-text-input>
-                <label
-                  class="label-with-limited-space"
-                  slot="label"
-                  for="donationAmount"
-                >
-                  Last Name
-                </label>
-                <input
-                  readonly="readonly"
-                  value="{{donor.last_name}}"
-                  slot="input"
-                />
-              </biggive-text-input>
-
-              <biggive-text-input>
-                <label
-                  class="label-with-limited-space"
-                  slot="label"
-                  for="donationAmount"
-                >
-                  Email Address
-                </label>
-                <input
-                  readonly="readonly"
-                  value="{{donor.email_address}}"
-                  slot="input"
-                />
-              </biggive-text-input>
+              <table id="personal-details">
+                <tr>
+                  <td>Name</td>
+                  <td>{{donor.first_name}} {{donor.last_name}}</td>
+                </tr>
+                <tr>
+                  <td>Email</td>
+                  <td>{{donor.email_address}}</td>
+                </tr>
+              </table>
 
               <div style="text-align: center">
                 <button style="width: 40%;"
--- a/src/app/regular-giving/regular-giving.component.html
+++ b/src/app/regular-giving/regular-giving.component.html
@@ -61,17 +61,12 @@
 
             </mat-step>
             <mat-step
-              formGroupName="amounts"
-              label="Your details"
-              class="c-your-donation">
-
+              label="Confirm"
+              class="c-make-your-donation"
+            >
               <p>
-                The details below are set in <a href="/my-account">your account</a> and will be used for these
-                donations. They are not editable.
-                If these details are wrong please <a href="#" (click)="logout()">log out of this account and </a> create a new
-                account if necassary, or contact Big Give.
+                Your details
               </p>
-
               <table id="personal-details">
                 <tr>
                   <td>Name</td>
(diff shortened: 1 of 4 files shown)


src/app/regular-giving/regular-giving.component.html
```diff
@@ -59,6 +59,74 @@
...
+                  for="donationAmount"
+                >
+                  Email Address
+                </label>
+                <input
+                  readonly="readonly"
+                  value="{{donor.email_address}}"
+                  slot="input"
+                />
+              </biggive-text-input>
```
bdsl, line 118: not 100% sure if we should use read only input elements here - it's not visually obvious that they are readonly. We could just make a table or something to display these details, which might be clearer but need a bit more styling work.

NoelLH: I think that readonly with no styling changes will likely confuse and frustrate.

I think most browsers give `disabled` fields more obvious default styling – is there a reason to avoid that attr here?

In any case, I think we would do well to pick some simple styling of our own for whichever state we choose – at a minimum, a slightly darker grey background and a 'not allowed' cursor for desktop.

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -39,16 +41,25 @@ export class RegularGivingComponent implements OnInit {
...
     private snackBar: MatSnackBar,
+    private identityService: IdentityService,
+    private router: Router,
   ) {
   }
 
   ngOnInit() {
+    const donor: Person | null = this.route.snapshot.data.donor;
+    if (! donor) {
+      throw new Error("Must be logged in to see regular giving page");
```
NoelLH, line 59: What will happen when SSR mode gets a request for this?

NoelLH (commented):
I'm a bit concerned that server-side renders lacking the resolver could cause the handover to be glitchy and break the preview on any direct shares of this page, which e.g. we might want to have working when we get to the point of promoting a trail on socials.

Apart from that and the readonly input styling, LGTM.

bdsl:
> I'm a bit concerned that server-side renders lacking the resolver could cause the handover to be glitchy and break the preview on any direct shares of this page, which e.g. we might want to have working when we get to the point of promoting a trail on socials.
> 
> Apart from that and the readonly input styling, LGTM.

This page is already behind the `requireLogin` guard - doesn't that mean that there won't be any SSR renders anyway? The server doesn't know about logged in users, and I think we agreed that for simplicity we will only show the set up regular giving page to a logged in user.

src/app/regular-giving/regular-giving.component.html
bdsl: @NoelLH No particular reason not to have more obvious styling for disabled no - I didn't turn any styling off. Can add some disabled style, but do you think we even want to use input elements here? I'm leaning towards actually probably not. I can change it to use a list or table or something like that with no suggestion of interactivity.

dorota-joanna (dismissed):
LGTM

src/app/regular-giving/regular-giving.component.ts
bdsl: This page requires login which SSR doesn't do, so navigation will be cancelled at this line: https://github.com/thebiggive/donate-frontend/blob/d92d7b2fa70cce9471759794aed19cdf99397333/src/app/app-routing.ts#L44

src/app/regular-giving/regular-giving.component.html
bdsl: @NoelLH do you have a view on if input elements are useful here?

bdsl: Changing the background is a bit complicated since it needs to go through the component library. I think I'll try just not using inputs.

bdsl: I could simplify it to this. Although we store first & last name separately we display together on the account page, probably can do the same here. Maybe this doesn't need to actually have its own step though - we could just show these as part of the confirmation step.

image

bdsl: On my local I just moved these down into the confirm step like so: 
image

On my sketch I also had the editable fields billing country and billing postcode in this step, but I think it actually makes more sense to do them in the same step as the payment card selection, so I've pushed that change to this branch as 9795a5c2.

NoelLH:
> TBH I'm not sure exactly what the sharing preview is, but it seems like making a version of the page that doesn't need a person record add quite a bit of complexity that we might not need. And the requireLogin guard is already in place. Could we deal with removing it separately if you think we need to remove it?

I was thinking about Open Graph snippets and search engine indexing ability mostly. However I'm also not completely sure if real donors following direct links to this page will successfully land on it, should the initial SSR render fail. If not, that's more of a deal breaker than the bot considerations. (For the index of _all available_ campaigns for Regular Giving, non-logged-in bots are quite important too IMO.)

Charities will inevitably link their own donor lists directly to their own donation pages, so if this new page doesn't work when that happens we will have unhappy charities – so I think we need to at least test what happens there with the Docker-ised SSR app if we're not sure that it works.

I see that as far as the resolver itself goes, a `null` value is allowed. So I'm also not sure I understand what we gain by being so uncompromising and throwing an `Error` when that's the value. If we know that nulls are inevitable in Angular SSR mode, couldn't we do something more helpful for donors in both contexts by doing an `isPlatformBrowser()` check and:
* if `donor` is null on the server, print a simple text error message on the page explaining why it can't load further, making the rest of the page content conditional on `donor`
* if `donor` is null on the browser, redirect after view init to login and display a toast with the same explanatory copy

I think this is likely to give donors a less confusing experience, whether they reached this page via direct link or internal navigation.

bdsl:
sorry closed by mistake

bdsl:
Discussed very briefly with @NoelLH this morning - remaining things on this PR are not blockers but we'll try and discuss more tomorrow or next week.

PR #1691, opened 2024-10-09
DON-1072: Fix login page repeat attempt issue

bdsl:
We have to reset the captcha to generate a new code for each login attempt

Currently the login page only works if you get your credentials right on the first attempt - if you get them wrong initially and then try again the captcha code is 'used up' on the server and we don't generate a new one.

image


bdsl:
Closing as I rebased these commits onto main and made a separate PR #1693 . Will merge main back into develop to get them on staging etc.

PR #1696, opened 2024-10-10
Add `setupFutureUsage: 'on_session'` to Stripe element options

NoelLH:
Might swerve edge case bug where a token ends up with no value for future usage, causing server side crashes later. Not tested end to end as I'm not sure exactly when there are issues/ how to repro.


bdsl (approved):
LGTM but like you @NoelLH I'm confused about why this seems to be needed but only affecting a very small fraction of donors. I think lets merge and test on staging as with other changes, not rush to hastily get it into prod.

PR #1702, opened 2024-10-11
DON-1078: Fix stripe error for mismatched donor ID on stripe element …

bdsl:
…& donation

When a donor logs in after starting a donation we cancel that donation and create a new one. We also need to recreate the stripe elements on the page, so that they can confirm the donation and also so they get access to any saved cards on their account

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -506,12 +506,13 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
               'cancel_auto',
               `Donation cancelled due to donor authentication change`,
             );
-
+            this.destroyStripeElements();
             if (this.donation) {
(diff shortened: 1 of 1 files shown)


NoelLH (commented):
Resetting here definitely makes sense.

I had a look at the 3 other places we `cancel()` too: one is for amount changes – seems likely fine to leave Stripe.js alone; one only applies to credit and so I don't think Stripe.js will be used; the 3rd might need attention.

It's in `getDialogResponseFn()` where the donation was eligible to be continued but the answer was no (`proceed` = false). We go on to clear the donation and re-load some person info in that fn. I'm not totally clear whether that means we also need to reset Stripe Elements but it seems maybe safer to do so just in case?

bdsl:
@NoelLH I can't think of a scenario where we'd need to reset the stripe elements in `getDialogResponseFn` - it only needs to be reset if we have a new stripe customer ID, which I think is already handled by the change here. I think I'd rather only make the minimal change that we need for now.

PR #1704, opened 2024-10-11
Use clearer donor-account API endpoint

NoelLH:
Safe-ish to merge now, but given live campaign it's probably best to do a MatchBot Production deploy first, then merge this to `develop`.


bdsl (approved):
Looks like the relevant matchbot prod deploy was done in https://github.com/thebiggive/matchbot/pull/1033/files

PR #1707, opened 2024-10-14
Try adding `server.ts` to includes just for linting

NoelLH:
And not main build, which I think has maybe added warnings like "Warning: /home/circleci/******/server.ts is part of the TypeScript compilation but it's unused. Add only entry points to the 'files' or 'include' properties in your tsconfig."

### DIFF
--- a/tsconfig.app.json
+++ b/tsconfig.app.json
@@ -36,7 +36,6 @@
     "src/polyfills.ts"
   ],
   "include": [
-    "src/**/*.d.ts",
-    "server.ts",
+    "src/**/*.d.ts"
   ]
 }
--- a/tsconfig.eslint.json
+++ b/tsconfig.eslint.json
@@ -1,7 +1,8 @@
 {
   "extends": "./tsconfig.app.json",
   "include": [
-    "**/*.ts"
+    "**/*.ts",
+    "server.ts"
   ],
   "exclude": [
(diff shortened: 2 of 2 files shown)


bdsl:
I made a throwaway branch to see if this stops our CI build from picking up a type error in server.ts, and it does: See https://github.com/thebiggive/donate-frontend/compare/tsconfig-only-lint-server-fault-injected?expand=1

But maybe its OK even if it does, we don't edit the file often and even when we do its likely to always work or always fail, not just fail for some users because of an edge case that ts picks up as a type error. It's only fairly recently that we've had the type checking in CI on for this file and we've mostly been OK without it.

bdsl:
I won't explicitly approve for now since this does downgrade our type checking but also won't object if you want to merge it to stop the warning @NoelLH . I couldn't immediately find a way to get this file checked and not have the warning come up.

bdsl:
@NoelLH It might be linting on this branch but it isn't type checking - as you can see on my fault injection branch it allows me to pass a value of type `unknown` to a `string` parameter. I think lets merge this anyway as you say to reduce the noise in the build output.

Separately it would still be great if we could reduce the build output noise much further to  something easily readable so we can see anything that appears there - `npm run ci` still produces over 1900 lines of output.

PR #1708, opened 2024-10-14
Safer resumable donation loading

NoelLH:
I've had this in memory on my local for a while, can't remember if I saw a real issue on Staging or just developed a suspicion that the resume logic may not be totally safe when donation credits are involved. It seems like either way this change should be harmless even if not strictly necessary, and it does seem to me like it may well be required for donors who either currently or recently had credits to have donation resumption work properly.


bdsl (approved):
LGTM. Possibly we should add https://typescript-eslint.io/rules/no-floating-promises/ prevent mistakes like this.

NoelLH:
> LGTM. Possibly we should add https://typescript-eslint.io/rules/no-floating-promises/ prevent mistakes like this.

Yep, seems worthwhile. 44 errors on `develop` now but it looks like a lot can be resolved by reviewing and adding `void`.

PR #1713, opened 2024-10-16
Don 998 matchbot integration

bdsl:
image

### DIFF
--- a/src/app/regular-giving/regular-giving.component.ts
+++ b/src/app/regular-giving/regular-giving.component.ts
@@ -6,7 +6,7 @@ import {CampaignInfoComponent} from "../campaign-info/campaign-info.component";
 import {ImageService} from "../image.service";
 import {Observable} from "rxjs";
 import {AsyncPipe} from "@angular/common";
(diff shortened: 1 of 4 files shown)


src/app/regular-giving/regular-giving.component.ts
```diff
@@ -62,33 +69,87 @@ export class RegularGivingComponent implements OnInit {
...
+      dayOfMonth,
+      campaignId: this.campaign.id,
+      currency: "GBP",
+      giftAid: false
+    }).subscribe({
+    next: (mandate: Mandate) => {
+      const prefix = "Regular giving mandate created: (todo - make 'thanks' page with polling for updates and/or make " +
+        "the mandate come back activated so it will show on my regular giving page) ";
+
+      alert(prefix + JSON.stringify(mandate));
```
bdsl, line 135: of course we won't keep the `alert` for long but I think it's OK to merge it as-is for now. It gives a relatively easy way to test that the mandate is created appropriately before we've either updated the status to active or built anything to display a pending mandate.

PR #1721, opened 2024-10-21
DON-998: Remove regular giving sidebar

bdsl:
As discussed we don't need a sidebar for this page

Before:

image

After:

image

### DIFF
--- a/src/app/regular-giving/regular-giving.component.html
+++ b/src/app/regular-giving/regular-giving.component.html
@@ -1,6 +1,5 @@
 <main class="b-container">
   <div class="c-form-container">
-    <div class="b-primary-column">
       <biggive-page-section>
         <form
           class="c-regular-giving-form"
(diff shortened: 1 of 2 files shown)


src/app/regular-giving/regular-giving.component.html
```diff
@@ -96,34 +95,5 @@
...
-          id="campaign-highlights-beneficiaries-and-categories-desktop"
-        >
-          <div class="match-funded-badge-wrapper">
-            @if (campaign.matchFundsRemaining > 0) {
-              <div class="match-funded-badge">
-                Match Funded
-              </div>
-            }
-          </div>
-          @if (bannerUri$ | async; as bannerUri) {
```
NoelLH, line 113: Could also drop `bannerUri$` from TS I think

NoelLH (approved):
1 minor possible piece of tidying inline, not a blocker

src/app/regular-giving/regular-giving.component.html
bdsl: Thanks @NoelLH , will do after lunch then assuming its a minor change merge without further review.

PR #1725, opened 2024-10-22
COM-102: Allow using rectangular banner image for selected meta campa…

bdsl:
…ign page(s)

If commented out line is put back in then we get the following appearance. Clearly we would need to change the actual image first:

image


NoelLH (approved):
Looks good apart from package update now conflicting

PR #1753, opened 2024-11-14
DON-1085 – support tip cancellation when returning to the Transfer Funds page; fix broken tip create from non-whole default amounts

NoelLH:


dorota-joanna (approved):
LGTM

PR #1758, opened 2024-11-18
Fix CCClose date to allow redirect to happen this year

bdsl:
Also moved declaration closer to usage to make it easier to see


dorota-joanna (approved):
LGTM

PR #1759, opened 2024-11-18
Don 1079 ii

bdsl:


message-slack.mjs
```diff
@@ -0,0 +1,30 @@
...
+    "Authorization": "Bearer " + env.SLACK_ACCESS_TOKEN
+  },
+  body: JSON.stringify(
+    {
+      "channel": "deployments",
+      "blocks": [{
+        "type": "section",
+        "text": {
+          "type": "mrkdwn",
+          "text": `*${env.CIRCLE_BRANCH}: Donate frontend deployed!* All visitors should see the updated app version within 30 minutes.\n` +
```
NoelLH, line 20: I guess for `develop` we will lose the distinction between Staging & Regression here? Could it use the `env` job parameter that we already pass in for Jira?

NoelLH (approved):
Looks good! One minor thing inline.

I also wonder if we will later want to think about publishing this in our own orb if we want other apps to do very similar, but that's probably a question for later.

message-slack.mjs
bdsl: Sort of but we haven't been sending these messages for regression in the past, only for staging, and I'm continuing to only show it for staging. I don't think we really need the notifications for regression, it happens at the same time.

bdsl:
> I also wonder if we will later want to think about publishing this in our own orb if we want other apps to do very similar, but that's probably a question for later.

I'm not so keen on orbs, it feels like it ties us in to circleci too much, which isn't a problem in terms of wanting to stop using circle but makes them harder to debug since they can only be run from circle workflows - not as easily afaik from command line even inside a circle executor. If we wanted to publish something we could do via NPM but I don't think the script I've written here is worth publishing separately, may be as good to copy and paste between repos.

PR #1763, opened 2024-11-20
DON-1085: Improve validation UX in donation funds form

bdsl:
Only tangentially related to the ticket but as we're looking at this form to QA it seems like we may as well make these improvements

- Adjusted error copy to account for possible error of using a non-whole number of pounds. Potentially we could do a smallish amount of extra work to show the specific error in each case, but I don't think we need to now. We could also show the min and max amount at all times on the form as part of the instructions instead of only as an error but again that's a bigger change.

- Adjusted form components to only update on blur instead of in real time as the user types. This means the validation error message will only appear when they blur the input, which is quite a bit more user friendly. See https://ux.stackexchange.com/questions/48268/validate-on-blur-or-keypress

image


dorota-joanna (approved):
LGTM

PR #1764, opened 2024-11-21
DON-1086: Reduce font size in back buttons

bdsl:
("Back to campaign" on donate page, and "All [Participating/ChampionName] campaigns" on campaign details page)

Before:

image
image

After:

image
image


dorota-joanna (approved):
LGTM

PR #1768, opened 2024-11-22
DON-1070 – add post-script copy when `optInCharityEmail` choice is Yes

NoelLH:
Adds copy like the example shown below on my local.

Screenshot 2024-11-22 at 16 11 27

Personally I think it looks a bit cramped, but this is the most consistent appearance vs. other copy we put in similar places so probably the safest option for now.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -493,6 +493,10 @@
                     Please note that you might continue to receive communications from the charity if you have already shared your details with them via other methods.
                   </p>
                 }
+                @if (marketingGroup.value.optInCharityEmail === true) {
+                  <p aria-live="polite" class="no-margin-top b-rt-sm b-grey">
+                    As you are providing personal information directly to {{ campaign.charity.name }}, please make sure you read and agree to their privacy statement.</p>
+                }
               </mat-hint>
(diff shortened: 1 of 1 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -493,6 +493,10 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
                     Please note that you might continue to receive communications from the charity if you have already shared your details with them via other methods.
                   </p>
                 }
+                @if (marketingGroup.value.optInCharityEmail === true) {
```
bdsl, line 496: I guess we could simplify like this?
```suggestion
                @if (marketingGroup.value.optInCharityEmail) {
```

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -493,6 +493,10 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
...
+                  <p aria-live="polite" class="no-margin-top b-rt-sm b-grey">
+                    As you are providing personal information directly to {{ campaign.charity.name }}, please make sure you read and agree to their privacy statement.</p>
```
bdsl, line 498: It's not really direct, is it? We're passing the info on.
```suggestion
                    As you are providing personal information to {{ campaign.charity.name }}, please make sure you read and agree to their privacy statement.</p>
```

bdsl (approved):
LGTM. I added some minor suggestions inline, not blockers.

PR #1769, opened 2024-11-22
Fix notifying Jira of Production deploys

NoelLH:


.circleci/config.yml
```diff
@@ -88,6 +88,8 @@ jobs:
     parameters:
       env:
         type: string
+      env_type:
```
NoelLH, line 91: Thought I'd make this a new var explicitly so we have the option of adding `regression` later and are less likely to accidentally pass it as Jira environment_type which I think wouldn't work.

PR #1770, opened 2024-11-22
BG2-2745: Allow registration via registration page without viewing do…

bdsl:
…nation funds form

By default after registering new donor account holders will now be directed to the "my account" page.

I've kept the copy about transfering funds on the registration page but only in the special case where a donor is on the way to view the transfer funds form. It might be better to move this copy to the top of that actual form - although people need to see it without logging in so maybe even more ideally we'd allow at least part of the form to be viewed without logging in. More than we want to change now.

I've also kept the "quick and easy" copy, although I'm not convinced its helpful.

### DIFF
--- a/src/app/register/register.component.ts
+++ b/src/app/register/register.component.ts
@@ -10,12 +10,13 @@ import {FormBuilder, FormGroup, ReactiveFormsModule, Validators} from "@angular/
 import {IdentityService} from "../identity.service";
 import {environment} from "../../environments/environment";
 import {EMAIL_REGEXP} from "../validators/patterns";
-import {Router} from "@angular/router";
+import {ActivatedRoute, Router} from "@angular/router";
 import {MatAutocompleteModule} from "@angular/material/autocomplete";
 import {DomSanitizer, SafeHtml} from "@angular/platform-browser";
 import {transferFundsPath} from "../app-routing";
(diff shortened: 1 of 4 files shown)


src/app/register/register.component.ts
```diff
@@ -152,7 +163,8 @@ export class RegisterComponent implements OnInit, OnDestroy, AfterViewInit {
             next: () => {
               // We can't re-use a captcha code twice, so auto-login won't work right now. For now we just
               // redirect to the login form
-              window.location.href = "/login";
+              const newPath = "/login" + '?r=' + encodeURIComponent(this.redirectPath);
+              window.location.href = newPath;
```
NoelLH, line 167: I think it's likely to be more consistent with our other navigation and perhaps lighter on the browser to use the `router` that's already injected so that Angular can potentially update just what's needed in the already loaded app. Not a blocker.

NoelLH (approved):
Looks good! One suggestion on the mechanism for redirecting, no blockers.

src/app/register/register.component.ts
bdsl: Thanks, I think you're probably right but will keep it as-is since that's closer to the existing working code, and I don't want to spend more time than necessary on this before CC.

PR #1771, opened 2024-11-25
Fix BG2-2772: Donation not possible when navigating form without usin…

bdsl:
…g continue buttons


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -584,11 +584,13 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     const stepperHeaders = stepper.getElementsByClassName('mat-step-header');
     for (const stepperHeader of stepperHeaders) {
       stepperHeader.addEventListener('click', (clickEvent: any) => {
```
bdsl, line 586: showing the evils of `any`. If we had `clickEvent: PointerEvent` instead of `clickEvent: any` then TS would not allow us to write `clickEvent.target.index` or `clickEvent.target.innerText` since `target` is possibly null.

NoelLH (approved):

PR #1773, opened 2024-11-26
DON-1092 – refine post-search scroll behaviour

NoelLH:
* [give post-search scroll position change enough time to go to the right place](https://github.com/thebiggive/donate-frontend/commit/f47a4cb6747f02ec616db2bc6d459680749b57ad)
* [fix no-op navigation with scroll position impact](https://github.com/thebiggive/donate-frontend/commit/bb8e7363f37b6f121bb7461c904d8c4ccf5cf703)

### DIFF
--- a/src/app/meta-campaign/meta-campaign.component.ts
+++ b/src/app/meta-campaign/meta-campaign.component.ts
@@ -381,9 +381,15 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
    * Update the browser's query params when a sort or filter is applied.
    */
   private setQueryParams() {
-    this.router.navigate([], {
-      queryParams: this.searchService.getQueryParams(this.getDefaultSort()),
-    });
+    const nextQueryParams = this.searchService.getQueryParams(this.getDefaultSort());
+    if (JSON.stringify(this.route.snapshot.queryParams) === JSON.stringify(nextQueryParams)) {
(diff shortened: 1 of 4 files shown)


src/app/meta-campaign/meta-campaign.component.ts
```diff
@@ -188,7 +188,7 @@ export class MetaCampaignComponent implements AfterViewChecked, OnDestroy, OnIni
 
         // Angular scrolls automatically, using setTimeout to delay this scroll to a later task so this gets to
         // set the position the page is left in.
```
bdsl, line 190: probably worth changing this comment, "later task" made sense when delay was set at zero, now we're delaying it more than that. Not sure exactly what it is that we're waiting the 200ms for. Maybe something like this?

```suggestion
        // Angular scrolls automatically, using setTimeout to delay this scroll to happen after the automatic scroll
```

bdsl (approved):
LGTM, added a minor non-blocking suggestion inline.

src/app/meta-campaign/meta-campaign.component.ts
NoelLH: I'm also not sure exactly, but think it's connected to the router action and possibly to its `scrollPositionRestoration` feature which we have set to true.

Updated the comments to match my understanding better at https://github.com/thebiggive/donate-frontend/pull/1773/commits/1bec8637ce8186fdc606ea32cf9b5ff7bed30946

PR #1774, opened 2024-11-26
DON-1084: don't show banner on IOS >= 15.4

dorota-joanna:


bdsl:
I don't think uploading `[.browserslistrc](https://github.com/thebiggive/donate-frontend/pull/1774/files#diff-43302d1852d8010ee012aab4dda2d01abb664e11f73617fe625f2ba112a958ed)` by itself does anything, we'd need to also run `npm run supportedBrowsers` to update the generated code in https://github.com/thebiggive/donate-frontend/blob/f4693f1af03cecb562c8e4a0ad0af6f08a1e2f9e/src/supportedBrowsers.ts#L1

NoelLH (approved):

PR #1780, opened 2024-11-27
DON-1092 – fix campaign card grid input focus breaking Safari scroll positions

NoelLH:


src/app/explore/explore.component.ts
```diff
@@ -124,6 +130,16 @@ export class ExploreComponent implements OnDestroy, OnInit {
   }
 
   onScroll() {
+    if (this.scroller.getScrollPosition()[1] < this.smallestSignificantScrollPx) {
+      this.blurredSinceLastMajorScroll = false;
+      return;
```
NoelLH, line 135: This early return may stop some `more()` calls, but since we do the equivalent on meta-campaign already I can't see that causing major problems.

src/app/explore/explore.component.ts
```diff
@@ -124,6 +130,16 @@ export class ExploreComponent implements OnDestroy, OnInit {
   }
 
   onScroll() {
+    if (this.scroller.getScrollPosition()[1] < this.smallestSignificantScrollPx) {
```
bdsl, line 133: I had to look up what [1] means, so:
```suggestion
    const scrollPositionY = this.scroller.getScrollPosition()[1];
    if (scrollPositionY < this.smallestSignificantScrollPx) {
```

src/app/explore/explore.component.ts
```diff
@@ -19,15 +20,19 @@ import {HighlightCard} from "../highlight-cards/HighlightCard";
...
   currencyPipeDigitsInfo = currencyPipeDigitsInfo;
   loading = false; // Server render gets initial result set; set true when filters change.
   /** Whether any non-default search logic besides an order change has been applied. */
   searched = false;
 
+  private blurredSinceLastMajorScroll = false;
   private offset = 0;
   private routeParamSubscription: Subscription;
   private searchServiceSubscription: Subscription;
+  private smallestSignificantScrollPx = 250;
```
bdsl, line 35:
```suggestion
  private readonly smallestSignificantScrollPx = 250;
```

src/app/explore/explore.component.ts
```diff
@@ -124,6 +130,16 @@ export class ExploreComponent implements OnDestroy, OnInit {
   }
 
   onScroll() {
+    if (this.scroller.getScrollPosition()[1] < this.smallestSignificantScrollPx) {
+      this.blurredSinceLastMajorScroll = false;
```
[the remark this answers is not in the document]
NoelLH: @bdsl The idea is that if you go back to the top, you are more likely to be about to interact with the search fields. So there's a reasonable chance that it will be useful to blur afresh when you're done. I'll add a comment.

bdsl (approved):
LGTM

PR #1781, opened 2024-11-27
BG2-2745: Show welcome message in place of auto-login following new a…

bdsl:
…ccount registration

Adds copy as shown below to the registration page in case of navigation following successful account registration.

Copy is newly written for this PR so can use review.

image


dorota-joanna (approved):
LGTM

PR #1788, opened 2024-12-03
Skip home redirect in Regression

NoelLH:


bdsl (approved):
Right, I guess its an infinite loop because the homepage is used as 404 and regression doesn't have a CC page.

PR #1789, opened 2024-12-03
Fix known-mismatched countries for filtering

NoelLH:
OPT (Palestine) and ones with 'smart quotes'


bdsl (approved):
Looks good to me, I cross checked with the country list we have in SF and yes we do have the ascii apostrophe in each case in SF.

PR #1792, opened 2024-12-04
DON-518 – Merge meta-campaign component into explore

bdsl:
These two components look very similar and had lots of repeated code between them. They both allowed exploring a set of campaigns, the difference was that one was for exploring all BG campaigns in general and the other was specific to campaigns within any given meta-campaign.

I went through and manually copied eveyry line of code from meta-campaign to explore, adjust where necassary to deal with the fact that having a meta-campaign defined is now optional


src/app/explore/explore.component.ts
```diff
@@ -37,47 +77,177 @@ export class ExploreComponent implements OnDestroy, OnInit {
...
+        this.setFundSpecificProps(this.fund, this.metaCampaign);
+      }
+    }
 
     this.beneficiaryOptions = CampaignGroupsService.getBeneficiaryNames();
     this.categoryOptions = CampaignGroupsService.getCategoryNames();
     this.locationOptions = CampaignGroupsService.getCountries();
-
     this.queryParamsSubscription = this.scrollToSearchWhenParamsChange();
-    this.highlightCards = this.route.snapshot.data.highlights;
```
NoelLH, line 80: I think we've maybe accidentally lost assigning the highlight cards as a fallback in Explore mode? Seems like the router still has the resolver but we need to assign it here.

src/app/explore/explore.component.ts
```diff
@@ -152,41 +342,116 @@ export class ExploreComponent implements OnDestroy, OnInit {
...
+        this.shouldAutoScroll = true;
+      }
+
+      this.loading = false;
+
+      return;
+    }
+
+    // Else need to load children newly in browser.
+    // this.doCampaignSearch(query as SearchQuery, true); // Clear existing children
```
NoelLH, line 453: This line's `MetaCampaignComponent` was uncommented I believe. Do we need to keep it but make it conditional now?

NoelLH (commented):
Two things inline, and I'm not sure if it's intentional not to target the frozen branch during CC? Looks good otherwise.

src/app/explore/explore.component.ts
bdsl: Thanks, fixed in pairing with @dorota-joanna yesterday.

src/app/explore/explore.component.ts
bdsl: Thanks, fixed in pairing with @dorota-joanna yesterday.

PR #1793, opened 2024-12-04
Don 998 show donation form based on campaign type

dorota-joanna:
When a logged in user tries accessing non-regular-giving campaign at /regular-giving/[normal-campaign-id], they will see this message (copy is just an example):

They can click on the normal campaign link to make a one-off donation.

Also added the message on a regular donation page:

But doesn't work as expected yet (had to use another property for the screenshot), because when I changed the campaign to isRegularGiving in SF, it shows as undefined in frontend;

To test it locally, I had to remove the entire CC24 redirect from [here](https://github.com/thebiggive/donate-frontend/blob/3daa03f3e9a6acb42d65f9107b55119f63d45124/src/app/home/home.component.ts#L75C1-L86)


src/app/regular-giving/regular-giving.component.ts
```diff
@@ -62,6 +62,13 @@ export class RegularGivingComponent implements OnInit {
     this.donor = donor
 
     this.campaign = this.route.snapshot.data.campaign;
+
+    if ( !this.campaign.isRegularGiving ) {
+      this.router.navigate(['/my-account']);
+      throw new Error("Campaign " + this.campaign.id + " is not a regular giving campaign");
+      return;
```
bdsl, line 69: What's the reason for navigating to my-account? I think it would make sense to either display an error on the page here, or maybe navigate to the site homepage like we do on what would be 404 errors on other sites.

bdsl (commented):
Added a comment inline. It's probably also worth adding something to the regular donation start page to make that one error out if the campaign *is* of the regular giving type.

src/app/regular-giving/regular-giving.component.ts
dorota-joanna: Thank you @bdsl , yes I wasn't sure either my-account or the regular campaign page or homepage, picked my-account just to make sure the redirect works and that's easy to change. But you're right we should probably display error message informing the user that this is not a regular giving campaign, but encourage to make one off donation and add a link to this campaign page?

bdsl: Thanks. Let's either keep this or the code you've put in the template, we don't need both, do we? I think the template change is fine so we could delete this code.

Also I forgot to say before it doesn't make sense to have `throw` and then `return` on the last line - any statement directly after throw is dead code.

dorota-joanna: Thank you @bdsl , I think I left the `return` by mistake trying different things to stop the non-regular campaign from rendering in ngOninit. I'll remove this and navigation, but do you think we should keep the exception to show error in the console?

bdsl: We can but I don't think we need it when the error is shown on the page. Can we take out the `this.router.navigate(['/my-account']);` anyway?

src/environments/environment.ts
```diff
@@ -12,6 +12,7 @@ export const environment: Environment = {
   production: false,
   productionLike: false,
   creditTipsCampaign: 'a056900002LDXWgAAP',
+  //apiUriPrefix: 'https://thebiggive--dorota.sandbox.my.salesforce-sites.com',
```
bdsl, line 15:
```suggestion
```

bdsl (commented):
This still needs changing https://github.com/thebiggive/donate-frontend/pull/1793/files#r1869661375

and I added a suggestion to delete commented out code. Everything else LGTM.

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -59,9 +59,20 @@ export class RegularGivingComponent implements OnInit {
...
+    this.donor = donor;
+
+    if ( !this.campaign.isRegularGiving ) {
+      throw new Error("Campaign " + this.campaign.id + " is not a regular giving campaign");
+    }
 
     this.campaign = this.route.snapshot.data.campaign;
+
+    if ( !this.campaign.isRegularGiving ) {
+      this.router.navigate(['/my-account']);
```
dorota-joanna, line 71:
```suggestion
```

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -59,9 +59,19 @@ export class RegularGivingComponent implements OnInit {
...
+
+    if ( !this.campaign.isRegularGiving ) {
+      throw new Error("Campaign " + this.campaign.id + " is not a regular giving campaign");
+    }
 
     this.campaign = this.route.snapshot.data.campaign;
+
+    if ( !this.campaign.isRegularGiving ) {
+      throw new Error("Campaign " + this.campaign.id + " is not a regular giving campaign");
+      return;
```
dorota-joanna, line 72:
```suggestion
```

bdsl (approved):

dorota-joanna:
> This still needs changing https://github.com/thebiggive/donate-frontend/pull/1793/files#r1869661375
> 
> and I added a suggestion to delete commented out code. Everything else LGTM.

Thank you @bdsl , I've changed it but didn't commit the file :see_no_evil:

> Added a comment inline. It's probably also worth adding something to the regular donation start page to make that one error out if the campaign _is_ of the regular giving type.

PR #1795, opened 2024-12-04
Enable no-floating-promisis eslint rule and fix issues it finds

bdsl:
See description of rule at https://typescript-eslint.io/rules/no-floating-promises/


src/app/charity/charity.component.spec.ts
```diff
@@ -9,7 +9,7 @@ describe('CharityComponent', () => {
   let component: CharityComponent;
   let fixture: ComponentFixture<CharityComponent>;
 
-  beforeEach(waitForAsync(() => {
+  beforeEach(() => {
```
bdsl, line 12: see https://dev.to/this-is-angular/simplifying-angular-testbed-setup-by-converting-to-synchronous-test-hooks-19cl

NoelLH: Ah, nice!

NoelLH (commented):
Not reviewed in detail but the approach all LGTM so far

We could add new-ish ticket 'DON-1098' to the title and consider that 'in dev' probably

bdsl:
> We could add new-ish ticket _DON-1098_ to the title and consider that 'in dev' probably

Added to the title. If this is merged then that ticket is effectively done.

I did this yesterday because with the CC going on I thought it made more sense than to ask people to collaborate on any more product focused tickets that need discussion. I still feel like things like this that are so internal to the code shouldn't need tickets, and e.g. there's no sense in doing QA for this IMHO.

dorota-joanna (approved):
LGTM

PR #1798, opened 2024-12-06
Don 998 display mandate thanks

dorota-joanna:

### DIFF
--- a/src/app/mandate/mandate.component.html
+++ b/src/app/mandate/mandate.component.html
@@ -1,127 +1,114 @@
 <main class="b-container">
-  @if (complete) {
-    <div id="banner-thank-you"></div>
-  }
-    <div>
-      @if (false) {
-        <p class="error" aria-live="assertive">
-          Your donation was not processed immediately &ndash; it may be that there is a short delay because we're very
-          busy. Please refresh this page to see the latest status. We'll also email your receipt within a few minutes when
-          your donation is processed.
-        </p>
-      }
-      <div aria-live="polite">
-        @if (complete) {
-          <div>
-            <div class="screenreader-only">
-              Thank you! Your generous regular donation has been set up.
-            </div>
-          </div>
-        }
-      </div>
-      @if (complete) {
-        <div>
-          <div class="thank-you">
-            <h2>Thank you!</h2>
-            <p class="thank-you-text">Your generous regular donation has been set up.</p>
-          </div>
-          <hr aria-hidden="true">
-          <div class="donation-summary">
-            <p><strong>Your donation summary</strong></p>
-            <p>Your donation of
-              <strong class="figure">{{ mandate.donationAmount.amountInPence + ' ' + mandate.donationAmount.currency }}</strong>
-              is worth
-              <strong class="figure">{{ mandate.donationAmount.amountInPence + mandate.matchedAmount.amountInPence + ' ' + mandate.matchedAmount.currency  + ' for the next ' + mandate.numberOfMatchedDonations + ' months!'}}</strong>
-            </p>
-            <br/>
-<!--            @if (mandate?.thankYouMessage) {-->
-<!--              <p id="charity-thank-you">{{ mandate?.thankYouMessage }}</p>-->
-<!--            }-->
-          </div>
-          <div>
-            <div class="receipt">
-              <p><strong>Your Regular Donation Schedule</strong></p>
-              <!-- Date time appears as e.g. "Sep 5, 2023" -->
-              <table>
-                <tr>
-                  <td>Active from</td>
-                  <td>{{ mandate.schedule.activeFrom | date: 'mediumDate'}}</td>
-                </tr>
-                <tr>
-                  <td>Day of month</td>
-                  <td>{{ mandate.schedule.dayOfMonth }}</td>
-                </tr>
-                <td></td>
-                <tr class="donationAmount">
-                  <td>Your donation</td>
-                  <td>{{ mandate.donationAmount.amountInPence }}  {{ mandate.donationAmount.currency }}</td>
-                </tr>
-                  <tr>
-                    <td>Matched Amount</td>
-                    <td>{{ mandate.matchedAmount.amountInPence }} {{ mandate.matchedAmount.currency }}</td>
-                  </tr>
-<!--  TODO: display gift Aid and tip amount on the template (format copied from one-off cmpaign thank you page -->
-<!--                @if (giftAidAmount > 0) {-->
-<!--                  <tr>-->
-<!--                    <td>Gift Aid Amount</td>-->
-<!--                    <td>{{ giftAidAmount | exactCurrency:donation.currencyCode }}</td>-->
-<!--                  </tr>-->
-<!--                }-->
-<!--                @if (donation.tipAmount > 0) {-->
-<!--                  <tr>-->
-<!--                    <td>Tip Amount</td>-->
-<!--                    <td>{{ donation.tipAmount | exactCurrency:donation.currencyCode }}</td>-->
-<!--                  </tr>-->
-<!--                }-->
-                <tr>
-                  <td colspan="2"><hr aria-hidden="true"></td>
-                </tr>
-                <tr class="receives">
-                  <td>Charity Receives</td>
-                  <td>{{ mandate.donationAmount.amountInPence + mandate.matchedAmount.amountInPence + ' ' + mandate.matchedAmount.currency }}</td>
-                </tr>
-                <tr class="total">
-                  <td><strong>Total amount you have been charged</strong></td>
-                  <td><strong>{{ mandate.donationAmount.amountInPence | exactCurrency:donation.currencyCode }}</strong></td>
-                </tr>
-              </table>
-            </div>
-            <hr class="shorter" aria-hidden="true">
-          </div>
-          <div class="socials">
-            <p>Boost your impact by sharing on social media</p>
-            <div class="social-icons-wrapper">
-              <span class="separator"></span>
-              <biggive-social-icon
-                service="Facebook"
-                label-prefix="Share"
-                url="https://www.facebook.com/sharer/sharer.php?u={{ encodedShareUrl }}" background-colour="tertiary"
-                icon-colour="black"
-                wide="true"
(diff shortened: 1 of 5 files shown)


src/app/mandate/mandate.component.html
```diff
@@ -0,0 +1,146 @@
...
+<!--            }-->
+          </div>
+          <div>
+            <div class="receipt">
+              <p><strong>Your Regular Donation Schedule</strong></p>
+              <!-- Date time appears as e.g. "Sep 5, 2023, 2:54 PM" -->
+              <table>
+                <tr>
+                  <td>Active from</td>
+                  <td>{{ mandate.schedule.activeFrom | date: 'shortTime'}} {{ mandate.schedule.activeFrom | date: 'mediumDate'}}</td>
```
bdsl, line 66: think we probably don't need to show the time, the date should be enough.

src/app/mandate/mandate.component.html
```diff
@@ -0,0 +1,146 @@
...
+                </tr>
+                <tr>
+                  <td>Day of month</td>
+                  <td>{{ mandate.schedule.dayOfMonth | date: 'shortTime'}} {{ mandate.schedule.dayOfMonth | date: 'mediumDate'}}</td>
```
bdsl, line 70: day of month is just a number from 1 to 28, you're showing it as a date in 1970 here which is not right of course.

src/app/mandate/mandate.component.html
```diff
@@ -0,0 +1,146 @@
...
+            <p class="thank-you-text">Your generous regular donation has been set up.</p>
+          </div>
+          <hr aria-hidden="true">
+          <div class="donation-summary">
+<!--            <p><strong>Hi {{ donation.firstName }},</strong></p>-->
+            <p><strong>Your donation summary</strong></p>
+            <p>Your donation of
+              <strong class="figure">{{ mandate.amount.amountInPence + ' ' + mandate.amount.currency }}</strong>
+              is worth
+              <strong class="figure">{{ totalValue + ' ' + mandate.amount.currency }}
```
bdsl, line 51: I don't think we can say it's worth double each month - its only going to be doubled for the first three months. Not sure what the best way to express that is.

And if possible IMHO we should avoid hard-coding the fact that it will be doubled for the first three months in the template and make that info somehow part of the mandate object that we will load from matchbot. Not sure exactly how we'd write it or how we'd put it in the object.

If we can work out a wording maybe hard code it in the template but with a very obvious todo comment that's visible in the browser so we won't forget to replace with actual data.

src/app/mandate/mandate.component.html
```diff
@@ -0,0 +1,146 @@
...
+                <td></td>
+                <tr class="donationAmount">
+                  <td>Your donation</td>
+                  <td>{{ mandate.amount.amountInPence }}  {{ mandate.amount.currency }}</td>
+                </tr>
+<!--                @if (donation.matchedAmount > 0) {-->
+                  <tr>
+                    <td>Matched Amount</td>
+                    <td>{{ mandate.amount.amountInPence }} {{ mandate.amount.currency }}</td>
+                  </tr>
```
bdsl, line 81: this isn't called matched amount in the mandate object, I don't think we should present it as matched amount here for same reason as above - its only going to be matched for some of the donations not all.

Actually it might be more accurate to say the donation amount is 10 GBP per month, and the matched amount is 30 GBP. But that matched amount figure should come from the mandate object. Fine to add another property to the mandate here and we'll put it in the matchbot api response later.

dorota-joanna: Thank you @bdsl , I think my last commit should resolve all your suggestions!

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -61,14 +61,10 @@ export class RegularGivingComponent implements OnInit {
...
 
-    if ( !this.campaign.isRegularGiving ) {
-      throw new Error("Campaign " + this.campaign.id + " is not a regular giving campaign");
-    }
-
     this.campaign = this.route.snapshot.data.campaign;
 
     if ( !this.campaign.isRegularGiving ) {
-      throw new Error("Campaign " + this.campaign.id + " is not a regular giving campaign");
+     // throw new Error("Campaign " + this.campaign.id + " is not a regular giving campaign");
```
bdsl, line 67: this shouldn't be left commented out, should it?

src/app/mandate/mandate.component.html
bdsl: I don't think it does - this table row still has a heading that doesn't match what's displayed. If the heading is "Matched Amount" then IMHO we should be displaying a field in the mandate called "matchedAmount". It might be the same number as `amount` but if so I think that's something that should be programmed into matchbot, not frontend.

I think we should aim for making it so that as a user if you happen to know JSON you learn the same things by either reading this page as normal or directly reading the Mandate as a JSON object from matchbot - frontend shouldn't be doing business logic to add extra information.


We could leave the matched amount off this page for now and have that as another todo on the ticket or a new ticket - not sure. Or add it as another field on the mandate object as part of this.

src/app/mandate/mandate.component.html
```diff
@@ -0,0 +1,145 @@
...
+                <tr>
+                  <td>Day of month</td>
+                  <td>{{ mandate.schedule.dayOfMonth }}</td>
+                </tr>
+                <td></td>
+                <tr class="donationAmount">
+                  <td>Your donation</td>
+                  <td>{{ mandate.amount.amountInPence }}  {{ mandate.amount.currency }}</td>
+                </tr>
+<!--                @if (donation.matchedAmount > 0) {-->
```
bdsl, line 76:
```suggestion
```

src/app/mandate/mandate.component.html
```diff
@@ -0,0 +1,145 @@
+<main class="b-container">
+  @if (complete) {
+    <div id="banner-thank-you"></div>
+  }
+
+<!--  @if (!complete && !noAccess && !timedOut) {-->
+<!--    <mat-spinner class="complete-spinner" color="primary"-->
+<!--                 diameter="30" aria-label="Loading donation details"></mat-spinner>-->
```
bdsl, line 8: can we remove all the stuff referencing `donation` before merging?

src/app/mandate/mandate.component.html
```diff
@@ -0,0 +1,145 @@
...
+<!--  }-->
+
+<!--  @if (isDataLoaded) {-->
+    <div>
+      @if (false) {
+        <p class="error">
+          Sorry, you don't have access to this page. To keep your information secure we only allow viewing this
+          page temporarily after making a donation. To see donation details please check your email for a donation
```
bdsl, line 16: this is talking about a donation, not relevant for a mandate

src/app/mandate/mandate.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  ],
+  templateUrl: './mandate.component.html',
+  styleUrl: './mandate.component.scss'
+})
+export class MandateComponent {
+  personId = 'donor-id';
+  complete = true;
+  encodedShareUrl = '';
+  encodedPrefilledText: string = '';
+  donation: any = {};
```
bdsl, line 27: we filled this in quickly to get the template working as it started, as a copy of the donate thanks template, but I think it needs more work to get it ready to merge.

E.g. this line is donation related not mandate, and we should avoid the `any` type.

I think quite a few other properties here may be not relevant to mandates.

src/app/mandate/mandate.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  totalValue: number = 20;
+  giftAidAmount: number = 0;
+  totalPaid: 0;
+  mandate: Mandate = {
```
bdsl, line 31: I think it's fine to keep this as a hard-coded dummy for merging this, but lets just add a todo comment.

```suggestion
  /** @todo-regular-giving : replace this with a mandate fetched from matchbot *//
  mandate: Mandate = {
```

src/app/mandate/mandate.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  id: 'f7037101-b555-4482-afff-43145fac78bb',
+  campaignId: 'a056900002TPVz5AAH',
+  charityName: '0011r00002Hoe8lAAB',
+  matchingPeriodText: 'for the next three months',
```
bdsl, line 35: This text is written as half a sentence, which means we'd end up with half of a piece of microcopy in matchbot and half in frontend ("Your donation of
              {amount} is worth {doubleAmount} for the next three months").
              
I don't think we should do that, it's going to make it complicated to edit that copy having half in one place and half in another. Even more so since the two apps get deployed separately.

Not sure what the best solution is, but maybe this?

```suggestion
  numberOfDonationsMatched: 3,
```

Also since this page can be viewed any time it won't always be the next three months, so maybe we should change the copy to say the first three donations will be matched instead of donations will be matched for the next three months?

src/app/regular-giving/regular-giving.component.html
bdsl, line 1: did you meant to commit changes in this file?

src/app/mandate/mandate.component.html
dorota-joanna: @bdsl , not sure if we want to keep the spinner part when data is loading?

bdsl: I don't think we need to, if we load it in a resolver then it will be loaded before the template renders anyway.

src/app/mandate/mandate.component.ts
dorota-joanna: changed in last commit!

src/app/mandate/mandate.component.html
dorota-joanna: thank you @bdsl , yes now updated!

src/app/mandate/mandate.component.ts
dorota-joanna: updated!

src/app/mandate/mandate.component.html
dorota-joanna: removed!

bdsl (approved):

PR #1800, opened 2024-12-11
DON-1066: Link to regular giving agreement from card payment

bdsl:
Screenshot to show how this would look with one regular donation and one ad-hoc donation side by side:

image

### DIFF
--- a/src/app/my-donations/my-donations.component.scss
+++ b/src/app/my-donations/my-donations.component.scss
@@ -27,10 +27,6 @@ div.donation {
   height: 100%;
   justify-content: space-between;
   flex-direction: column;
-  a.charityName:link, a.charityName:visited, a.charityName:hover {
-    color: inherit !important;
-    text-decoration: none;
-  }
   *:first-child {
     margin-top: 0;
     flex-grow: 1;
--- a/src/app/my-donations/my-donations.component.scss
+++ b/src/app/my-donations/my-donations.component.scss
@@ -6,7 +6,10 @@
 main > div {
   color: white;
   background-color: $colour-primary;
+}
 
+p:not(biggive-container-card *) {
+  // biggive-container-cards have white backgrounds so white links in there would be invisible.
   a:link, a:visited, a:hover {
     color: white;
   }
@@ -24,7 +27,7 @@ div.donation {
   height: 100%;
   justify-content: space-between;
   flex-direction: column;
-  a:link, a:visited, a:hover {
+  a.charityName:link, a.charityName:visited, a.charityName:hover {
     color: inherit !important;
     text-decoration: none;
   }
--- a/src/app/donation.model.ts
+++ b/src/app/donation.model.ts
@@ -22,6 +22,16 @@ export const maximumDonationAmountForFundedDonation = 200_000;
  * * after the donation is fully processed and webhook returned (e.g. `matchedAmount`).
  */
 export interface Donation {
+    /**
+     * The regular giving agreement relating to this donation, if any. Optional property because production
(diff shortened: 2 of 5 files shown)


src/app/my-donations/my-donations.component.scss
```diff
@@ -24,7 +27,7 @@ div.donation {
   height: 100%;
   justify-content: space-between;
   flex-direction: column;
-  a:link, a:visited, a:hover {
+  a.charityName:link, a.charityName:visited, a.charityName:hover {
```
NoelLH, line 30: I'm not sure about the mix of link styles here, and the card not _all_ being linked like the home highlight cards. Is there a way donors would figure out that a bold charity name is a mandate link based on other styling around the app?

NoelLH (commented):
I think link scope, style and related implementation CSS is maybe getting a bit confused now. Everything else LGTM.

src/app/my-donations/my-donations.component.scss
bdsl: The charity name isn't a link to the mandate, it's a link to the campaign. The link to the mandate is blue and not bold.

I'm also not sure about mixing the styles though. The campaign link style is unchanged. I can't remember for sure why we didn't make that one blue but I guess we thought it was not very important and having it blue might make the page too busy.

The mandate links are probably more vital for people to be able to find easily.

NoelLH: Right, sorry for typo. I think I did see what it linked to but still feel it's better to either re-style the campaign link somehow or not link at all.

I agree the mandate one is the priority and is good as-is. I don't think that precludes the rest of the card effectively becoming a big campaign link if we can make that work, since that mostly mirrors an interface pattern we have already used on the home page.

bdsl: @NoelLH do you have a view on a way forward for this? I looked on the comment on DON-983 to see if we discussed the links being black rather than blue. There are a lot of comments there so I may have missed it, but I can't see that we did.

I think having the entire card linked probably doesn't make sense unless it was a link to the donation, which we aren't currently doing. For now we link the campaign and the mandate. I think we may well want to add a link to the donation as a its own page (which may look exactly the same as our current donation thank you page) in future.

bdsl: @NoelLH do you think we should make the link to the campaign blue? Would certainly make it more obvious. 

@DomTBG and @BethBigGive you were both quite involved in working out the design of this page back in July, so you might have views about changes we do now. So far I've in this PR I've added the link to the mandate but not changed any of the existing stuff.

bdsl: Would look like this if we just deleted these lines to let the charity name use standard link styling:

image

DomTBG: Thanks @bdsl from a design perspective, there is a lot of white space on the right hand side that it would be good to use / remove if possible, maybe through more padding on the left

bdsl: @DomTBG The width is going to be quite dependant on the number of donations you have and the width of your screen, and if we adjust it we'll need to test with a range of screen sizes again so I don't know if we need to get into that now - it doesn't seem like an issue that comes from adding the link to the regular giving mandate.

Do you have a view about the link styling for the link to the mandate and the link to the charity?

DomTBG: OK @bdsl let's leave the white space question for now.
I don't know that we need to link to the campaign for any donations - it's not like we link to the campaign in the donation receipt.

DomTBG: we are considering removing all of that info on the donation receipt though @bdsl

PR #1801, opened 2024-12-13
DON-998: add resolver for an single mandate

dorota-joanna:


src/app/app-routing.ts
```diff
@@ -73,6 +74,17 @@ const LoggedInPersonResolver: ResolveFn<Person | null> = async () => {
...
 }
 
+const mandateResolver: ResolveFn<Mandate> = async (route: ActivatedRouteSnapshot) => {
+  const mandateService = inject(MandateService);
+  const mandateId = route.paramMap.get('mandateId');
+  console.log({mandateId});
+  if (!mandateId) {
+    throw new Error('mandateId param missing in route');
+  }
+  const mandate$ = mandateService.getActiveMandate(mandateId);
```
bdsl, line 84:
```suggestion
```

bdsl (dismissed):
LGTM - will be broken right now as we need an update to the mandate API model in matchbot but that's OK, we're not relying on mandate stuff working just yet.

PR #1804, opened 2024-12-17
Don 1065 logout in menu

bdsl:
image


src/app/transfer-funds/transfer-funds.component.ts
```diff
@@ -335,20 +335,6 @@ export class TransferFundsComponent implements AfterContentInit, OnInit {
     return Math.floor(creditAmount * (tipPercentage / 100));
   }
 
-  logout() {
```
NoelLH, line 338: 🎉

NoelLH (dismissed):
LGTM. Seems like the tests are passing now too

NoelLH (approved):

PR #1805, opened 2024-12-18
Don 1065 account page redesign

bdsl:
image

image


bdsl:
Not sure yet why I'm getting this test failure:

> Expected 'Your Payment MethodsDonation Funds Saved Cards ' to contain 'ACMECARD Card Ending: 1234  Expiry Date: > 05/2030'.

src/app/my-payment-methods/my-payment-methods.component.ts
```diff
@@ -0,0 +1,157 @@
...
+      return;
+    }
+
+    this.paymentMethods = await this.donationService.getPaymentMethods({cacheBust: true})
+  };
+
+  async loadPaymentMethods() {
+    // not so keen on the component using the donation service and the identity service together like this
+    // would rather call one service and have it do everything for us. Not sure what service would be best to put
+    // this code in.
```
NoelLH, line 70:
```suggestion
```

Looks like this comment has come along to the new component but is about an old version of the code below?

src/app/my-payment-methods/my-payment-methods.component.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  deleteMethod(method: PaymentMethod) {
+    this.paymentMethods = undefined;
+
+    this.donationService.deleteStripePaymentMethod(this.person, method, this.jwtAsString()).subscribe(
+      this.loadPaymentMethods.bind(this),
+      error => {
+        this.loadPaymentMethods.bind(this)()
+        alert(error.error.error)
```
NoelLH, line 81: Not new code, but would ideally probably be better to use a toast or print error somewhere. Similarly, I think `registerErrorDescription` below may not be rendered anywhere?

NoelLH (approved):
Looks good! Minor things inline, no blockers

src/app/my-payment-methods/my-payment-methods.component.ts
bdsl: Thanks, deleted in 4d6b1d0f4

PR #1806, opened 2024-12-18
DON-998: view mandate link

dorota-joanna:

### DIFF
--- a/src/app/my-regular-giving/my-regular-giving.component.html
+++ b/src/app/my-regular-giving/my-regular-giving.component.html
@@ -35,6 +35,14 @@
                 </ul>
                 </div>
 
+                <biggive-button
+                  space-above="5"
+                  colour-scheme="primary"
+                  label="View Mandate"
+                  full-width="true"
+                  size="medium"
+                  url="/my-account/regular-giving/{{mandate.id}}"
+                />
               </biggive-container-card>
             }
           </biggive-grid>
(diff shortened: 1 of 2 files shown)


bdsl (dismissed):
Button LGTM. I don't know really if we want a button here or a regular link but IMHO that can be dealt with in QA.

Do you want to also replace the `alert` with a redirect to the new mandate page?
https://github.com/thebiggive/donate-frontend/blob/aae2377d086b4259eb81a5ea2184d340e71e9097/src/app/regular-giving/regular-giving.component.ts#L125

dorota-joanna:
> alert(prefix + JSON.stringify(mandate

> Button LGTM. I don't know really if we want a button here or a regular link but IMHO that can be dealt with in QA.
> 
> Do you want to also replace the `alert` with a redirect to the new mandate page?
> 
> https://github.com/thebiggive/donate-frontend/blob/aae2377d086b4259eb81a5ea2184d340e71e9097/src/app/regular-giving/regular-giving.component.ts#L125

@bdsl do we need the redirect, when the button is assigned url ?

bdsl:
> > alert(prefix + JSON.stringify(mandate
> 
> > Button LGTM. I don't know really if we want a button here or a regular link but IMHO that can be dealt with in QA.
> > Do you want to also replace the `alert` with a redirect to the new mandate page?
> > https://github.com/thebiggive/donate-frontend/blob/aae2377d086b4259eb81a5ea2184d340e71e9097/src/app/regular-giving/regular-giving.component.ts#L125
> 
> @bdsl do we need the redirect, when the button is assigned url ? (updated in my last commit, tested it still works, lmk if this is what you had in mind!

Yep, 501d865516398a is what I had in mind. Thanks!

bdsl (approved):

PR #1810, opened 2024-12-23
Be consistent about SSR `canActivate` route guards

NoelLH:
In all cases where Identity local tokens are needed, pass the decision to the browser

### DIFF
--- a/src/app/app-routing.ts
+++ b/src/app/app-routing.ts
@@ -9,7 +9,7 @@ import {isAllowableRedirectPath, LoginComponent} from "./login/login.component";
 import {inject, PLATFORM_ID} from "@angular/core";
 import {IdentityService} from "./identity.service";
 import {RegisterComponent} from "./register/register.component";
-import {isPlatformBrowser} from "@angular/common";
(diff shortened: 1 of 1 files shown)


src/app/app-routing.ts
```diff
@@ -235,15 +251,12 @@ const routes: Routes = [
...
-    component: LoginComponent, // Angular requires we set a component but it will never be used as `canActivate` always
-                               // redirects.
+    component: LoginComponent, // Angular requires we set a component but it will never be used client-side as      
+                                    // `canActivate` always redirects there.
     path: 'logout',
     pathMatch: 'full',
     canActivate: [
-      async () => {
-        inject(IdentityService).logout();
-        await inject(Router).navigate(['/'], {replaceUrl: true});
```
NoelLH, line 245: I replaced direct navigation with a URL tree return, having read at https://github.com/angular/angular/issues/56691#issuecomment-2191394819 that
> Navigating inside a guard is not recommended. you should return an URL tree of the boolean.

dorota-joanna (commented):
LGTM

PR #1811, opened 2025-01-02
DON-1066: Add "Regular Giving" badget for past donation

bdsl:
image


NoelLH (approved):
LGTM. I think it might be nice to later support both badge styles inside `components` where they could share a mixin, but for Regular Giving pilot and certainly prior to wider discussion of the final design, it makes sense to do it like this.

PR #1812, opened 2025-01-03
DON-998: Take donor billing country & postcode as part of mandate cre…

bdsl:
…ation

This change also includes displaying their existing details on the mandate creation form. We will soon want to also display it on the account page I think, but as there's no way to set it except on the mandate creation page it's ok for now to only display it here.

image


dorota-joanna (approved):
LGTM

PR #1816, opened 2025-01-07
DON-998: Display human-readable error message on regular giving manda…

bdsl:
…te setup failure

### DIFF
--- a/src/app/regular-giving/regular-giving.component.ts
+++ b/src/app/regular-giving/regular-giving.component.ts
@@ -170,9 +170,9 @@ export class RegularGivingComponent implements OnInit {
     next: async (mandate: Mandate) => {
       await this.router.navigateByUrl(`${myRegularGivingPath}/${mandate.id}`);
     },
-      error: (error: Error) => {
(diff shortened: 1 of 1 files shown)


src/app/regular-giving/regular-giving.component.ts
```diff
@@ -170,9 +170,9 @@ export class RegularGivingComponent implements OnInit {
     next: async (mandate: Mandate) => {
       await this.router.navigateByUrl(`${myRegularGivingPath}/${mandate.id}`);
     },
-      error: (error: Error) => {
+      error: (error: {error: {error: {description?: string} }}) => {
       console.error(error);
-      const message = error.message
+      const message = error.error.error.description ?? 'Sorry, something went wrong';
```
bdsl, line 175: Probably worth refactoring soon to deal with this repeated triple error - I might do tomorrow. Maybe we can simplify the structure sent from matchbot, if not we can just dedupe some of the code to deal with it here.

image

NoelLH (approved):

PR #1818, opened 2025-01-08
DON-998: Send Stripe confirmation token when setting up regular g…

bdsl:
…iving mandate

With related changes in matchbot this will allow the first donation to be confirmed & collected, and the payment method to be saved on the Donor's file for use with the later donations related to this and any other regular giving mandate.

### DIFF
--- a/src/app/regular-giving/regular-giving.component.html
+++ b/src/app/regular-giving/regular-giving.component.html
@@ -1,11 +1,12 @@
 <main class="b-container">
-  @if(!campaign.isRegularGiving) {
-      <div>
-          <p class="error">
-            Sorry, campaign {{ campaign.title }} is not a regular giving campaign.</p>
-          <p>To make a one-off donation to {{ campaign.charity.name }} please visit the regular <a href="/campaign/{{campaign.id}}">campaign page</a>.
-          </p>
-      </div>
+  @if (!campaign.isRegularGiving) {
+    <div>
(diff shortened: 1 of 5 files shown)


src/app/regular-giving/regular-giving.component.html
```diff
@@ -1,11 +1,12 @@
...
-          <p class="error">
-            Sorry, campaign {{ campaign.title }} is not a regular giving campaign.</p>
-          <p>To make a one-off donation to {{ campaign.charity.name }} please visit the regular <a href="/campaign/{{campaign.id}}">campaign page</a>.
-          </p>
-      </div>
+  @if (!campaign.isRegularGiving) {
+    <div>
+      <p class="error">
+        Sorry, campaign {{ campaign.title }} is not a regular giving campaign.</p>
+      <p>To make a one-off donation to {{ campaign.charity.name }} please visit the regular <a
```
NoelLH, line 6:
```suggestion
      <p>To make a one-off donation to {{ campaign.charity.name }} please visit the <a
```

Maybe best to avoid using 'regular' two different ways in the same page?

NoelLH (approved):
LGTM!

src/app/regular-giving/regular-giving.component.html
bdsl: Yes that's a good catch, thanks.

PR #1819, opened 2025-01-10
Meta campaign page for regular giving

bdsl:
Only a few adjustments to the existing meta-campaign page for the regular giving case.  Specifically removing the ticker, removing the match Funds Remaining indication on each card, changing "Donate now" to "Donate monthly", and changing the "find out more" link target.

This is how it looks with our test campaign. (I added a banner from unsplash & a definition from OED):

image

I think as it doesn't have those changes we'll also need to either exclude regular giving campaigns from the general explore page, or (IMHO less likely what we want) allow that page to show a mix of regular giving and ad-hoc giving campaigns.

### DIFF
--- a/src/app/explore/explore.component.html
+++ b/src/app/explore/explore.component.html
@@ -1,5 +1,5 @@
 <main>
-  @if (tickerMainMessage || tickerItems.length > 0) {
+  @if ((tickerMainMessage || tickerItems.length > 0) && ! metaCampaign?.isRegularGiving) {
     <biggive-totalizer
       primary-colour="primary"
       primary-text-colour="white"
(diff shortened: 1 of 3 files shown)


src/app/explore/explore.component.html
```diff
@@ -67,26 +67,52 @@
...
+            metacampaign page -->
+            @for (campaign of individualCampaigns; track campaign.id) {
+              <div>
+                <biggive-campaign-card
+                  [banner]="campaign.imageUri | optimisedImage:830 | async"
+                  [campaignType]="campaign.isMatched ? 'Match Funded' : null"
+                  [campaignTitle]="campaign.title"
+                  [donateButtonUrl]="'/regular-giving/' + campaign.id"
+                  [donateButtonLabel]="'Donate monthly'"
+                  [moreInfoButtonUrl]="'/charity/' + campaign.charity.id"
```
bdsl, line 81: Linking to the charity page may not be ideal as that page doesn't actually show any details of the charity at the moment, beyond their name and the list of campaigns they have with BG, but it may be the best option right now as the campaign page does not currently support regular giving campaigns (which might be something to fix).

We could also link to the charity's website. That's not available in the `CampaignSummary` type in FE but I think it is actually available at runtime when the data is sent via our CampaignSearchServiceModel in SF.

PR #1821, opened 2025-01-10
DON-1112: Add user-friendly validation for amount errors on regular g…

bdsl:
…iving form

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -52,9 +52,9 @@ import {IdentityService} from '../../identity.service';
 import {ConversionTrackingService} from '../../conversionTracking.service';
 import {PageMetaService} from '../../page-meta.service';
 import {Person} from '../../person.model';
-import {PostcodeService} from '../../postcode.service';
+import {billingPostcodeRegExp, postcodeFormatHelpRegExp, postcodeRegExp, PostcodeService} from '../../postcode.service';
 import {retryStrategy} from '../../observable-retry';
-import {StripeService} from '../../stripe.service';
+import {getStripeFriendlyError, StripeService} from '../../stripe.service';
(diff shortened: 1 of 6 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1439,70 +1428,17 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
   ) {
     this.submitting = false;
-    this.stripeError = this.getStripeFriendlyError(error, context);
+    this.stripeError = getStripeFriendlyError(error, context);
     this.toast.showError(this.stripeError);
     this.stripeResponseErrorCode = error?.code;
 
     this.jumpToStep('Payment details');
     this.goToFirstVisibleError();
   }
```
bdsl, line 1441: function below moved to stripe.service.ts to allow sharing with new component. Not sure if that's the best place to move it to - not sure if its idiomatic or not to put extra stuff in the file alongside the service class in Angular.

src/app/stripe.service.ts
```diff
@@ -241,3 +241,53 @@ export class StripeService {
     return Math.round((donation.tipAmount + donation.donationAmount) * 100);
   }
 }
+
+
+export function getStripeFriendlyError(
+  error: StripeError | {message: string, code: string, decline_code?: string, description?: string} | undefined,
+  context: 'method_setup'| 'card_change'| 'confirm',
```
NoelLH, line 248: Liking this narrower type; assuming rest of this is moved only

NoelLH (approved):
Slightly guessed about which bits are moved as running out of time today but all LGTM. Stepper interception approach is interesting, might be worth considering mirroring in the main form if we find it works well in practice.

src/app/stripe.service.ts
bdsl: Yes it is just moved.

PR #1825, opened 2025-01-14
DON-998: Add copy to login & register pages to explain use of account…

bdsl:
… for regular giving

Also added a link from registration back to login

image

image

### DIFF
--- a/src/app/login/login.component.html
+++ b/src/app/login/login.component.html
@@ -81,7 +81,7 @@
         } @else {
           @if (redirectPath.startsWith('/regular-giving')) {
             <p>Please log-in or <a href="{{registerLink}}">register</a> for a Big Give donation account to set up your regular giving agreement.</p>
-            <p>You will be able to log-in in future to see your donations or cancel the agreement.</p>
+            <p>You will be able to log-in any time to see your donations or cancel the agreement.</p>
           }
           <form
             [formGroup]="loginForm"
--- a/src/app/login/login.component.html
+++ b/src/app/login/login.component.html
@@ -79,6 +79,10 @@
             </div>
           </div>
         } @else {
+          @if (redirectPath.startsWith('/regular-giving')) {
+            <p>Please log-in or <a href="{{registerLink}}">register</a> for a Big Give donation account to set up your regular giving agreement.</p>
+            <p>You will be able to log-in in future to see your donations or cancel the agreement.</p>
+          }
           <form
             [formGroup]="loginForm"
             (keydown.enter)="login()"
--- a/src/app/register/register.component.html
+++ b/src/app/register/register.component.html
@@ -14,7 +14,7 @@
       </div>
       @if (redirectPath.startsWith('/regular-giving')) {
         <p>Please register for a a Big Give donation account to set up your regular giving agreement.</p>
-        <p>You will be able to log-in any time to see your regular or cancel the agreement.</p>
+        <p>You will be able to log-in any time to see your donations or cancel the agreement.</p>
       } @else {
       <p>
         Registering for a Donor Account is quick and easy.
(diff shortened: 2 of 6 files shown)


src/app/login/login.component.html
```diff
@@ -79,6 +79,10 @@ <h2 mat-dialog-title>Reset Password</h2>
             </div>
           </div>
         } @else {
+          @if (redirectPath.startsWith('/regular-giving')) {
+            <p>Please log-in or <a href="{{registerLink}}">register</a> for a Big Give donation account to set up your regular giving agreement.</p>
```
NoelLH, line 83: Maybe `routerLink` for slightly smoother nav?

src/app/register/register.component.html
```diff
@@ -12,9 +12,14 @@
           text="Donor Registration"
         ></biggive-heading>
       </div>
+      @if (redirectPath.startsWith('regular-giving')) {
```
NoelLH, line 15: Should we standardise these so that everything either does or doesn't include a leading `/`? I think that for redirect purposes it didn't matter, but now that we're checking start of strings specifically it's a bit confusing to introduce variation between Login & Register and makes it harder to see at a glance whether both are going to work.

src/app/register/register.component.html
```diff
@@ -12,9 +12,14 @@
...
+        <p>Please register for a a Big Give donation account to set up your regular giving agreement.</p>
+        <p>You will be able to log-in in future to see your regular or cancel the agreement.</p>
```
NoelLH, line 17:
```suggestion
        <p>You will be able to log in any time to see or cancel the agreement.</p>
```
Maybe tweak roughly like this to avoid "in in"? Also fixes a typo.

src/app/register/register.component.html
```diff
@@ -105,6 +110,8 @@
           }
         </div>
       </div>
+      <hr style="margin: 20px auto;">
+      Already have an account? <a href="{{loginLink}}">Login here</a>
```
NoelLH, line 114: Think this could be `routerLink` as above.

NoelLH (commented):
I mostly like the added copy, added some small suggestions.

src/app/register/register.component.html
bdsl: Yes. Done in 0fb53570

src/app/register/register.component.html
bdsl: Yep, done in 90d31af0

src/app/register/register.component.ts
```diff
@@ -16,7 +16,7 @@ import {DomSanitizer, SafeHtml} from "@angular/platform-browser";
 import {transferFundsPath} from "../app-routing";
 import {WidgetInstance} from "friendly-challenge";
 import {flags} from "../featureFlags";
-import {isAllowableRedirectPath, LoginNavigationState} from "../login/login.component";
+import {isAllowableRedirectPath, LoginNavigationState, normaliseRedirectPath} from "../login/login.component";
```
NoelLH, line 19: Wondering if this helper might be better in `NavigationService` or similar, so we don't have a page depending upon another?

bdsl: Done in f981ea1d

src/app/register/register.component.html
```diff
@@ -12,9 +12,14 @@
           text="Donor Registration"
         ></biggive-heading>
       </div>
+      @if (redirectPath.startsWith('/regular-giving')) {
+        <p>Please register for a a Big Give donation account to set up your regular giving agreement.</p>
+        <p>You will be able to log-in any time to see your regular or cancel the agreement.</p>
```
NoelLH, line 17:
```suggestion
        <p>You will be able to log-in any time to see your donations or cancel the agreement.</p>
```
I missed that there was another 'your regular'. This change matches the one you committed above.

bdsl: Thanks!

NoelLH (approved):

PR #1827, opened 2025-01-17
BG2-2776 – update home cards & redirect logic

NoelLH:
* Add evergreen CC home redirect support
* Automatically determine whether menu needs ?noredirect on Explore component pages
* Remove temporary home card & redirect workarounds


src/app/app.component.ts
```diff
@@ -75,6 +76,10 @@ export class AppComponent implements AfterViewInit, OnDestroy, OnInit {
     private matomoTracker: MatomoTracker,
     private router: Router,
   ) {
+    this.isPlatformBrowser = isPlatformBrowser(this.platformId);
```
NoelLH, line 79: My editor was complaining about the old position of this. In the constructor seems like a normal place to do it.

src/app/donation-start/donation-start-container/donation-start-container.component.spec.ts
```diff
@@ -48,6 +48,7 @@ describe("DonationStartContainer", () => {
           siteId: "",
           trackerUrl: "",
         }),
+        RouterLink,
```
NoelLH, line 51: Added to remove a test output warning.

src/app/highlight-cards/HighlightCard.spec.ts
```diff
@@ -112,23 +112,6 @@ describe('highlightCard', () => {
     expect(highlightCardForHomepage.button.href.href).toBe('https://example-blog.com/some-path');
   });
 
-  it('should replace CC24 link with metacampaign page', () => {
```
NoelLH, line 115: This will be set correctly at source after https://github.com/thebiggive/salesforce/pull/2175 is out, no more rewrites.

src/app/home/home.component.ts
```diff
@@ -48,41 +48,22 @@ export class HomeComponent implements OnInit {
     this.stats = this.route.snapshot.data.stats;
     this.highlightCards = this.route.snapshot.data.highlights;
     const queryParams = this.route.snapshot.queryParams;
-
-    if (environment.environmentId !== 'production') {
-      if (queryParams?.simulatedDate) {
```
NoelLH, line 53: I think it makes sense to lose simulated dates now the logic is driven by factors we can edit in SF data at will.

src/app/navigation.service.ts
```diff
@@ -22,6 +33,29 @@ export class NavigationService {
...
+      // CC 'Donate Today' or 'Donate Now' cards mean there should be a full page redirect (unless donor clicked
+      // the logo specifically to avoid that).
+      if (card.campaignFamily === 'christmasChallenge' && card.button.text.startsWith('Donate ')) {
+        redirectPath = card.button.href.pathname;
+      }
+    });
+
+    // At actual runtime in browser we expect this signal to be ready before redirect paths are checked because
+    // `setPossibleRedirectSignal()` is called in AppComponent's constructor, but this isn't guaranteed in all
+    // test contexts for now.
```
NoelLH, line 51: IMO this is OK for the time being. The more substantial logic is unit tested in `NavigationService`; the signal use is simple and has been + will be further manually tested.

src/app/app.component.ts
bdsl: Right, phpstorm doesn't complain but I think vscode rightly says "Property 'platformId' is used before its initialization.ts(2729)".

src/app/home/home.component.ts
bdsl: I think so, the simulated date was only used to control the redirect. If we want it it probably needs to be sent to SF as part of the query for highlighted campaigns so we can preview what that will return. Makes sense to take it away until and unless we need to do that.

PR #1828, opened 2025-01-17
DON-1014: Move total raised right on regular giving campaign summary …

bdsl:
Also allowing regular giving & ad-hoc to be mixed in the same grid

image

image

image


src/app/campaign-summary.model.ts
```diff
@@ -1,5 +1,6 @@
 export type CampaignSummary = {
     percentRaised?: number|string;
+    isRegularGiving: boolean;
```
bdsl, line 3: this needs to be added in SF where we send these summaries. Doesn't actually have to be done before this is deployed to prod as long as there are no regular giving campaigns in prod since undefined is falsy.

src/app/explore/explore.component.html
```diff
@@ -67,11 +67,12 @@
           (scrolled)="onScroll()"
           [scrollWindow]="true"
           >
-          @if (this.metaCampaign?.isRegularGiving) {
-            <!-- We assume that we show regular giving campaigns only iff we are looking at a regular giving
-            metacampaign page -->
             @for (campaign of individualCampaigns; track campaign.id) {
               <div>
+                @if (campaign.isRegularGiving) {
```
bdsl, line 72: I think I'm likely to extract everything inside this if soon to go into a new Angular component to re-use in `charity.component.html`

src/app/campaign-summary.model.ts
bdsl: done

dorota-joanna (approved):
LGTM

PR #1834, opened 2025-01-29
DON-1002: Ask for marketing email consent for us & charity when s…

bdsl:
…etting up regular giving mandate

image


src/app/regular-giving/regular-giving.component.html
```diff
@@ -138,6 +138,89 @@
                 </button>
               </div>
 
+            </mat-step>
```
bdsl, line 141: code below mostly copied from donation start form. I don't think there's any nice way to make it be shared, and we may find the two evolve separately.

PR #1835, opened 2025-01-29
BG2-2820: Fix sort order for completed campaigns

bdsl:
Looks like this was broken by changes in ca74a93 / DON-676


bdsl:
closing to favour https://github.com/thebiggive/donate-frontend/pull/1835 with better branch name.

PR #1836, opened 2025-01-30
BG2-2820 fix default sort order

bdsl:


src/app/campaign-details/campaign-details.component.spec.ts
```diff
@@ -68,7 +68,7 @@ describe('CampaignDetailsComponent', () => {
       countries: ['United Kingdom'],
       currencyCode: 'GBP',
       donationCount: 4,
-      endDate: new Date(),
```
bdsl, line 71: These were not realistic tests - data fetched by HTTP using standard JSON deserialize will never contain Date objects. Adjusted types force us to write something more realistic, with strings here instead.

src/app/campaign.service.ts
```diff
@@ -56,11 +56,11 @@ export class CampaignService {
     let dateToUse: Date|undefined;
 
     if (this.isInFuture(campaign)) {
-      dateToUse = campaign.startDate;
```
bdsl, line 59: this is a runtime change - now we actually return a Date object from here at runtime instead of a string. But it shouldn't matter, it gets passed to Angular's DatePipe that can handle either.

dorota-joanna (approved):
LGTM

PR #1841, opened 2025-01-31
DON-1114: Add small charity week to menus outside production

bdsl:
To make this screenshot I've changed the campaign family for the first highlight card on the homepage to `smallCharity`:

The new menu items will be displayed only outside production, so the whole can be QA'd without blocking deployments.

image


src/app/app.component.html
```diff
@@ -23,7 +23,7 @@ <h2>Unsupported Browser</h2>
     [blogUrlPrefix]="blogUriPrefix"
     [experienceUrlPrefix]="experienceUriPrefix"
     [isLoggedIn]="isLoggedIn"
-    [myAccountFlagEnabled]="true"
```
bdsl, line 26: myAccountFlagEnabled is no longer needed to be passed, the component now treats it as always enabled. Flag was just there temporarily for dealing with my account as a new feature in the past.

dorota-joanna (approved):
LGTM

PR #1842, opened 2025-02-03
DON-1115: Remove blank campaign info page sections

bdsl:
We're not expecting these to be filled in for regular giving campaigns, so we don't want to show headings for empty sections on a page.

Similarly for categories and beneficiaries, we shouldn't show the heading if there is no data recorded.

Before:

image

After:

image

### DIFF
--- a/src/app/campaign.model.ts
+++ b/src/app/campaign.model.ts
@@ -33,16 +33,16 @@ export interface Campaign {
      * ISO 8601 formatted datetime
      **/
     endDate: string;
-    impactReporting: string;
-    impactSummary: string;
+    impactReporting: string | null;
+    impactSummary: string | null;
     isMatched: boolean;
     matchFundsRemaining: number;
     matchFundsTotal: number;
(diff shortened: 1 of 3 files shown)


src/app/campaign.model.ts
```diff
@@ -33,16 +33,16 @@ export interface Campaign {
...
     isMatched: boolean;
     matchFundsRemaining: number;
     matchFundsTotal: number;
     parentUsesSharedFunds: boolean;
-    problem: string;
+    problem: string | null;
     quotes: Array<{person: string, quote: string}>;
     ready: boolean;
-    solution: string;
+    solution: string | null;
```
bdsl, line 45: I've found three of these properties to be null for the campaign I'm testing with in `full`, only `problem` is not null but that's also not blank. I think it's safest to treat them all as nullable, particularly while they're being sent from apex which doesn't mark nullability in its type system.

NoelLH (approved):

PR #1843, opened 2025-02-03
DON-1008: WIP add gift aid section to regular giving form

bdsl:


src/app/postcode.service.spec.ts
```diff
@@ -1,18 +1,18 @@
 import { provideHttpClientTesting } from '@angular/common/http/testing';
 import { TestBed } from '@angular/core/testing';
 
-import { PostcodeService } from './postcode.service';
+import { AddressService } from './address.service';
 import { provideHttpClient, withInterceptorsFromDi } from '@angular/common/http';
 
 describe('PostcodeService', () => {
```
NoelLH, line 7: Probably slightly better to rename the file and change its `describe(...)` to avoid confusion

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -456,4 +524,32 @@ export class RegularGivingComponent implements OnInit, AfterViewInit {
...
     return !!this.paymentInfoErrorMessage;
   }
+
+  private validateGiftAidStep(): boolean  {
+    const errors: string[] = [];
+    if (typeof this.giftAid !== 'boolean') {
+      errors.push('Please choose whether you wish to claim Gift Aid.');
+    }
+
+    if (this.giftAid && !this.homeAddressFormValue) {
```
NoelLH, line 534: For the long term I'm not too sure about doing these checks one by one. I think that Angular controls having validators attached – even if we have to add and remove them – might be less error prone. The built-in validators are also what we use elsewhere, so might better set us up for the possibility of reusable Form Groups, allowing different forms to share code for near-identical steps. Definitely not a blocker for now though.

NoelLH (approved):
Looks good, great to see the donate form get a little smaller and using shared class's logic.

Minor things inline, no blockers.

src/app/address.service.ts
```diff
@@ -0,0 +1,117 @@
...
+  constructor(private http: HttpClient) {}
+
+  getSuggestions(partialAddress: string): Observable<GiftAidAddressSuggestion[]> {
+    // Get up to top 20 suggestions – the maximum allowed.
+    const uri = `${environment.postcodeLookupUri}/autocomplete/${encodeURIComponent(partialAddress)}?top=20&api-key=${environment.postcodeLookupKey}`;
+
+    return this.http.get<GiftAidAddressSuggestion[]>(uri).pipe(
+      map((response: any) => response.suggestions),
+      catchError((error) => {
+        console.log('PostcodeService.getSuggestions() error', error);
```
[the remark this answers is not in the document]
bdsl: Thanks, fixed along with some similar things in 23606d24

src/app/postcode.service.spec.ts
bdsl: Yep, done in 23606d24

PR #1844, opened 2025-02-04
DON-1008: Remove warning message about missing matchbot implemenation

bdsl:
For merging after https://github.com/thebiggive/matchbot/pull/1199


bdsl:
Or before TBH, doesn't affect prod today if this particular feature works or not.

PR #1845, opened 2025-02-05
DON-1008: Show gift aid amount on regular giving mandate viewing page

bdsl:
In principle has to wait behind
https://github.com/thebiggive/matchbot/pull/1200 although since regular giving isn't possible yet on prod the prod deployment order doesn't matter.

image

### DIFF
--- a/src/app/exact-currency.pipe.ts
+++ b/src/app/exact-currency.pipe.ts
@@ -27,7 +27,7 @@ export class ExactCurrencyPipe implements PipeTransform {
     // Donor inputs get passed to this pipe directly, so we need to ensure non-numeric
     // input can't crash the app.
     value = Number(value);
-    if (isNaN(value)) {
+    if (! Number.isFinite(value)) {
       return undefined;
     }
 
--- a/src/app/mandate.model.ts
+++ b/src/app/mandate.model.ts
@@ -1,10 +1,13 @@
+export type Money = {
+  "amountInPence": number,
+  "currency": "GBP"
+};
+
 export interface Mandate {
   id: string,
   campaignId: string,
   charityName: string,
   status: 'active'|'pending',
-  // @todo-regular-giving-DON-1008 : we might want to keep the total donated amount as part of the regular giving mandate
-  // giftAidAmount: number = 0;
   "schedule": {
     "type": "monthly",
     "dayOfMonth": number,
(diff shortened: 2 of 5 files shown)


src/app/exact-currency.pipe.ts
```diff
@@ -27,7 +27,7 @@ export class ExactCurrencyPipe implements PipeTransform {
     // Donor inputs get passed to this pipe directly, so we need to ensure non-numeric
     // input can't crash the app.
     value = Number(value);
-    if (isNaN(value)) {
+    if (! Number.isFinite(value)) {
```
bdsl, line 30: not particularly related to the current task but thought we might as well return undefined for infinity as well as for NaN.

src/app/mandate.model.ts
```diff
@@ -13,12 +16,8 @@ export interface Mandate {
...
-    "currency": "GBP"
-  },
-  matchedAmount: {
-    "amountInPence": number,
-    "currency": "GBP"
-  }
+  donationAmount: Money,
+  matchedAmount: Money,
+  totalCharityReceivesPerInitial: Money,
+  giftAidAmount: Money,
```
bdsl, line 22: wondering if giftAidAmount (and I guess matchedAmount) should be optional / nullable. Maybe doesn't make much difference, although if they were optional and expected not to appear if zero then it would simplify the display logic a bit and we wouldn't need separate boolean properties.

A missing giftAidAmount feels a bit more correct than a zero amount when the donor has said they don't want to claim gift aid.

src/app/mandate/mandate.component.html
```diff
@@ -54,19 +54,18 @@ <h2>Thank you!</h2>
...
-            <!--  @todo-regular-giving-DON-1008: consider displaying gift Aid and tip amount on the template (format copied from one-off cmpaign thank you page -->
-            <!--                @if (giftAidAmount > 0) {-->
-            <!--                  <tr>-->
-            <!--                    <th scope="row">Gift Aid Amount</th>-->
-            <!--                    <td>{{ giftAidAmount | exactCurrency:donation.currencyCode }}</td>-->
-            <!--                  </tr>-->
-            <!--                }-->
+            @if (mandate.giftAid) {
+              <tr>
+                <th scope="row">Gift Aid</th>
```
bdsl, line 65: the word 'Amount' in copy as planned here seems unnecessary.

src/app/money.pipe.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Formats a money object as a string. Currently, the money object is only used in regular giving mandates, (hence
+ * the type being defined in the mandate model file) but it may be expanded to much wider use in future).
+ */
+@Pipe({
+  standalone: true,
+  name: 'money',
+})
+export class MoneyPipe implements PipeTransform {
+  transform(money: Money): string {
```
bdsl, line 14: angular is clever enough to read this and stop us using this pipe with any value that isn't statically typed as `Money`.

NoelLH (approved):

PR #1848, opened 2025-02-05
DON-1114: Fix mandate list display - show correct zero GA when applic…

bdsl:
…able

See previous discussion on ticket DON-1025

### DIFF
--- a/src/app/mandate.model.ts
+++ b/src/app/mandate.model.ts
@@ -18,6 +18,10 @@ export interface Mandate {
   giftAid: boolean,
   donationAmount: Money,
   matchedAmount: Money,
-  totalCharityReceivesPerInitial: Money,
   giftAidAmount: Money,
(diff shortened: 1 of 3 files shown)


src/app/mandate.model.ts
```diff
@@ -18,6 +18,10 @@ export interface Mandate {
   giftAid: boolean,
   donationAmount: Money,
   matchedAmount: Money,
-  totalCharityReceivesPerInitial: Money,
   giftAidAmount: Money,
+  /** Includes main amount & GA but not matching */
+  totalIncGiftAid: Money,
```
bdsl, line 23: new property to be calculated in matchbot - I feel we should avoid doing any calculations (other than e.g. converting pence to pounds for display) in FE and keep all the calculations together in matchbot code.

NoelLH (approved):

PR #1850, opened 2025-02-06
Fix Regular Giving jumping back from payment details step right after it's first entered

NoelLH:


bdsl (approved):
LGTM, payment step index will have increased when we added the gift aid question.

PR #1861, opened 2025-02-11
Don 1008 regular giving ga cont

bdsl:


src/app/regular-giving/regular-giving.component.ts
```diff
@@ -230,6 +230,8 @@ export class RegularGivingComponent implements OnInit, AfterViewInit {
       let errorMessage = 'Form error: ';
       if (this.mandateForm.get('donationAmount')?.hasError('required')) {
         errorMessage += "Monthly donation amount is required";
```
bdsl, line 232: I don't think the above can ever actually happen, should have been validated in an earlier step. But not much harm leaving it here for now.

src/app/regularGiving.service.ts
```diff
@@ -20,8 +20,8 @@ type StartMandateParams = {
   /** ISO-2 code for country. Must match that on the DonorAccount if the latter is non-null */
   billingCountry: string,
 
-  /** Must match postcode on the DonorAccount if the latter is non-null */
-  billingPostcode: string,
+  /** Must be null match postcode on the DonorAccount if the latter is non-null */
```
bdsl, line 23:
```suggestion
  /** Must be null or match postcode on the DonorAccount if the latter is non-null */
```

NoelLH (approved):

PR #1862, opened 2025-02-12
Fix cut off 'Regular Giving' at some breakpoints

NoelLH:
Resolves this: 

Screenshot 2025-02-12 at 11 49 57

### DIFF
--- a/src/app/my-account/my-account.component.ts
+++ b/src/app/my-account/my-account.component.ts
@@ -67,7 +67,7 @@ export class MyAccountComponent implements OnInit {
             color: 'brand-afa-pink',
             image: new URL(environment.donateUriPrefix + '/assets/images/peach-texture.jpg')
           },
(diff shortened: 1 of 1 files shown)


src/app/my-account/my-account.component.ts
```diff
@@ -67,7 +67,7 @@ export class MyAccountComponent implements OnInit {
             color: 'brand-afa-pink',
             image: new URL(environment.donateUriPrefix + '/assets/images/peach-texture.jpg')
           },
-          headerText: "Your Regular\u00A0Giving", // u00A0 = non-breaking space. We prefer the break after "Your".
```
bdsl, line 70: Think this looks good. We'd like to encode that we prefer the break after Your but a break after Regular is also allowed, but seems there's no way to do that https://stackoverflow.com/a/18240699/2526181

PR #1870, opened 2025-02-14
DON-997, DON-1102 – new build system; various updates

NoelLH:
~~Maybe close-ish to workable now. Known issues:~~
~~1. Angular Material doesn't seem to use our font as before. e.g. campaign detail tab headers look off at least when served locally.~~
~~2. Angular Material Icons aren't loading, at least sometimes. The only concrete example I've noticed so far is the back icon on 'All participating campaigns' link.~~
~~3. There are some server security etc. middlewares that need to be reinstated into the new/moved `server.ts`~~

Works now, hopefully! Will probably merge to confirm on Staging pre-review.


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -191,14 +189,10 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
   private stepChangeBlockedByCaptcha = false;
   @Input({ required: true }) donor: Person | undefined;
 
-  public tipControlStyle: 'dropdown'|'slider';
-
-  private tipAmountFromSlider: number;
+  public tipControlStyle: 'dropdown';
```
NoelLH, line 192: My understanding is that if we trial a slider now, it will likely be using an off-the-shelf component with better accessibility. So rather than figures out its dependencies, use of NgZone etc., I removed the custom slider for now.

Dockerfile
```diff
@@ -25,7 +25,7 @@ COPY . .
 ARG BUILD_ENV
 
 # Build client bundle and prepare for Server-Side Rendering
-RUN npm run build:ssr:${BUILD_ENV}
+RUN npm run build:${BUILD_ENV}
```
NoelLH, line 28: My understanding from comparing new project boilerplate and searching the web is that there isn't a separate `ssr` build any more, so I tried to make our script names less confusing accordingly.

angular.json
```diff
@@ -15,13 +15,16 @@
       "prefix": "app",
       "architect": {
         "build": {
-          "builder": "@angular-devkit/build-angular:browser",
+          "builder": "@angular-devkit/build-angular:application",
```
NoelLH, line 18: In this file broadly I tried to reintroduce as few of our previous customisations as I could without things breaking, as I wasn't really sure which if any were deliberate adjustments to the old build's boilerplate.

src/app/app-routing.ts
```diff
@@ -1,31 +1,29 @@
-import {ActivatedRouteSnapshot, CanActivateFn, ResolveFn, Router, Routes} from '@angular/router';
+import {ActivatedRouteSnapshot, CanActivateFn, Router, Routes} from '@angular/router';
 
 import {CampaignListResolver} from './campaign-list.resolver';
 import {CampaignResolver} from './campaign.resolver';
 import {CharityCampaignsResolver} from './charity-campaigns.resolver';
-import {campaignStatsResolver} from "./campaign-stats-resolver";
-import {highlightCardsResolver} from "./highlight-cards-resolver";
+import {CampaignStatsResolver} from "./campaign-stats-resolver";
```
NoelLH, line 6: Resolvers became classes and one-per-class mostly because I was not sure about the causes of `inject()` errors. And I'm still not certain if it was necessary but I do find the routes more manageable with this setup so I think it is probably worth keeping regardless.

bdsl:
4986ccae DON-997, DON-1102 first pass updating to new build system

LGTM good, and I can see its following advice from https://angular.dev/tools/cli/build-system-migration#manual-migration-to-the-new-application-builder, 
e.g. to remove venderChunk and buildOptimizer.

10b764ef DON-997, DON-1102 – resolvers to classes so less things are injected; turning off various pieces possibly implicated in issues

Comments out most of the implementations of some functions and routes we'll need `redirectIfAlreadyLoggedIn`, `redirectIfAlreadyLoggedIn`, `handleLogout`, `LoggedInPersonResolver` and
`LoggedInPersonResolver` but these do all I think get restored later in 1aea3a4. Changing resolvers from functions to classes feels a bit like a step backwards compared
to the way Angular is going, but it should work just as well in the short term so not a blocker. Would still be great to work out why we're were getting the error with the
`inject` function but we can try introducing one function resolver or changing one of these back to a function at a time to see if it is that. Or even just changing replacing
the constructor DI in one of these with a call to inject for one function at a time.  

Not a blocker but probably worth explicitly implementing the Resolve interface - on these classes, makes it a bit more self-explanatory how they work as resolvers. I'll do that
in a follow-up PR to this.

d3991e7b DON-997, DON-1102 – WIP – keep App non-standalone [tried it again, seems it can't work well with non-standalone Stencil] but take steps towards new-build-system compatible config

LGTM to me as far as I can tell, lots of small changes that I'm sure all have reasons but can't see what they all are. Curious about the reason for change from . to [] syntax
e.g. `this.selected.beneficiary` to `this.selected['beneficiary']`. Looks like it would work either way. Maybe just clearer to use the square brackets when the object is working as a map?

f061cbe5 DON-997, DON-1102 – WIP – fix various new TS errors etc.

Change to `columnCount="3"` is good, fixes a warning in the IDE. Old way doesn't seem to be a build error, would be even better if we could get issues like 
`Attribute columnCountxxx is not allowed here` to break the build. Good to see we now have TS checks for uninitialized properties.

dcca8525 DON-997, DON-1102 – configure HTTP client better

LGTM

bd5b4846 DON-997, DON-1102 – reinstate previous assets config

LGTM

39b4e9e5 DON-997, DON-1102 – check in the current mix & match of partly commented routes; fix components that are not commented and had TS errors

Noticing things like `stats!` in HomeComponent because of the new TS uninitialized properties rule. I wonder if instead of using the exclamation mark
that effectively suppresses the rule we could move the assignment to that field and similar from `ngOnInit` to `constructor`. Or at just put similar things in constructors
in future.  I think we should keep `RegularGivingComponent.donor` non-nullable, since that page doesn't make sense without knowing who the donor is.

And I get now that the change from . to [] started a few commits back is complying with noPropertyAccessFromIndexSignature. That does look like a good rule to try following
for the future, but I get why it's turned off for now.

1aea3a46 DON-997, DON-1102 – bring back all original routes; put guards inline in routes file again; fix lots of type stuff

Generally LGTM, not commenting on every detail. Agree it makes sense to remove the slider stuff - even if we do think we might eventually bring back
our custom slider it's not worth maintaining it in the codebase in the meantime.

80b8c258 DON-997, DON-1102 – update SSR build config
LGTM

2ef477d6 DON-997, DON-1102 – skip header `closeMobileMenuFromOutside()` for now

LGTM, I see this is reversed later in 3005a9b2

ca6422a5 DON-997, DON-1102 – bring back Stencil rendering of components before Express serves a response

LGTM, I see the todo added here is dealt with in 749a8a892

47c31f2d DON-997, DON-1102 – reduce bogus campaign resolver load & noise

LGTM

ab0fd5c5 DON-997, DON-1102 – fix app base HREF

LGTM

81e36319 DON-997, DON-1102 – reinstate previous `server.ts` providers

LGTM

ee72441b DON-997, DON-1102 – turn off linting for now

LGTM, makes sense to turn it off for now since it's suddenly become a lot more agressive with what it finds. Looks like 
`angular-eslint` does give a useful report, maybe we can see if any of the eslint baselining packages are good or configure
it to be less aggressive and put it back in soon. I think not a blocker, pretty much the only sort of issue I think lint was
catching up to now was unused variables, which are worth removing to make code easier to read but not particularly likely to become bugs.

d8095e6b DON-997, DON-1102 – update tests, finish tidying up dead imports

LGTM

7f44eaec DON-997, DON-1102 – bring back e2e config

LGTM

3a5f6b11 Upgrade build to Node 22

LGTM, 

3005a9b2 DON-997, DON-1102 – reinstating features & tidying up

I see TS is now inferring `Object.keys(sortOptions)` as just string which is why we need the `as` type assertion, not sure
why it doesn't infer the more specific type though - would be good to avoid the assertion if we can. Maybe worth adding a comment 
to either explain why or say we don't know. Otherwise, all LGTM.

007939f9 DON-997, DON-1102 – reinstate assets; fixes font & Material icons

LGTM

606c662c DON-997, DON-1102 – update imports; fixes donate form as non-standalone child component

LGTM

749a8a89 DON-997, DON-1102 – bring back our server customisations

LGTM, not reviewing in details because assuming its just bringing back in this branch what we had already.

7f6bba1d Back to normal `donateUriPrefix` for staging

LGTM

a1ed5a07 Reinstate Matomo tracking injects & update a test

LGTM

d9190c55 Merge pull request #1870 from thebiggive/DON-997-1102-build-config

Looking at overall PR diff and concluding:

Nothing to add beyond comments made on individual commits. Probably if this wasn't so big already I'd suggest trying to avoid some of the type assertions with either
`as` or `!` but I think at this point we probably need to live with them and just try to avoid introducing new similar ones, and/or remove them individually as
and when they each affect our work. As mentioned will make follow-up PR about resolvers.

Definitely still a lot more changes than we'd ideally like to make at one time, which I think comes from working around various bugs that appeared with the
move to 19 that we still don't have great explanations for. Probably nothing to do about that at this point, but just means we need to ready to fix
any possible new bugs introduced with these changes. Looks like there's at least one around the layout on the regular giving mandate creation form that
@DomTBG  found in testing today.

I'll also reply to @NoelLH 's inline comments.

src/server.ts
```diff
@@ -0,0 +1,192 @@
...
+  } else {
+    res.send('User-agent: *\nDisallow: /');
+  }
+});
+
+app.get('/.well-known/apple-developer-merchantid-domain-association', (_req: Request, res: Response) => {
+  res.sendFile(`${browserDistFolder}/assets/stripe-apple-developer-merchantid-domain-association`, {
+    maxAge: '7 days',
+  });
+});
```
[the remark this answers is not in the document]
bdsl: I don't think this is a new issue, probably not anything that needs to be fixed urgently.

src/app/app-routing.ts
bdsl: Thanks, commented on this in my main review comment.

NoelLH:
> Curious about the reason for change from . to [] syntax
e.g. this.selected.beneficiary to this.selected['beneficiary']. Looks like it would work either way. Maybe just clearer to use the square brackets when the object is working as a map?

Yes, as you found later this had errors post-updates from `noPropertyAccessFromIndexSignature`, and I then turned it off rather than continue with far more of these.

As discussed, I'm not sure about using constructors over `ngOnInit` when Angular service DI is needed for setup. I don't know enough about Angular internals to be very confident in that change and it seems to go against received wisdom for Angular.

> https://github.com/thebiggive/donate-frontend/commit/2ef477d67e58c673df308c18cd7ad9a12b50c11f DON-997, DON-1102 – skip header closeMobileMenuFromOutside() for now
>
> LGTM, I see this is reversed later in https://github.com/thebiggive/donate-frontend/commit/3005a9b25ff72ce918f01e3e62452c542b09e54f

Yes – think this became browser-only though. There's a chance this was implicated in server runtime errors but I'm not at all sure.

PR #1871, opened 2025-02-17
DON-1102: Refactor: Explicitly implement Angular's Resolve interface …

bdsl:
…on resolver classes

Doing this also forced me to fix the type of one of the unused _route params.

There's effectively no type checking on resolvers used in app-routing.ts I think since they're used in a union with DeprecatedGuard which is a union with `any`.

See comment at https://github.com/thebiggive/donate-frontend/pull/1870#issuecomment-2662994284


NoelLH (approved):
Nice, thanks! I did briefly think while looking at these it was a bit odd they didn't always use an interface, but I hadn't seen one before and was doing lots of things at once.

PR #1879, opened 2025-02-19
Don-1022: Add mandate cancellation form

bdsl:
image

### DIFF
--- a/src/app/cancel-mandate/cancel-mandate.component.html
+++ b/src/app/cancel-mandate/cancel-mandate.component.html
@@ -1,0 +1,1 @@
+<p>cancel-mandate works!</p>
--- a/src/app/cancel-mandate/cancel-mandate.component.html
+++ b/src/app/cancel-mandate/cancel-mandate.component.html
@@ -1,1 +1,52 @@
-<p>cancel-mandate works!</p>
+<div id="banner"></div>
+<main>
+  <div>
+    <biggive-page-section>
+      <div class="heading-wrapper">
+        <biggive-heading
+          colour="tertiary"
+          size="1"
+          align="center"
+          [text]="'Cancel regular giving to ' + mandate.charityName"
+        ></biggive-heading>
+      </div>
+      <p>
+        You are currently giving <strong>{{ mandate.donationAmount | money }}</strong> to
+        <strong>{{ mandate.charityName }}</strong>,
+        on the {{ mandate.schedule.dayOfMonth | ordinal }} of each month.
+        Click below to stop future payments.
+      </p>
+
+      <p>This will not affect any payments that have already been collected or are currently being processed.</p>
+
+      <form [formGroup]="cancellationForm" (ngSubmit)="cancel(mandate)">
+        <biggive-text-input spaceBelow="4">
+          <label slot="label" for="reason">Please tell us why you wish to stop giving to {{ mandate.charityName }}
+            (optional)</label>
+          <input
+            slot="input"
+            matInput type="email" id="reason" formControlName="reason" autocapitalize="off" [maxlength]="reasonMaxLength">
+        </biggive-text-input>
+
+        <div class="actions">
+          <div>
+            <a href="/my-account/regular-giving/{{mandate.id}}">Back to mandate details</a>
+          </div>
+          <div>
+          <biggive-button
+            space-above="5"
+            spaceBelow="0"
+            colour-scheme="tertiary"
+            [label]="'Stop payments to ' + mandate.charityName"
+            full-width="true"
+            disabled="!loginForm.valid"
+            size="medium"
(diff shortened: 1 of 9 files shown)


src/app/cancel-mandate/cancel-mandate.component.html
```diff
@@ -0,0 +1,52 @@
...
+      <p>
+        You are currently giving <strong>{{ mandate.donationAmount | money }}</strong> to
+        <strong>{{ mandate.charityName }}</strong>,
+        on the {{ mandate.schedule.dayOfMonth | ordinal }} of each month.
+        Click below to stop future payments.
+      </p>
+
+      <p>This will not affect any payments that have already been collected or are currently being processed.</p>
+
+      <form [formGroup]="cancellationForm" (ngSubmit)="cancel(mandate)">
```
bdsl, line 22: this ngSubmit is not actually used so I should probably remove it. I don't think we can use it with `biggive-button` as it is now as our submit button

bdsl:
Pushed a few changes as discussed with @NoelLH and @DomTBG . New screenshot:

image

src/app/cancel-mandate/cancel-mandate.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    MatInput,
+    ReactiveFormsModule
+  ],
+  templateUrl: './cancel-mandate.component.html',
+  styleUrl: './cancel-mandate.component.scss'
+})
+export class CancelMandateComponent implements OnInit, OnDestroy {
+  protected mandate: Mandate;
+  protected readonly reasonMaxLength = 500;
+  protected cancellationForm = new FormGroup({
```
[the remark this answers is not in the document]
NoelLH: Yep, it's pretty nice! We opted existing forms in them after an Angular 14 migration made them temporarily untyped in https://github.com/thebiggive/donate-frontend/pull/659/commits/3b219cbc828f6036567a22d7ed3c7b68fde94f3a

src/app/cancel-mandate/cancel-mandate.component.html
```diff
@@ -0,0 +1,57 @@
...
+      <p>
+        You are currently giving <strong>{{ mandate.donationAmount | money }}</strong> to
+        <strong>{{ mandate.charityName }}</strong>,
+        on the {{ mandate.schedule.dayOfMonth | ordinal }} of each month.
+        Click below to stop future donations.
+      </p>
+
+      <p>This will not affect any donations that have already been collected or are currently being processed.</p>
+
+      <form [formGroup]="cancellationForm" (ngSubmit)="cancel(mandate)">
```
bdsl, line 23:
```suggestion
      <form [formGroup]="cancellationForm"">
```

src/app/cancel-mandate/cancel-mandate.component.html
bdsl: Done.

src/app/cancel-mandate/cancel-mandate.component.ts
```diff
@@ -0,0 +1,58 @@
...
+  ) {
+    this.mandate = route.snapshot.data.mandate;
+  }
+
+  ngOnInit() {
+    this.pageMeta.setCommon('Cancel Regular Giving to ' + this.mandate.charityName, '', null);
+    if (isPlatformBrowser(this.platformId)) {
+      console.log("setting body to primary colour");
+      document.body.classList.add('primary-colour');
+    }
```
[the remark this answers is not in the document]
bdsl: the console log needs to go away but this isn't an experiment. It's the same thing that we do on the login page. Will add a comment.

src/app/cancel-mandate/cancel-mandate.component.html
```diff
@@ -0,0 +1,57 @@
...
+        </biggive-text-input>
+
+        <div class="actions">
+          <div>
+            <a href="/my-account/regular-giving/{{mandate.id}}">Back to mandate details</a>
+          </div>
+          <div>
+            @if (processing) {
+              <mat-spinner color="primary" diameter="30" aria-label="Processing"></mat-spinner>
+            } @else {
```
bdsl, line 40: the spinner seems to never be appearing in my tests, not sure why. Doesn't matter if things work quickly but we should give some indication that something is happening if the connection or matchbot is slow.

NoelLH: @bdsl Your component might need `MatProgressSpinnerModule`?

NoelLH (approved):

src/app/cancel-mandate/cancel-mandate.component.html
bdsl: Thanks I'll check that tomorrow.

PR #1884, opened 2025-02-21
Remove non-working social share links from Mandate view page

bdsl:
As discussed today on Slack #develop

Before:

image

(but clicking the links doesn't work because we never programmed them to share anything specific)

After:

image


src/app/mandate/mandate.component.ts
bdsl, line 1: this file is so short and simple!

PR #1885, opened 2025-02-21
DON-989: Fix display of e.g. "Your card was declined" message on regu…

bdsl:
…lar giving signup page

### DIFF
--- a/src/app/backendError.ts
+++ b/src/app/backendError.ts
@@ -1,6 +1,8 @@
 import {Money} from "./mandate.model";
 
 type errorDetail = {
+  message?: string;
+  publicMessage?: string;
   description?: string
   type: string
   htmlDescription?: string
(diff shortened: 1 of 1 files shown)


src/app/backendError.ts
```diff
@@ -35,7 +37,13 @@ export type InsufficientMatchFundsError = BackendError &
...
 
 export function errorDescription(error: BackendError): string {
-  return error.error.error.description || error.message || 'Sorry, something went wrong'
+  const errorDetail = error.error.error;
+
+  return errorDetail.description
+    || errorDetail.publicMessage
+    || errorDetail.message
+    || error.message
+    || 'Sorry, something went wrong'
```
bdsl, line 46: message now appears like so:

image

Without this change instead of saying "your card was decline a more technical message about an http response code.

PR #1887, opened 2025-02-25
Don 1119 regular giving 3 ds next action

bdsl:
PR stacked on top of #1886 

See also https://github.com/thebiggive/matchbot/pull/1234

If mandates were in live use in prod then the two PRs would need to be deployed simultaneously, but as they're not we can deploy in either order.

### DIFF
--- a/src/app/mandate/mandate.component.ts
+++ b/src/app/mandate/mandate.component.ts
@@ -54,7 +54,7 @@ export class MandateComponent implements OnInit {
           window.clearInterval(this.mandateRefreshTimer);
         }
       },
-      5_000
+      2_000
     )
   }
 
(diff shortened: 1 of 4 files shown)


src/app/mandate/mandate.component.ts
```diff
@@ -1,30 +1,67 @@
...
+        if (updatedMandate && updatedMandate.status !== 'pending') {
+          this.mandate = updatedMandate;
+          window.clearInterval(this.mandateRefreshTimer);
+        }
+
+        if (this.timedOut) {
+          window.clearInterval(this.mandateRefreshTimer);
+        }
+      },
+      5_000
```
NoelLH, line 57:
```suggestion
      2_000
```
Maybe? 5s minimum seems a little long to have to wait.

NoelLH (dismissed):
Notwithstanding the Q about when 3DS fails (more on MatchBot PR), LGTM

PR #1890, opened 2025-02-28
DON-1005: Don't show regular giving form if donor already has active …

bdsl:
…regular giving agreement for same charity

image


src/app/my-mandates.resolver.ts
bdsl, line 1: renamed because this no longer shows only active mandates. It shows any mandates that the donor would expect to see, including e.g. cancelled.

PR #1892, opened 2025-03-03
DON-1021: Allow changing card used for regular giving

bdsl:
Added this part on the payment methods page (from selected heading down), only for the case where the donor has a set regular giving payment method. (so should not be visible on prod where no-one has such a method yet)

image

And this page that the button links to:

image


bdsl:
Pushed a WIP commit - will come back to this later today image

bdsl:
Closing as this isn't quite working - see Jira for details. May re-open or redo tomorrow.

PR #1896, opened 2025-03-04
Update SCSS rules to remove deprecated declarations after nested rule…

bdsl:
…s, keep behaviour

The use of declarations after nested rules is currently deprecated in Sass. Moving the declarations above the nested rules preserves the existing behavour in current and future versions of Sass.

See https://sass-lang.com/documentation/breaking-changes/mixed-decls/

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -43,6 +43,11 @@
             "prerender": true,
             "ssr": {
               "entry": "src/server.ts"
+            },
+            "stylePreprocessorOptions": {
+              "sass": {
+                "fatalDeprecations": ["mixed-decls", "color-functions", "global-builtin", "import"],
+              }
             }
           },
           "configurations": {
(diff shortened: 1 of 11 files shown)


angular.json
```diff
@@ -43,6 +43,11 @@
             "prerender": true,
             "ssr": {
               "entry": "src/server.ts"
+            },
+            "stylePreprocessorOptions": {
+              "sass": {
+                "fatalDeprecations": ["mixed-decls", "color-functions", "global-builtin", "import"],
```
bdsl, line 49: This means that any use of these deprecated sass facilities will now be a build error, as shown at https://github.com/thebiggive/donate-frontend/pull/1897 . 

I'd rather have just put a version number here (e.g. "1.83.1") which according to sass docs should make this apply for all deprecations introduced in that version and earlier, but it doesn't seem to work: I get
`[WARNING] Invalid deprecation "1.83.1". [plugin angular-sass]`

PR #1908, opened 2025-03-18
DON-1021: Allow removing card used for regular giving.

bdsl:
Requires accompanying change in matchbot

image

### DIFF
--- a/src/app/my-payment-methods/my-payment-methods.component.html
+++ b/src/app/my-payment-methods/my-payment-methods.component.html
@@ -133,6 +133,17 @@
           size="medium"
           url="/my-account/payment-methods/change-regular-giving"
         />
+
+        <!-- not necessarily needed for regular giving MVP but would be good to have a busy state option built
+        in to the button in case the donor is waiting for a slow network -->
+        <biggive-button
+          space-above="5"
+          colour-scheme="white"
+          label="Remove regular giving payment method"
+          full-width="false"
+          size="medium"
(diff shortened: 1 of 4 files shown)


bdsl:
Error message appearance when we can't remove the card because there is an existing mandate:

image

Sahil suggested I think rightly that people won't understand the error message and we'll get support requests - we should provide a link or button for people to cancel their mandates with the error. And also that the word 'mandate' in general might not be understood.

src/app/my-payment-methods/my-payment-methods.component.html
```diff
@@ -133,6 +133,17 @@ <h3>
           size="medium"
           url="/my-account/payment-methods/change-regular-giving"
         />
+
+        <!-- not necessarily needed for regular giving MVP but would be good to have a busy state option built
```
NoelLH, line 137: In other places we're using conditionals and a (non-Components) spinner for this. So maybe it would work better as an Angular component wrapping that logic?

NoelLH (approved):

src/app/my-payment-methods/my-payment-methods.component.html
bdsl: Yes that might work, we'd have to get out of the habit of directly using our component button in angular then, at least for anything that might require user to wait.

I guess even nicer than a wait state that's manually assigned might be if we could pass the component an async onclick function and it would automatically change its appearance before and after running that function.

PR #1911, opened 2025-03-19
DON-1131 – move last pre-capture donation Update call earlier

NoelLH:
And improve Matomo error logging in that area slightly.


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -689,6 +696,68 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+      // todo log to Matomo
+      this.toast.showError("Sorry, we can't submit your donation right now.");
+      return;
+    }
+
+    if (this.paymentGroup.value.billingPostcode) {
+      this.donation.billingPostalAddress = this.paymentGroup.value.billingPostcode;
+      this.donation.countryCode = this.paymentGroup.value.billingCountry;
+      this.donationService.updateLocalDonation(this.donation);
+    }
```
NoelLH, line 733: With this in its later position I got a MatchBot validation / assertion error on billing postcode; can't remember now if on Update or Confirm.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -689,6 +697,64 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+
+    this.donationService.update(this.donation)
+      .pipe(retry({ count: 3, delay: getDelay() }))
```
NoelLH, line 735: Keeping a retry but replacing the deprecated RxJS operator use, and slightly simplifying our delay timing logic for legibility

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -632,9 +632,14 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
                 Donate now
               </button>
             }
+            <div aria-live="polite">
+              @if (runningFinalPreSubmitUpdate) {
```
NoelLH, line 636: This spinner replaces the payment button _before_ it is pressed so we know we are ready to ask Stripe to confirm almost immediately. Typically it should show before the button for less than a second in Production.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: @NoelLH to clarify was that an expected error when you provided an invalid postcode? Or an error that's a problem? I'm assuming the former.

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -621,7 +621,7 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
                 </tr>
               </table>
             }
-            @if (!submitting) {
+            @if (!submitting && !runningFinalPreSubmitUpdate) {
```
bdsl, line 624: Haven't read all the way through but if I'm understanding right we need to prevent the donation being made until the final pre submit update is complete. So I think it's more robust to make this like 

```suggestion
            @if (!submitting && finalPreSubmitComplete) {
```

`finalPreSubmitComplete` would be a property that is initialized to false and set true when we get a successful return from matchbot.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
NoelLH: In its previous position in the code with some other updates, it was an unexpected error. The value was not in the payload sent to MatchBot even though it was on the page. I lost the thread as to exactly why the move was needed and have little memory of it now.

Moved here, so far it seems to work.

src/app/donation-start/donation-start-form/donation-start-form.component.html
NoelLH: Not sure – it seems harder to track robustly because donors could go back to earlier steps and we want to do this every time they [re-]enter Confirm.

src/app/donation-start/donation-start-form/donation-start-form.component.html
bdsl: Should be fine, as on previous comment maybe make it show if `! finalPreSubmitComplete` instead of if `runningFinalPreSubmitUpdate` . The donor isn't really interested in what we're running, they only care if we're ready for them to click.

But another thought is to avoid the distracting spinner we could show the button optimistically expecting that the donor won't click it during that one second. Probably not a good experience for donors who do click quickly and we'd have to give an error message to saying they were too quick.

NoelLH: I don't think the last option is a good one when times are around a second and there is nothing else required on the step.

bdsl: Even more robust might be if there's a way to key it to the state of the form - maybe if we can get a hash of it or something - and make the confirm button only available if the final submit has been run.

Or we just start with final `finalPreSubmitComplete` = false, also set it to false any time the form is updated, e.g. `this.donationForm.valueChanges.subscribe(()=> this.finalPreSubmitComplete = false);` (although I know calling subscribe is discouraged) and set it to true only on the 2xx from matchbot.

bdsl: on the other hand the most critical part of the donation is the amount which is immutable anyway, so we don't have a risk of taking the wrong amount by not tracking this. Maybe we don't need to be that robust.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -689,6 +697,64 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     }
   }
 
+  private runFinalPreSubmitUpdate() {
+    // Even if next guard fails, we want to prevent attempting to pay.
+    this.runningFinalPreSubmitUpdate = true;
+
+    let requiredDonationProperties = [this.donation, this.campaign, this.campaign.charity.id, this.psp];
+    // Return if any of the required properties are missing.
```
bdsl, line 705: I'm not seeing a return statement here, am I missing something?

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -689,6 +697,64 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+      this.donation.countryCode = this.paymentGroup.value.billingCountry;
+      this.donationService.updateLocalDonation(this.donation);
+    }
+
+    this.donationService.update(this.donation)
+      .pipe(retry({ count: 3, delay: getDelay() }))
+      .subscribe({
+        next: async (donation: Donation) => {
+          this.runningFinalPreSubmitUpdate = false;
+          this.donationService.updateLocalDonation(donation);
```
bdsl, line 739: Doesn't imediatly matter since updateLocalDonation is non-async but better to delay runningFinalPreSubmitUpdate until its done.

```suggestion
          this.donationService.updateLocalDonation(donation);
          this.runningFinalPreSubmitUpdate = false;
```

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -689,6 +697,64 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     }
   }
 
+  private runFinalPreSubmitUpdate() {
+    // Even if next guard fails, we want to prevent attempting to pay.
+    this.runningFinalPreSubmitUpdate = true;
+
+    let requiredDonationProperties = [this.donation, this.campaign, this.campaign.charity.id, this.psp];
```
bdsl, line 704:
```suggestion
    const requiredDonationProperties = [this.donation, this.campaign, this.campaign.charity.id, this.psp];
```

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -741,7 +807,12 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
   }
 
   async submit() {
-    if (this.donation && this.donationForm.invalid) {
+    // Can't proceed if donation or campaign data issue. Prior check fn has already reported the issue.
+    if (this.donationUpdateError) {
+      return;
```
bdsl, line 812: the user just clicked the button now though didn't they? I think we should show an error toast any time the user clicks and we can't do what they want, even if we have already told them about the issue. The alternative is to disable the button but we moved away from that.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1424,14 +1433,21 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
     return false;
   }
 
+  private getCurrencySymbol(currencyCode: string): string {
```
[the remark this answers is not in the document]
bdsl: Assuming this works looks fine. Could simplify if we say we only have GBP supported but lets leave as you have it.

src/app/observable-retry.ts
```diff
@@ -1,32 +1,20 @@
...
-    finalize(() => {}),
-  );
-};
+const excludedStatusCodes = [500];
+export const getDelay = () => {
+  let scalingDuration = 1_000; // ms, initial
+  return (
+  (error: HttpErrorResponse, retryCount: number) => {
+    // if response is a status code we don't wish to retry, throw error
+    if (excludedStatusCodes.find(e => e === error.status)) {
```
bdsl, line 10: I think this is a simpler way to write the same thing:

```suggestion
    if (excludedStatusCodes.includes(error.status)) {
```

src/app/observable-retry.ts
```diff
@@ -1,32 +1,20 @@
...
+      return throwError(() => error);
```
bdsl, line 11:
```suggestion
      throwError(() => error);
```

bdsl (commented):
Reviewed, all comments inline.

I think these two points just should be resolved before merge:

https://github.com/thebiggive/donate-frontend/pull/1911#discussion_r2003678251

https://github.com/thebiggive/donate-frontend/pull/1911#discussion_r2003687538

src/app/donation-start/donation-start-form/donation-start-form.component.html
bdsl: I don't think it's a critical issue.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
NoelLH: Nope think this is from before I moved to using `forEach()`, fixed in b37a54e471683882629b389d9c9964de067f68f3

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: So I guess if the user goes to the confirm step, `donationUpdateError` is set true because `update` throws. 

Then the user takes as long as they like to think about it, forgets what the issue was, and then they click `Donate now` which is directly bound to this submit function.

Just thought we should repeat the toast to remind them.

NoelLH: Thanks, as discussed I was confused about the specific event. Donors could have entered the step seconds or minutes before so we should show an extra toast if the button is to remain active. Done in 00d2a9dc0fdbfed2d474fcb54977eb66d915fbf3

PR #1917, opened 2025-03-20
Remove listener & Stripe element on Regular Giving component destroy

NoelLH:
Same as single donation counterpart mostly


bdsl (approved):
Not sure for certain whether part or all of this is needed to prevent memory leaks but won't do any harm.

NoelLH:
I think I'm agreed that I'm now unsure if it's strictly needed. Maybe if we get clues otherwise or are working in similar areas we can delete both components' tidying later; for now, merging so they're consistent with each other.

PR #1918, opened 2025-03-20
Don 1122 know that donations will be matched

bdsl:
Two additions to the regular giving form - both copied from and sharing styling with the ad-hoc giving form:

Green match funds available highlight at the top:

image

Amounts summary table immediately before the confirmation button at the end:

image


NoelLH (approved):
LGTM

PR #1919, opened 2025-03-20
BG2-2846 – fix infinite scroll after cached child campaigns are restored

NoelLH:


src/app/explore/explore.component.ts
```diff
@@ -421,24 +421,8 @@ export class ExploreComponent implements AfterViewChecked, OnDestroy, OnInit {
...
-          undefined,
-          this.matomoTracker,
-        );
-        this.loading = false;
-      },
-    });
-
     if (!isPlatformBrowser(this.platformId)) { // Server renders don't need the scroll restoration help
-      // this.doCampaignSearch(query as SearchQuery, true); // Clear existing children, though there _should_ be none on server
+      this.doCampaignSearch(query as SearchQuery, true); // Clear existing children, though there _should_ be none on server
```
NoelLH, line 425: I'm not really clear why this was commented while the above block doing something similar was added; I'm assuming an oversight. I think that running this at this early _iff_ on the server makes sense; we saw bugginess when the browser both ran a new search (above) and restored its old data from Storage, because the offset & the data ended up mutually contradictory.

bdsl: Yes, it must have been a mistake, if it was deliberate I would have either fully deleted or made comment to explain why I thought it was interesting enough to leave commented-out. Apologies.

bdsl (approved):
LGTM

src/app/explore/explore.component.ts
NoelLH: All good! I think there's been some committing and reverting in this area so I'm not even sure if it was your change originally.

PR #1920, opened 2025-03-20
MAT-409: Cancel donation on failed card 3DS authentication

bdsl:
Unless we want to make changes in matchbot as well we need to abandon this donation and create a new one to allow the donor to try using a different card.

Attemping to confirm a donation a second time after 3DS authentication has been attempted leads to Stripe throwing this exception at matchbot:

  InvalidRequestException: This PaymentIntent's capture_method could not be
  updated because it has a status of requires_action. You may only update the
  capture_method of a PaymentIntent with one of the following statuses:
  requires_payment_method, requires_confirmation.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -61,7 +61,6 @@ import {updateDonationFromForm} from "../updateDonationFromForm";
 import {sanitiseCurrency} from "../sanitiseCurrency";
 import {PaymentReadinessTracker} from "./PaymentReadinessTracker";
 import {requiredNotBlankValidator} from "../../validators/notBlank";
-import {flags} from "../../featureFlags";
 import {WidgetInstance} from "friendly-challenge";
 import {Toast} from "../../toast.service";
 import {GIFT_AID_FACTOR} from '../../Money';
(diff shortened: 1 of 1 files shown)


NoelLH (approved):
LGTM, definitely seems better than what we have now.

Do you know if Stripe offers any way to re-enter 3DS as a further possible optimisation? e.g. send a second text. Or is it game over for that payment intent from their perspective too?

bdsl:
> LGTM, definitely seems better than what we have now.
> 
> Do you know if Stripe offers any way to re-enter 3DS as a further possible optimisation? e.g. send a second text. Or is it game over for that payment intent from their perspective too?

From what I could tell it's possible to re-enter 3DS with the same payment card - we'd just have to call the same function again - but not to change to a different card. To use a different card I think you need a new PI.

NoelLH:
> From what I could tell it's possible to re-enter 3DS with the same payment card - we'd just have to call the same function again - but not to change to a different card. To use a different card I think you need a new PI. Our form design currently doesn't have a way to tell the user they need to retry the authentication with the same card and not change to a different card.

OK, thanks. I guess most 3DS systems keep you in the process unless you press a give up button or look highly suspicious, in which case assuming a new card is the best recovery path makes sense. If there were a significant number of banks that gave up faster I would suggest asking the donor which course of action to take instead. But I think the quite low volume of the error we've seen suggests that preferring to switch card is a good assumption.

PR #1922, opened 2025-03-21
Maybe remove leftover billing address validation errors when those fields become superfluous

NoelLH:


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -2027,6 +2027,8 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
   private removeStripeCardBillingValidators() {
     this.paymentGroup.controls.billingCountry!.setValidators([]);
     this.paymentGroup.controls.billingPostcode!.setValidators([]);
+    this.paymentGroup.setErrors(null);
```
NoelLH, line 2030: This line kind of based on https://stackoverflow.com/questions/54627627/clearvalidators-used-with-updatevalueandvalidity-does-not-clear-errors-from. Not sure if it’s still true that `updateValueAndValidity()` alone is insufficient as that was some years ago.

bdsl (approved):

PR #1925, opened 2025-03-21
DON-1126: Split mandate page into two variants: one for immediate con…

bdsl:
…firmation / thanks and one for later review

Before the mandate page always appeared the same, whether viewed as a redirect from creating the mandate or found at a later date via My account:

image

_______

After this PR this splits into two variants. Changes are mostly to reflect the different use cases of confirming/thanking for a new mandate and reviewing an old one, but also include some other stuff I noticed like replacing "Your Donation Summary" with "Your Regular Donation to {{charityName}}" which I think is clearer.

Thanks variant:

image

Non-thanks variant:

image

There is also an automatic redirect from the thanks variant to non-thanks in case the mandate is more than 24 hours old.

### DIFF
--- a/src/app/app-routing.ts
+++ b/src/app/app-routing.ts
@@ -148,6 +148,20 @@ export const routes: Routes = [
     redirectTo: 'donate/:campaignId',
   },
   {
+    path: `${myRegularGivingPath}/:mandateId/thanks`,
+    pathMatch: 'full',
+    component: MandateComponent,
+    data: {
+      isThanks: true
+    },
+    canActivate: [
+      requireLogin,
+    ],
+    resolve: {
+      mandate:  MandateResolver,
+    },
+  },
+  {
     path: `${myRegularGivingPath}/:mandateId`,
     pathMatch: 'full',
     component: MandateComponent,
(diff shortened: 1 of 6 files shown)


src/app/app-routing.ts
```diff
@@ -147,6 +147,20 @@ export const routes: Routes = [
     pathMatch: 'full',
     redirectTo: 'donate/:campaignId',
   },
+  {
+    path: `${myRegularGivingPath}/:mandateId/thanks`,
```
NoelLH, line 151: Wondering if this might be slightly quicker for devs to follow if we used a query param instead of an extra path piece? But I don't feel strongly, this is also good if you think 2 routes is clearer.

NoelLH (approved):
LGTM

src/app/app-routing.ts
bdsl: I think I'd rather keep it as two routes, I think it looks neater without query params, and using a query param more invites users to edit the URL manually. These are really two different pages that just happen to be very similar and use the same data from the backend. I partly thinking it might be better if they were separate components, since there's several bits varied, but I won't change that now.

PR #1930, opened 2025-03-31
DON-1128: Start displaying mandate cancellation form while mandate lo…

bdsl:
…ading

Previously none of the form displayed until the mandate was returned from the server, meaning we only showed the header and footer of the page in that time. Although I had anticipated that the component would be delayed loading until the resolver resolves the mandate, that didn't really happen as the resolver was only returning an observable of the mandate, not the mandate itself.

There must be something built in to Angular that's implicitly unwrapping the observable so we can use the mandate on this page.

I wasn't sure what the best way to fix this was and ended up being a bit indecsive. The solution here is to display a lot more of the page before the mandate is loaded, so that users get a better idea of what's happening and can start to understand the new context while they wait for loading to finish. If the mandate doesn't load within 1 second then I also show a placeholder title and a spinner. I didn't want to go the effort of building a full skelton UI.

Although this improves the Cancel Mandate page I think we still have the same issue on the View Mandate page - so would be worth discussing how to deal with this type of issue generally on the site.

Othere options I looked at were:

-  Making the observable return a promise of a mandate instead of an observable. This has the advantage of fully stopping the loading of the component page until the mandate is there, so we don't have to handle the case of it not being loaded within the component. Would simplify things especially if we used that pattern on several pages. But it has the problem that there is no visible indication of loading - the URI doesn't update while waiting for the resolve as we use the default 'deferred'  urlUpdateStrategy, not 'eager'. If we wanted to use this method I think we'd also need to create a visible loading indication in the App component, maybe with something like https://blog.angular-university.io/angular-loading-indicator/ . It might need to be built into the main menu component.

- Caching the mandate so we don't have to wait for it to load. We might want to do something similar to how the CampaignResolver uses an inject TransferState (although this wouldn't be transferred from the server) to cache copies of Campaigns, or maybe something more like how DonationService stores donation data in local storage (although I think storage attached to the Window object is probably enough for mandates, we don't to use Local Storage). This has a risk of showing outdated mandate details if we don't get it quite right. I tried using a record mapping mandate IDs observables with the shareReplay pipe in the Mandate service to do something like this, but didn't quite get it to work. But that is why links between two mandate pages now use routerLink instead of href - href links destroy service state, which routerLinks don't.

### DIFF
--- a/src/app/cancel-mandate/cancel-mandate.component.ts
+++ b/src/app/cancel-mandate/cancel-mandate.component.ts
@@ -54,8 +54,6 @@ export class CancelMandateComponent implements OnInit, OnDestroy {
 
   ngOnInit() {
     this.mandateSubscription = this.mandate$.subscribe(mandate => {
-      console.log({mandate});
-
       if (mandate.status === 'cancelled') {
         void this.router.navigateByUrl(`/my-account/regular-giving/${mandate.id}`)
         return;
--- a/src/app/cancel-mandate/cancel-mandate.component.ts
+++ b/src/app/cancel-mandate/cancel-mandate.component.ts
@@ -1,18 +1,19 @@
 import {Component, inject, OnDestroy, OnInit, PLATFORM_ID} from '@angular/core';
-import {ActivatedRoute, Router} from '@angular/router';
+import {ActivatedRoute, Router, RouterLink} from '@angular/router';
 import {ComponentsModule} from '@biggive/components-angular';
 import {Mandate} from "../mandate.model";
 import {MoneyPipe} from '../money.pipe';
 import {OrdinalPipe} from '../ordinal.pipe';
 import {FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators} from '@angular/forms';
 import {MatInput} from '@angular/material/input';
-import {isPlatformBrowser} from '@angular/common';
+import {AsyncPipe, isPlatformBrowser} from '@angular/common';
 import {PageMetaService} from '../page-meta.service';
 import {RegularGivingService} from '../regularGiving.service';
 import {BackendError, errorDescription} from '../backendError';
 import {Toast} from '../toast.service';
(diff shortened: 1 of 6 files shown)


bdsl:
Screencast with an artificial 2 second pause added on the server to simulate a slow connection or very busy server:

[Screencast from 2025-03-31 19-02-52.webm](https://github.com/user-attachments/assets/dfa3966a-1d5e-4ac0-b82d-9b00985e6e72)


Screencast with no artificial pause:

[Screencast from 2025-03-31 19-04-30.webm](https://github.com/user-attachments/assets/603e0024-bc8d-4147-8758-380f68a8953b)

And for comparison previous version of frontend with the 2s artificial pause on server:

[Screencast from 2025-03-31 19-08-05.webm](https://github.com/user-attachments/assets/e4d9ac1d-fb77-404a-a1a3-a2550f63d21b)

src/app/cancel-mandate/cancel-mandate.component.ts
```diff
@@ -45,30 +49,36 @@ export class CancelMandateComponent implements OnInit, OnDestroy {
...
   }
 
   ngOnInit() {
-    if (this.mandate.status === 'cancelled') {
-      void this.router.navigateByUrl(`/my-account/regular-giving/${this.mandate.id}`)
-      return;
-    }
+    this.mandateSubscription = this.mandate$.subscribe(mandate => {
+      console.log({mandate});
+
```
NoelLH, line 58:
```suggestion
```

NoelLH (dismissed):
I think this approach looks good for now.

I think we should be using `routerLink` as standard where it is a straightforward choice anyway, so I'm not sure I understand the benefit of the local caching option. It seems that apart from this specific navigation, we would likely nearly always need to check for the latest state of the Mandate before proceeding.

I've replied more about the general case of avoiding full page loads in Jira.

PR #1936, opened 2025-04-08
Id 47 email verification patch donor info earlier

bdsl:


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -703,6 +703,62 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+    };
+
+    if (donation.giftAid) {
+      person.home_address_line_1 = donation.homeAddress;
+      person.home_postcode = donation.homePostcode === OVERSEAS ? undefined : donation.homePostcode;
+      person.home_country_code = donation.homePostcode === OVERSEAS ? OVERSEAS : 'GB';
+    }
+
+    return person;
   }
```
bdsl, line 762: functions moved mostly unchanged from thanks component

bdsl: possibly small risk that if someone reloads the site before going to thanks and switches from a version before this PR to a version after they won't be able to save their password, but I think that's an acceptable risk if we don't deploy at a busy time.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -703,6 +703,62 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
         billingPostcode: this.giftAidGroup.value.homePostcode,
       });
     }
+
+    if (
+      event.selectedIndex > this.paymentStepIndex &&
```
NoelLH, line 708: I still have a slight preference for using the step names for new logic, but as we already have this index to maintain it seems reasonable to use it an extra time. I'm less sure about calling the update on both entry to the comms step and entry to the confirm step; seems slightly wasteful while we are not saving comms answers to the Person. Not a blocker.

NoelLH (approved):
LGTM. Inline suggestions are to save (I think) roughly 1 API call per donation and tidy up moved code slightly for its new context, but aren't blockers.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
bdsl: I think we need to do it on both the comms step and the confirm step because it's possible to have already filled in the comms step, go back to change your email address, then skip the comms step and go straight to confirm. We could optimize by checking if the email address has been changed but I'll leave it for now and maybe consider that later.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -703,6 +703,62 @@ export class DonationStartFormComponent implements AfterContentChecked, AfterCon
...
+  }
+
+  /**
+   * Updates the name and email address of the new person in the backend. Required at this stage for
+   * the email address in particular so that there will be a ready to use email verification token
+   * for when the donation is confirmed.
+   */
+  private updateNewPersonInBackend(donation: Donation) {
+    const idAndJWT = this.identityService.getIdAndJWT();
+    if (idAndJWT) {
```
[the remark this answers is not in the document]
bdsl: Yes, done in https://github.com/thebiggive/donate-frontend/pull/1936/commits/94820b80aac5baca25dd328cc3368aeed97eab9e . Guard clauses are much easier to read, thanks.

I think the first if should always be true by the time this part of the form is reached so I've made that guard clause log an error if it's not.

PR #1939, opened 2025-04-10
ID-47: Allow donor to view post-donation registration form

bdsl:
The plan is they will click a link in the donation thanks email (link generation not yet implemented) to a URI that includes the six digit secret token and their identity UUID, e.g.
/register?c=019816&u=1f0161fa-63ea-6fa8-888d-b1edd20e1c85

Donate then queries the identity backend for the details of the person, and displays in the registration form as implemented in this commit.

The next step is to create a route in identity to allow registration to be done using those details plus the newly chosen password, and implement calling that route here.

image

### DIFF
--- a/src/app/register/register.component.html
+++ b/src/app/register/register.component.html
@@ -15,7 +15,7 @@
       @if (redirectPath.startsWith('/regular-giving')) {
         <p>Please register for a a Big Give donation account to set up your regular giving agreement.</p>
         <p>You will be able to log-in any time to see your donations or cancel the agreement.</p>
-      } @else {
+      } @else if (! this.emailVerificationToken){
       <p>
         Registering for a Donor Account is quick and easy.
       </p>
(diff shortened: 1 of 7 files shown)


src/app/register/register.component.html
```diff
@@ -42,6 +42,70 @@
...
         </app-verify-email>
 
+      } @else if (this.emailVerificationToken?.valid) {
+        <p>
+          For a smoother giving experience create a Big Give account. Your information will be saved securely, making
+          future donations quicker and easier.
+        </p>
+
+        <p>
+          <strong>Your payment card</strong> will be saved for reuse, if you selected "
```
NoelLH, line 52: Seems like this opening quote would show more normally and look better on the next line. Not sure if that's also an issue in the dialogs or if something about the Angular Material styling makes it work better there.

NoelLH (approved):
LGTM, no blockers

PR #1940, opened 2025-04-11
ID-47: Allow setting password for new account after donating and reci…

bdsl:
…eving email

Adding the link to the registration page to the email is still to be implemented in matchbot and mailer

### DIFF
--- a/src/app/identity.service.ts
+++ b/src/app/identity.service.ts
@@ -269,6 +269,7 @@ export class IdentityService {
    * message.
    */
   async setFirstPasswordWithToken(password: string, {person_uuid, secretNumber}: EmailVerificationToken): Promise<Object> {
+    // @todo add an Identity route with UUID included & then switch Donate's used route over
     const uri = `${environment.identityApiPrefix}/people/setFirstPassword`;
 
     return firstValueFrom(this.http.post(
--- a/src/app/identity.service.ts
+++ b/src/app/identity.service.ts
@@ -262,7 +262,25 @@ export class IdentityService {
 
     return firstValueFrom(this.http.get(uri).pipe(map(((response: any) => response.token))));
   }
+
+  /**
+   * After a non-logged in donor donates they have an account but there is no password set on it. This method
+   * allows setting a password using the secret number that we emailed to them in their donation confirmation
+   * message.
+   */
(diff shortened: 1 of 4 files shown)


src/app/register/register.component.ts
```diff
@@ -247,8 +242,42 @@ export class RegisterComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    this.processing = true;
+    this.registerPostDonationForm.markAllAsTouched();
+    const password = this.registerPostDonationForm.value.password
+
+    const passwordErrors = this.registerPostDonationForm.controls.password.errors;
+    if (passwordErrors?.required) {
+      this.error = "Password is required"
+      this.processing = false
+      return;
+    }
```
bdsl, line 255: really we should be checking for other possible errors as well here, e.g. the new password being too short. But if we don't check it it's not that bad, identity server checks it the donor sees a readable error message.

I feel like there must be a much better way to handle form errors, it should be too hard to iterate through all errors on a form or form control and make a human readable message.  But that seems like something that should be in a library for Angular rather than custom code. I could find lots of blogs and tutorials about showing angular form errors but not so far a library function we can just pull from NPM.

src/app/identity.service.ts
```diff
@@ -262,7 +262,25 @@ export class IdentityService {
...
     return firstValueFrom(this.http.get(uri).pipe(map(((response: any) => response.token))));
   }
+
+  /**
+   * After a non-logged in donor donates they have an account but there is no password set on it. This method
+   * allows setting a password using the secret number that we emailed to them in their donation confirmation
+   * message.
+   */
+  async setFirstPasswordWithToken(password: string, {person_uuid, secretNumber}: EmailVerificationToken): Promise<Object> {
+    const uri = `${environment.identityApiPrefix}/people/setFirstPassword`;
```
NoelLH, line 272: `${environment.identityApiPrefix}/people/:uuid/setFirstPassword` is a bit more consistent with our approach elsewhere, but not a blocker. (Should have mentioned on ID PR sorry)

NoelLH, line 272:
```suggestion
    // @todo add an Identity route with UUID included & then switch Donate's used route over
    const uri = `${environment.identityApiPrefix}/people/setFirstPassword`;
```

NoelLH (approved):

PR #1942, opened 2025-04-14
Id 47 register send code

bdsl:


src/app/verify-email/verify-email.component.ts
```diff
@@ -23,6 +23,10 @@ export class VerifyEmailComponent {
 
   @Input({required: true}) emailAddress!: string;
 
+  // @todo id-47 - work out why this doesn't seem to be updating when value in parent `register` component is updated
+  // in case where wrong verification code is entered.
```
NoelLH, line 27: If this isn't working on first pass, would it be worth trying Signals as a possibly easier alternative?

PR #1943, opened 2025-04-14
ID-47: Require email address verification for new accounts everywhere…

bdsl:
… but prod

Needs to wait to merge until after

https://github.com/thebiggive/donate-frontend/pull/1942 , https://github.com/thebiggive/identity/pull/202 , and https://github.com/thebiggive/matchbot/pull/1304
otherwise it will be impossible to create new accounts.


NoelLH (approved):

bdsl:
Actually I realised this doesn't need to wait on https://github.com/thebiggive/matchbot/pull/1304 , they're about two slightly separate things - this is about sign up before donating, 1304 is about signup after donating. They can be turned on separately.

PR #1953, opened 2025-04-17
ID-47: Remove feature flags and code for old password setting system

bdsl:
note PR target is `trunk` not `develop`, i.e. this is for deployment after GMF.


src/app/register/register.component.ts
```diff
@@ -273,7 +273,7 @@ export class RegisterComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 
   get readyToTakeAccountDetails(): boolean {
-    return !!this.verificationCodeSupplied || this.emailVerificationToken?.valid || ! flags.requireEmailVerification
+    return !!this.verificationCodeSupplied || !!this.emailVerificationToken?.valid
```
NoelLH, line 276: Curious whether there's a particular reason this needs the extra `!!` now? Seems like `this.emailVerificationToken` is typed strictly enough for it to be safe without, but I guess there's value for quick reading in making it similar to the left hand side?

NoelLH (approved):
LGTM. Always nice to see an avoidable modal go!

src/app/register/register.component.ts
bdsl: Without the extra `!!` there's a type error `Type boolean | undefined is not assignable to type boolean`

Previously it wasn't needed since on the left of a `||` undefined acts as false.

PR #1959, opened 2025-04-23
Tweak callout logs

NoelLH:
Give them their own group, to separate from `donate_error`s easily, and a less misleading identifier intro

### DIFF
--- a/src/app/logCalloutError.ts
+++ b/src/app/logCalloutError.ts
@@ -3,20 +3,20 @@ import {MatomoTracker} from 'ngx-matomo-client';
 export const logCalloutError = (
   isBrowser: boolean,
   context: string,
-  calloutUrl?: string,
+  calloutIdentifier?: string,
   matomoTracker?: MatomoTracker,
 ) => {
   if (!isBrowser) {
-    console.error('Server-side error in: ' + context + '. Callout URL: ' + calloutUrl);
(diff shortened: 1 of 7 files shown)


src/app/logCalloutError.ts
```diff
@@ -3,20 +3,20 @@ import {MatomoTracker} from 'ngx-matomo-client';
...
+    console.error('Server-side error in: ' + context + '. Callout identifier: ' + calloutIdentifier);
     return;
   }
 
-  console.error('Client-side error in: ' + context + '. Callout URL: ' + calloutUrl);
+  console.error('Client-side error in: ' + context + '. Callout identifier: ' + calloutIdentifier);
   if (matomoTracker) {
     matomoTracker.trackEvent(
-      'donate_error',
+      'campaign_error',
```
bdsl, line 17: not a blocker since it's no worse than what was there before, but I don't see how the `logCalloutError` function can know that this is a campaign related error. I think it's only the callers to this function that know if they're dealing with a campaign (which I can see they all are) so this probably should be passed in as an argument, or the function should be renamed to something like `logCampaignCalloutError`.

image

bdsl (approved):

src/app/logCalloutError.ts
NoelLH: Yep makes sense, renamed the fn in 83d0c2baf4f0c1fb2bffc9cdb171ce5e514125f6

PR #1963, opened 2025-04-27
Forms refactor

bdsl:


NoelLH (approved):
Nice. If I'm understanding it right, `formGroup` would be useful to keep in for bigger forms where we might want to validate a whole group in one go, but not particularly here.

bdsl:
@NoelLH 

> Nice. If I'm understanding it right, `formGroup` would be useful to keep in for bigger forms where we might want to validate a whole group in one go, but not particularly here.

Yes I think that's right, although in our big form I'm not sure we've actually used the formgroups as anything more than simple containers for controls (I don't know if you saw in the commits of this PR I started doing the same refactor on that, but reverted as fixing the type errors in the typescript was too much work for now)

PR #1965, opened 2025-04-29
Make code more pretty

bdsl:
Prettier installed and ran with all default settings except printWidth changed from default 80 to 120

Having more consistent formatting and no accidental indent level changes may particularly help if and when we have a new developer who needs to read this code for the first time.

I think the thing I most dislike about this is that it changes `<br>` to `<br />`, (and worse `<p>` to `<p />` if there is no closing tag, adding the slash that does nothing in html but makes the tag look like an self-closing XML element. Issue about it here https://github.com/prettier/prettier/issues/15336 .

Prettier doesn't like configurability so looks like we can't turn that off. Probably worth living with to the benefits of consistent formatting.


NoelLH (approved):
Spot checked a few changes in a few file types, LGTM.

What do you think about https://github.com/awmottaz/prettier-plugin-void-html, from the issue you linked, to work around the self-closing tags @bdsl?

bdsl:
> Spot checked a few changes in a few file types, LGTM.
> 
> What do you think about https://github.com/awmottaz/prettier-plugin-void-html, from the issue you linked, to work around the self-closing tags @bdsl?

Thanks @NoelLH. I don't have a strong feeling. I'm very happy to add that plugin if you'd like me to do it. or for you to add it. Otherwise I think I'll just stick with prettier defaults for simplicity. 

I checked the prettier playground and it's a bit better than I thought. It doesn't make `p` or `li` look like self-closing tags if we don't explicitly close them, it just adds an explicit closing tag in the right place:

[prettier playground example](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeADgPgDpQBJwBOcA5AM4AEAhhelYVQOYPoAWuuGO+RplVUCDFZFa9Ji1YAaCsICWlaHAqsq-CgBs5UZagBGhbgbhUA1tSgATCmQgBbZVrIwKc+HcrbXMAHQcoqFqYSpxBMADuEKFymMLEcCBSIBDoMHLQZMig9IQQ4QAK9AiZKFQa4VQAnplJBlRgpnAwAMpUDgAy2glIAGZlZHBJEHoAVnBgMADqLMgg6MQDhABuCbUMDU3NdGDajMgwhACugyADdnL7RydwAB7oRHIOsGUA8vcMMBCE+RBkbulQWYISyJEC3d6PBAwMoAFSIUHocjgJT6GgGST+UEYGjgAEVDkJuqj0SARmQbs1djj8YTkMSTgBHAnwfK5dAlEBqAC0Ojglj5oIOVDkWixAGF7HYqLMyhpQZjsXAAIIwA5yPSHFlETo6On9E6sGB2DSTVhuZHbODNYr-JZuSqzMBkGogJbHACSVihzTAhDkqSVVmaMEqOL1aJO81+cGmVHQs3myKIK1B2kWMFZTCl4ZJdEIi1mhuNoPm2imckswmQAA4AAxJYhMuTETOMbO9fVJaF6SYVqtIABMSUOAxhVD0KM7YLsej5-Ms7QEjEOTDgADEvlLVbsZZqICAAL4HoA)

NoelLH:
> Thanks @NoelLH. I don't have a strong feeling. I'm very happy to add that plugin if you'd like me to do it. or for you to add it. Otherwise I think I'll just stick with prettier defaults for simplicity.
> 
> I checked the prettier playground and it's a bit better than I thought. It doesn't make `p` or `li` look like self-closing tags if we don't explicitly close them, it just adds an explicit closing tag in the right place:
> 
> [prettier playground example](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeADgPgDpQBJwBOcA5AM4AEAhhelYVQOYPoAWuuGO+RplVUCDFZFa9Ji1YAaCsICWlaHAqsq-CgBs5UZagBGhbgbhUA1tSgATCmQgBbZVrIwKc+HcrbXMAHQcoqFqYSpxBMADuEKFymMLEcCBSIBDoMHLQZMig9IQQ4QAK9AiZKFQa4VQAnplJBlRgpnAwAMpUDgAy2glIAGZlZHBJEHoAVnBgMADqLMgg6MQDhABuCbUMDU3NdGDajMgwhACugyADdnL7RydwAB7oRHIOsGUA8vcMMBCE+RBkbulQWYISyJEC3d6PBAwMoAFSIUHocjgJT6GgGST+UEYGjgAEVDkJuqj0SARmQbs1djj8YTkMSTgBHAnwfK5dAlEBqAC0Ojglj5oIOVDkWixAGF7HYqLMyhpQZjsXAAIIwA5yPSHFlETo6On9E6sGB2DSTVhuZHbODNYr-JZuSqzMBkGogJbHACSVihzTAhDkqSVVmaMEqOL1aJO81+cGmVHQs3myKIK1B2kWMFZTCl4ZJdEIi1mhuNoPm2imckswmQAA4AAxJYhMuTETOMbO9fVJaF6SYVqtIABMSUOAxhVD0KM7YLsej5-Ms7QEjEOTDgADEvlLVbsZZqICAAL4HoA)

Cool, yeah I'm happy to leave it as-is for now and see whether we find it's ever doing anything too unhelpful in practice.

PR #1968, opened 2025-05-02
MAT-390: Only save payment method from donation if donor tells stripe…

bdsl:
… to save it

Previiously we passed `futureUsage: 'on_session'` to stripe when creating the elements for a donation, which overrode the customer's choice to not save by using a tickbox (and we have to do extra work to ensure we never present those methods for re-use).

Hopefull this will make stripe save the method for re-use only in the right cases.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -399,7 +399,7 @@
             }
             <div>
               @if (showCardReuseMessage) {
-                Your card will only be saved for reuse if you set a password on the next page.
+                Your card will only be saved for reuse if you tick the box above and create a donor account
               }
             </div>
           </div>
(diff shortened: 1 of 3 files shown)


NoelLH (dismissed):

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -399,7 +399,7 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
             }
             <div>
               @if (showCardReuseMessage) {
-                Your card will only be saved for reuse if you set a password on the next page.
+                Your card will only be saved for reuse if you tick the box above and create a donor account
```
NoelLH, line 402:
```suggestion
                Your card will only be saved for reuse if you tick the box above and create a donor account.
```

NoelLH (dismissed):

PR #1969, opened 2025-05-03
Linting etc

bdsl:


NoelLH (approved):

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -291,7 +294,7 @@ export class DonationStartFormComponent
 
       // ngx-matomo sets up window._paq internally, and doesn't have
       // A/B test methods, so we work with the global ourselves.
-      if (environment.matomoAbTest && globalThis.hasOwnProperty('_paq')) {
+      if (environment.matomoAbTest && Object.prototype.hasOwnProperty.call(globalThis, '_paq')) {
```
NoelLH, line 297: I had a quick look into options for this one. Maybe `hasOwn()` will be better eventually, but probably not wide enough support yet and not enough in our code to be worth a polyfill.

PR #1974, opened 2025-05-08
Don 521 a11y improvements

bdsl:
Previously a few pages (e.g. http://localhost:4200/transfer-funds and http://localhost:4200/explore ) did not set specific titles so we just showed 'Big Give' which is unhelpful in a list of tabs.

This commit makes sure every component has a title, and that the entry in our routes array explicitly declares a title entry (although it may be undefined) for every route.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -3,4 +3,4 @@
 # github does this automatically.
 
 # setup prettier
-273e57eb2adb571
+273e57eb2adb5710dddcd2737e566fa1c606b62b
--- a/src/app/app-routing.ts
+++ b/src/app/app-routing.ts
@@ -29,6 +29,7 @@ import { CancelMandateComponent } from './cancel-mandate/cancel-mandate.componen
 import { ChangeRegularGivingComponent } from './change-regular-giving/change-regular-giving.component';
 import { setupIntentResolver } from './setupIntent.resolver';
 import { EmailVerificationTokenResolver } from './email-verification-token.resolver';
+import { bigGiveName } from '../environments/common';
 
 export const registerPath = 'register';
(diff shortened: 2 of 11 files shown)


.git-blame-ignore-revs
```diff
@@ -3,4 +3,4 @@
 # github does this automatically.
 
 # setup prettier
-273e57eb2adb571
+273e57eb2adb5710dddcd2737e566fa1c606b62b
```
bdsl, line 6: previous version broke git blame, apparently we have to give the commit id in full

src/app/app-routing.ts
```diff
@@ -93,29 +95,39 @@ const handleLogout: CanActivateFn = () => {
   return inject(Router).parseUrl('/');
 };
 
-export const routes: Routes = [
+export const routes: (Route & {
+  // every route should either have a title explicitly defined here, or explicitly have an undefined
+  // title and rely on a call to PageMetaService.setCommon once loaded.
```
NoelLH, line 100: Very nice!

NoelLH (approved):
Looks good! Happy to see it doesn't take more code to make the fixed titles (or skipping explicitly) compulsory.

src/app/page-meta.service.ts
```diff
@@ -20,7 +21,7 @@ export class PageMetaService {
     private title: Title,
   ) {}
 
-  setCommon(title: string, description: string, imageUri: string | null) {
+  setCommon(title: string, description: string = '', imageUri: string | null = null) {
```
bdsl, line 24: IMHO meta descriptions have very limited value, and none when set client side. Not worth requiring ourselves to type in a description every time.

PR #1976, opened 2025-05-09
REG-31 – first batch of accessibility fixes

NoelLH:
Mostly focused on the Donate page, with small app-wide and campaign detail page improvements.


src/app/app.component.html
```diff
@@ -6,40 +6,45 @@ <h2>Unsupported Browser</h2>
...
-  <biggive-main-menu
-    [donateUrlPrefix]="donateUriPrefix"
-    [blogUrlPrefix]="blogUriPrefix"
-    [experienceUrlPrefix]="experienceUriPrefix"
-    [isLoggedIn]="isLoggedIn"
-    [smallCharityWeekEnabled]="true"
-    [someCampaignHasHomePageRedirect]="someCampaignHasHomePageRedirect()"
-  ></biggive-main-menu>
-}
+<div role="banner" aria-label="Menu &amp; skip">
```
NoelLH, line 10: The aim with this and similar containers below is to quickly fix errors from the skip link etc. not being inside a landmark. I didn't want to create nested `navigation` role landmarks so a catch-all of `banner` seemed the best quick win available.

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -18,16 +18,16 @@
...
-          <div>
-            <p class="b-rt-0 b-m-0">You are supporting:</p>
-            <p class="b-rt-0 b-m-0 b-bold">{{ campaign.charity.name }}</p>
-          </div>
-          <div>
-            <p class="b-rt-0 b-m-0">for:</p>
-            <p class="b-rt-0 b-m-0 b-bold">{{ campaign.title }}</p>
-          </div>
-        </div>
+        <h1 class="donationInfoWrapper">
```
NoelLH, line 21: I started with a separate visually hidden `h1`, but ultimately this is doing that job already. I think slightly clunky markup to allow for varying styles within the title is a smaller cost than maintaining two different elements.

src/app/app.component.html
bdsl: Looks good, I didn't know much about landmarks but looks like you're following this rule https://dequeuniversity.com/rules/axe/4.1/region 

Would it make sense to use a `<header>` instead of `<div>`? 

Not sure if the aria-label is accessary, I guess menu & skip is what people expect to find inside the banner anyway, might just add to listening time?

src/app/app.component.html
```diff
@@ -6,40 +6,45 @@ <h2>Unsupported Browser</h2>
...
 
-<biggive-footer
-  [headingLevel]="3"
-  [usePresetFooter]="true"
-  [donateUrlPrefix]="donateUriPrefix"
-  [blogUrlPrefix]="blogUriPrefix"
-  [experienceUrlPrefix]="experienceUriPrefix"
-  [smallCharityWeekEnabled]="true"
-></biggive-footer>
+<div role="contentinfo">
```
bdsl, line 38: could this be a <footer> element instead?

Also maybe something to do in components library instead of here so it can benefit the wordpress site at the same time?

bdsl:
added comments line

src/app/app.component.html
```diff
@@ -53,7 +58,7 @@ <h2>Unsupported Browser</h2>
     isPlatformBrowser &&
     (userHasExpressedCookiePreference$ | async) === false)
 ) {
-  <div id="cookie-banner">
+  <div role="banner" aria-label="Cookie preference" id="cookie-banner">
```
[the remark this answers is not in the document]
bdsl: MDN says `aria-labelledby` is preferable to `aria-label` but makes sense that we can't use the former here since the element we'd have to refer to is inside the component and we don't have its ID visible here.

But then since we show the same banner in wordpress maybe we should move this role marker to inside the component, and then we could use `aria-labelledby` to refer to the element that says "Our website uses cookies"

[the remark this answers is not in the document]
NoelLH: Yes you're right about the role. Do you think `complementary` implies too much of a connection to the specific page's content?

I think if we try to house everything in components we will have the same problem above that the tools still want custom components themselves to be inside landmarks.

src/app/app.component.html
NoelLH: I think some kind of distinguishing label is compulsory when there is more than one of the same landmark on the page, which there will be for `banner`.

Using `<header>` instead sounds like a good idea, changed in 316fb1e33623adbd25d95211d9af8334276594c4

src/app/app.component.html
```diff
@@ -6,40 +6,45 @@ <h2>Unsupported Browser</h2>
...
-  <biggive-main-menu
-    [donateUrlPrefix]="donateUriPrefix"
-    [blogUrlPrefix]="blogUriPrefix"
-    [experienceUrlPrefix]="experienceUriPrefix"
-    [isLoggedIn]="isLoggedIn"
-    [smallCharityWeekEnabled]="true"
-    [someCampaignHasHomePageRedirect]="someCampaignHasHomePageRedirect()"
-  ></biggive-main-menu>
-}
+<header aria-label="Menu &amp; skip">
```
bdsl, line 10: I think the label is probably superfluous here but not a blocker

src/app/app.component.html
```diff
@@ -53,7 +58,7 @@ <h2>Unsupported Browser</h2>
     isPlatformBrowser &&
     (userHasExpressedCookiePreference$ | async) === false)
 ) {
-  <div id="cookie-banner">
+  <div role="region" aria-label="Cookie preference" id="cookie-banner">
```
bdsl, line 61:
```suggestion
  <div role="region" aria-label="Our website uses cookies" id="cookie-banner">
```

bdsl (approved):
LGTM

src/app/app.component.html
NoelLH: Yeah I think I'm inclined to leave this one for now so long as Axe seems happy with the updated version in other contexts like regression tests too. I agree it might make sense to give the inner `components` element a more general element and leave landmarks to the parent apps, if we work on it again.

src/app/app.component.html
NoelLH: Yep agreed, I was testing this because I got worried something about the container was affecting the skip link's keyboard focusability. Seems like I did something wrong or it was an unrelated WIP change.

PR #1978, opened 2025-05-09
DON-1104: Add big red 'offline' banner to bottom of donate page if us…

bdsl:
…er's device disconnects from network

image


bdsl:
For this white on red text, using our custom red which is a bit darker than default red, the webaim contrast checker shows there is enough contrast except for normal text at accessibility level AAA. I think that's OK since this is larger than standard text and only a single line to read: https://webaim.org/resources/contrastchecker/

image

PR #1981, opened 2025-05-12
DON-1104: Allow retrying setting up donation and stripe connection in…

bdsl:
… case of unreliable connection

New retry button replaces continue in case we couldn't set up the pending donation:

image

PR also includes new handling to allow retrying stripe service initialisation, and require it to be initialised before progress to step 2 in donation form.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -613,6 +613,18 @@
               if this message persists.
             </p>
           }
+          @if (finalPreSubmitUpdateFailed) {
+            <button
+              style="width: 40%"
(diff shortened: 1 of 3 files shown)


NoelLH (approved):
LGTM.

Possibly worth a Matomo log to tell how often people are using the retry button in particular after facing problems? But I see we already have one for when problems materialise and we can probably to some extent infer from this plus later successful events. Not a blocker.

bdsl:
> LGTM.
> 
> Possibly worth a Matomo log to tell how often people are using the retry button in particular after facing problems? But I see we already have one for when problems materialise and we can probably to some extent infer from this plus later successful events. Not a blocker.

Thanks, will add in a separate commit.

PR #1989, opened 2025-05-22
MAT-411: Use matchbot campaign api instead of salesforce to load donate pages in dev env

bdsl:
FE (and readers of its code) should know when its talking to SF and when its not, since that is about to change.


smalizaidi (approved):
Looks good to me.

NoelLH (commented):
Reviewed late, LGTM. I mostly focused on the bits of CampaignService not superseded by the newer PR I reviewed today.

PR #1994, opened 2025-05-27
MAT-405: Start distinguishing between Charity Campaigns (aka Campaign…

bdsl:
…s) and MetaCampaigns

Matchbot will likely serve these as distinct models with distinct API
routes, and we will be able to start separating out their fields

### DIFF
--- a/src/app/campaign.model.ts
+++ b/src/app/campaign.model.ts
@@ -1,7 +1,7 @@
 /**
  * @link https://app.swaggerhub.com/apis/Noel/TBG-Campaigns/#/Campaign
  */
-export type CampaignOrMetaCampaign = {
+type CampaignOrMetaCampaign = {
   id: string;
   aims: string[];
   amountRaised: number;


src/app/campaign.model.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * @link https://app.swaggerhub.com/apis/Noel/TBG-Campaigns/#/Campaign
  */
-export type Campaign = {
+export type CampaignOrMetaCampaign = {
```
bdsl, line 4: not exported, so code in other files has to distinguish whether it wants a charity campaign (aka simply a campaign) or a metacampaign.

We do have at least one routes that can technically be used for both, but I don't think in practice we've ever encouraged requesting a metacampaign by its id instead of it slug.

bdsl:
@NoelLH do you think it's right that we never encouraged people to view metacampaigns by id, and it will be OK to break that use case?

Afaik charity campaigns never have slugs so there isn't anything to break there.

NoelLH:
> @NoelLH do you think it's right that we never encouraged people to view metacampaigns by id, and it will be OK to break that use case?
> 
> Afaik charity campaigns never have slugs so there isn't anything to break there.

Yes, I had a quick think about both sides of that and I don't think this can break anything we expect to be useful. Even accessing meta-campaign at `{base URI}/{sf ID}` was already not an option, so I think it really is one or the other the way we currently offer the URLs.

PR #1995, opened 2025-05-28
MAT-405: Adjust Campaign / MetaCampaign status type to better reflect…

bdsl:
… what SF can send

Campaigns that are not yet approved to launch on the BG site have null status, and should not be findable by the public but may be previewed by the people creating them.


smalizaidi (approved):
LGTM

PR #2000, opened 2025-05-30
MAT-405: Fully separate (charity) campaign and metacampaign types

bdsl:
Should be type-level only changes, but will help us see what's required as we implement serving campaigns and metacampaigns separately from matchbot


src/app/campaign.model.ts
```diff
@@ -1,12 +1,35 @@
 /**
  * @link https://app.swaggerhub.com/apis/Noel/TBG-Campaigns/#/Campaign
+ *
+ * Represents an individual campaign to raise funds for one charity - distinct from a metacampaign
+ * which has its own model
  */
-type CampaignOrMetaCampaign = {
```
bdsl, line 4: losing the shared base type, I think explicitly repeating the properties that are used in both types is clearer than having a shared base type, since they're more different than similar.

NoelLH (approved):

PR #2020, opened 2025-06-24
DON-1148 – refresh Payment Element when possibly necessary

NoelLH:
Aiming to fix rare edge cases where the choice to setup the payment method for future usage or not is 'null' after an initial decline.

Broadly I'm trying to call the fetch fn whenever:
1. there was just a decline, and
2. we might be about to reconfigure or recycle an Element that wasn't fully destroyed, perhaps in a recovery edge case we haven't specifically thought about


bdsl (approved):
LGTM

PR #2026, opened 2025-07-01
DON-1150: Reduce two buttons in regular giving stepper to 40% width

bdsl:
We had a mix of full width and 40% width buttons - making them all 40% width.

There's potential to refactor to reduce the duplication but not necassarily right now.

image

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -166,7 +166,6 @@
                 if this message persists.
               </p>
               <button
-                style="width: 40%"
                 type="button"
                 class="continue b-w-100 b-rt-0"
                 mat-raised-button
@@ -177,9 +176,8 @@
               </button>
             } @else {
               <button
-                style="width: 40%"
                 type="button"
-                class="continue b-w-100 b-rt-0"
+                class="continue"
                 mat-raised-button
                 color="primary"
                 (click)="progressToNonAmountsStep()"
@@ -309,7 +307,7 @@
           <div style="text-align: center">
             <button
               type="button"
-              class="continue b-w-100 b-rt-0"
+              class="continue"
               mat-raised-button
               color="primary"
               (click)="progressFromStepGiftAid()"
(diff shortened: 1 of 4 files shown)


NoelLH (dismissed):
I think that using inline styles exacerbates the tendency to get confused about what is and isn't using the same style and create more problems like this in the future.

Donation start form component for example is already doing a similar (but not guaranteed to stay the same) thing for two sets of buttons. Stepper progressions are (sometimes?) using inline style like this, while the final button uses a class which sets a max-width of 300px. So it is already really time consuming to work out whether and why buttons look the same width.

Not a blocker but I think this would be a lot less painful if we use CSS classes more consistently.

bdsl:
> I think that using inline styles exacerbates the tendency to get confused about what is and isn't using the same style and create more problems like this in the future.
> 
> Donation start form component for example is already doing a similar (but not guaranteed to stay the same) thing for two sets of buttons. Stepper progressions are (sometimes?) using inline style like this, while the final button uses a class which sets a max-width of 300px. So it is already really time consuming to work out whether and why buttons look the same width.
> 
> Not a blocker but I think this would be a lot less painful if we use CSS classes more consistently.

Thanks, fair point. I think I was considering refactoring by creating a new Angular component (maybe because that's what I'd just been looking at in the Angular course) and it seemed like slightly too much effort, but just applying a CSS class is a lot simpler. I'll give that a go and ask for a new review.

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -177,9 +176,8 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
               </button>
             } @else {
               <button
-                style="width: 40%"
                 type="button"
-                class="continue b-w-100 b-rt-0"
```
bdsl, line 182: w-100 meant width 100%, was overridden by the inline style so deleted. I used scss `@extend` to do the `b-rt-0` via the continue class.

NoelLH (approved):

PR #2027, opened 2025-07-01
DON-1134: Improve UI for making unmatched regular giving donation

bdsl:
I think a checkbox inside a list doesn't really work well, so it makes more sense to make the two checkboxes (aged over 18 and making unmatched donation) sit together one after the other, and just use normal running copy to describe the error.

Also making the error appear and disappear reactively when the donation amount is reduced above or below the matchable maximum, and when unmatched option is clicked.

### DIFF
--- a/src/app/regular-giving/regular-giving.component.html
+++ b/src/app/regular-giving/regular-giving.component.html
@@ -61,6 +61,18 @@
                   <input maxlength="10" formControlName="donationAmount" id="donationAmount" matInput slot="input" />
                 </biggive-text-input>
               </div>
+
+              @if (showUnmatchedDonationOption) {
+                <mat-checkbox formControlName="unmatched" id="unmatched">Make an unmatched donation</mat-checkbox>
+                <br />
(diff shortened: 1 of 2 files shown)


src/app/regular-giving/regular-giving.component.html
```diff
@@ -76,39 +88,24 @@
                     {{ this.amountErrorMessage }}
                   </div>
                 }
-                @if (maximumMatchableDonation.amountInPence === 0) {
+                <!--                insufficientMatchFundsAvailable: {{ insufficientMatchFundsAvailable }},-->
+                <!--                unmatched: {{ insufficientMatchFundsAvailable }},-->
+                <!--                newDonationAmountOverMaxMatchable: {{newDonationAmountOverMaxMatchable}}-->
+
```
NoelLH, line 94:
```suggestion
```

NoelLH (approved):
Suggested removing some debug out, not essential. All LGTM

PR #2031, opened 2025-07-03
MAT-405: Provide dedicated route URL paths for early previews of camp…

bdsl:
…aigns

Path such as /campaign-preview/a05xxxxxxxxxxxxxxx

Intended specifically for use by charity representatives to see a campaign as they edit it - for linking to from the charity portal. Will be the only way to see a campaign that is saved into the SF database but not yet existing in the matchbot database.


NoelLH (commented):
Apart from prettier checks, LGTM

PR #2032, opened 2025-07-03
MAT-405: Clarify whether campaign page being viewed is an early / pri…

bdsl:
…vate preview

Previews are intended for charity use only, not for public viewing, but there are currently no technical limits to stop the public seeing them.

So for now on a public preview we use a snackbar message, 'PREVIEW' in the html title and in the heading on the page, and noIndex for robots to discourage sharing with the public.

We could also add a link to the public page (same URL bout without '-preview', but that will only work when the campaign is ready. Not sure if it's better to do that here or in the charity portal.

Also trying out using signals and route data input binding a bit more.

### DIFF
--- a/src/app/campaign-details/campaign-details.component.html
+++ b/src/app/campaign-details/campaign-details.component.html
@@ -53,7 +53,7 @@
         html-element="h1"
         size="3"
         align="left"
-        [text]="campaign.title"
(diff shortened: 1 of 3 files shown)


NoelLH (approved):
Suggested a typo fix, otherwise all LGTM

PR #2035, opened 2025-07-08
MAT-413 – start using MatchBot's campaign search when new feature flag `useMatchbotCampaignSearchApi` is on

NoelLH:


src/app/campaign.service.ts
```diff
@@ -216,7 +208,12 @@ export class CampaignService {
       params = params.append('term', searchQuery.term);
     }
 
-    return this.http.get<CampaignSummary[]>(`${environment.sfApiUriPrefix}${this.apiPath}/campaigns`, { params });
+    switch (flags.useMatchbotCampaignApi) {
+      case false:
+        return this.http.get<CampaignSummary[]>(`${environment.sfApiUriPrefix}${this.apiPath}/campaigns`, {params});
+      case true:
+        return this.http.get<CampaignSummary[]>(`${environment.matchbotApiPrefix}/campaigns`, {params});
```
bdsl, line 215: maybe worth making the matchbot api return an object containing a list of campaign summaries now to allow room for future expansion, since I think once this is on in prod it will be much harder to change? With all the new APIs like this I've been piping and mapping the responses.

For search I guess we might well want stuff like total counts and info about any meta-campaign in addition to the list of summaries.

NoelLH: Yeah, I think it's worth it for the flexibility. I'm not about those extra properties right now, but it's true that we are already making slightly breaking changes and could well want to add something else in future.

PR #2041, opened 2025-07-14
Don 1134: Don't show message about limited match funds when at least 1500 available

bdsl:
If the max we can match based on match funds is more than the max we allow per donation we don't need to tell the donor that there are limited match funds as it isn't relevant to their donation.

PR also includes a change discussed in retro to display error messages on staging.


NoelLH (approved):
Suggested a comment update that's outside the diff but shown on the PR. Not a blocker, LGTM

PR #2043, opened 2025-07-15
DON-1120: Take all properties of custom metacampaign banner layout fr…

bdsl:
…om matchbot

For now just moving from hard-coding here to hard-coding in matchbot, but it's a step towards having them entered in SF.

### DIFF
--- a/src/app/explore/explore.component.html
+++ b/src/app/explore/explore.component.html
@@ -17,12 +17,19 @@
     </biggive-totalizer>
   }
 
-  @if (metaCampaign && metaCampaign.useDon1120Banner) {
+  @if (metaCampaign && !!metaCampaign.bannerLayout) {
     <app-meta-campaign-banner
       [logo]="fund?.logoUri ? { url: fund?.logoUri!, alt: fund?.name } : undefined"
       [mainTitle]="title"
       [teaser]="fund && fund.description ? fund.description : metaCampaign.summary"
       [mainImageUrl]="metaCampaign.bannerUri"
+      [backgroundColour]="metaCampaign.bannerLayout.backgroundColour"
+      [focalPoint]="{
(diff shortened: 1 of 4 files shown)


src/app/explore/explore.component.html
```diff
@@ -17,12 +17,19 @@
...
+  @if (metaCampaign && !!metaCampaign.bannerLayout) {
     <app-meta-campaign-banner
       [logo]="fund?.logoUri ? { url: fund?.logoUri!, alt: fund?.name } : undefined"
       [mainTitle]="title"
       [teaser]="fund && fund.description ? fund.description : metaCampaign.summary"
       [mainImageUrl]="metaCampaign.bannerUri"
+      [backgroundColour]="metaCampaign.bannerLayout.backgroundColour"
+      [focalPoint]="{
+        x: metaCampaign.bannerLayout.focalArea.topLeftXpos,
+        y: metaCampaign.bannerLayout.focalArea.topLeftYpos,
```
bdsl, line 29: for now just taking the top left position, using the bottom right as well is for the future but I'm not sure if we'll actually need it - it depends whether the size of the box will actually vary enough between metacampaigns to matter.

bdsl: might have been better to pass the entire bannerLayout object into the new component as a single input. I don't think that should be a blocker though.

src/app/metaCampaign.model.ts
```diff
@@ -42,6 +42,15 @@ export type MetaCampaign = {
   campaignCount?: number;
   usesSharedFunds: boolean;
 
-  /** Whether to apply the new banner layout. Treat undefined as false. */
-  useDon1120Banner?: boolean;
+  bannerLayout?: {
```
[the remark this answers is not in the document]
NoelLH: Could be good, but doesn't seem critical as long as it goes out before an important metacampaign time

PR #2045, opened 2025-07-17
Don 1120 - more metacampaign rectangle banner improvements

bdsl:
https://github.com/user-attachments/assets/4f8cd86b-c682-440c-8beb-35cd7cc91882

### DIFF
--- a/src/app/explore/meta-campaign-banner/meta-campaign-banner.component.scss
+++ b/src/app/explore/meta-campaign-banner/meta-campaign-banner.component.scss
@@ -36,11 +36,12 @@ div.sleeve {
   box-sizing: border-box;
   height: 100%;
 
-  @media #{abstract.$breakpoint-smallish} {
+  @media (min-width: #{abstract.$smallish}) {
     flex-direction: column;
   }
 
-  @media ("min-width: 480px") {
+  @media (min-width: #{abstract.$smallish + 30px}) {
+    // adding 30px to standard breakpoint to match padding.
     padding-left: 20px;
     align-items: center;
     height: 600px;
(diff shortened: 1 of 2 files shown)


src/app/explore/meta-campaign-banner/meta-campaign-banner.component.scss
```diff
@@ -9,31 +11,72 @@ div.banner {
...
   max-width: 1200px;
-  height: 600px;
   margin-left: auto;
   margin-right: auto;
-  padding: 0 20px;
+  padding: 0;
   box-sizing: border-box;
+  height: 100%;
+
+  @media #{abstract.$breakpoint-smallish} {
```
NoelLH, line 39: I'm a bit confused about using this var for one change but also a similar but distinct breakpoint (480px vs 450px) through the rest of this diff.

Could we pick one size and use a var consistently? I think it's quite helpful to be able to see the app's common breakpoints of interest in one place. I think it would also be fine to redefine `$breakpoint-smallish` to 480px for simplicity; it would just mean that a small number of additional mobile sizes might see the 'Donation to...' label on the donate form over 2 lines.

src/app/explore/meta-campaign-banner/meta-campaign-banner.component.html
```diff
@@ -40,5 +30,14 @@
       <h1 class="main-title">{{ mainTitle() }}</h1>
       <div class="teaser">{{ teaser() }}</div>
     </div>
+    <!-- Stacked image below only for narrow screens -->
+    <img
+      class="stacked"
+      [src]="mainImageOptimisedUri.value()"
+      alt=""
+      [ngStyle]="{
+        'object-position': focalPoint().x + '% ' + focalPoint().y + '%',
```
NoelLH, line 39: Nice!

NoelLH (commented):
I think we can make the breakpoint rules clearer & more consistent, otherwise LGTM

src/app/explore/meta-campaign-banner/meta-campaign-banner.component.scss
NoelLH: Cool, yeah that sounds good, I don't think it's very obvious that they're deliberately couple right now.

PR #2048, opened 2025-07-18
DON-1127 – enable Regular Giving in Production

NoelLH:


bdsl:
> This is one of the todos in frontend, probably should complete before turning this on:
> image

@NoelLH worth reviewing the other todo-regular-giving comments in this repo - see https://github.com/search?q=repo%3Athebiggive%2Fdonate-frontend%20todo-regular-giving&type=code . I'm not immediately sure if any of them should be blockers for this.

src/app/my-regular-giving/my-regular-giving.component.html
```diff
@@ -48,11 +48,9 @@ <h3>
                 </h3>
                 <ul>
                   <li>Amount: {{ mandate.donationAmount | money }}</li>
-                  <!-- @todo-regular-giving consider whether to also show total with match
-                              funds when the next donation will be matched. -->
```
NoelLH, line 52: I think as this page has been through fairly detailed QA now we can probably assume that the figures shown are enough now

PR #2053, opened 2025-07-24
Don-1120 home and explore pages

bdsl:
Sets up the new rectangular style banner for the home and explore pages - feature flagged to only apply outside production for now.


src/app/heading-banner/heading-banner.component.ts
```diff
@@ -12,14 +12,17 @@ import { firstValueFrom } from 'rxjs';
...
-  styleUrl: './meta-campaign-banner.component.scss',
+  templateUrl: './heading-banner.component.html',
+  styleUrl: './heading-banner.component.scss',
 })
-export class MetaCampaignBannerComponent {
+export class HeadingBanner {
   imageService = inject(ImageService);
   logo = input<{ url: string; alt: string | undefined } | undefined>();
+  /** Optional slightly smaller text to appear above the main title */
+  slug = input('');
```
bdsl, line 24: I don't really like the name slug, I would expect a slug to be a URL component, but slug matches what we're using on the old component and I can't immediately think of anything better.

PR #2067, opened 2025-07-31
Fix stray closing tag

NoelLH:


bdsl:
@NoelLH I just did the same thing on develop

PR #2076, opened 2025-08-07
Revert "MAT-405: Switch to using Matchbot not SF for metacampaign and search …"

bdsl:
Reverts thebiggive/donate-frontend#2074

We found a couple of discrepancies between the old Salesforce search implementation and the new one built in Matchbot.


bdsl:
Thanks @DomTBG - but I'm now hoping that https://github.com/thebiggive/matchbot/pull/1554 which is on its way to prod will fix everything we need fixed and we don't need to do this revert. Closing the PR but we can re-open and mrege if needed.

PR #2079, opened 2025-08-12
New homepage content

bdsl:
Before:

image

After:

image

As requested by Alex via @BethBigGive


bdsl:
> > I guess this will require regression tests' SSR expected copy to be changed back around the same time
> 
> yep I forgot that, I'll do that too if this is approved.

changed here to check content that isn't changing: https://github.com/thebiggive/regression-tests/commit/8879910c4441af96349a9e620969421eefa74a68

PR #2083, opened 2025-08-14
DON-1107: Adjust campaign detail info display for shared funds (i.e. …

bdsl:
…emergency & regular-giving)

- Show "Amount raised" for this charity campaign instead of the overall total for the parent.
- Don't show any progress bar
- Show "Total Match Funds Available" using the parent match funds available figure, instead of the overall target

image

### DIFF
--- a/src/app/campaign.service.spec.ts
+++ b/src/app/campaign.service.spec.ts
@@ -163,7 +163,7 @@ describe('CampaignService', () => {
     expect(CampaignService.percentRaisedOfCampaign(campaign)).toBe(49);
   });
 
-  it('should return undefined its parent does use shared funds', () => {
+  it('should return undefined when its parent does use shared funds', () => {
     const campaign = getDummyCampaign();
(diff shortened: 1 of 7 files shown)


src/app/campaign.service.spec.ts
```diff
@@ -160,10 +160,10 @@ describe('CampaignService', () => {
     campaign.amountRaised = 98;
     campaign.target = 200;
 
-    expect(CampaignService.percentRaisedOfCampaignOrParent(campaign)).toBe(49);
+    expect(CampaignService.percentRaisedOfCampaign(campaign)).toBe(49);
   });
 
-  it('should return the % raised for the parent campaign when its parent does use shared funds', () => {
+  it('should return undefined its parent does use shared funds', () => {
```
NoelLH, line 166:
```suggestion
  it('should return undefined when its parent does use shared funds', () => {
```

NoelLH (approved):

NoelLH (approved):

PR #2085, opened 2025-08-15
MAT-429 – Support Stripe Pay By Bank

NoelLH:


src/app/donation-thanks/donation-thanks.component.html
```diff
@@ -12,6 +12,12 @@
     ></mat-spinner>
   }
 
+  @if (payByBankPossibleError && donation) {
+    <p aria-live="assertive">
+      Payment may have not have been completed. <a routerLink="/donate/{{ donation.projectId }}">Try again?</a>
```
bdsl, line 17: this seems like a concerning message to get as a donor - if you can't tell me for sure that the payment has failed I don't think I'd want to try again.

bdsl (commented):
Added one comment inline, otherwise all LGTM

bdsl (approved):

src/app/donation-thanks/donation-thanks.component.html
NoelLH: Yep, thanks. Done in c55734855671f73945691b4260a5f1b0bd27f7f5

PR #2090, opened 2025-08-19
Refactor: remove all feature flags - now all been on in all envs for …

bdsl:
…some time


bdsl:
SF api is now only used for Campaign Impact Stats, funds, and homepage highlight cards:

image

PR #2093, opened 2025-08-20
DON-1107: Shift amount raised for shared-funds campaigns on explore p…

bdsl:
…age to left

Based on new comment on ticket.

Also want to shift it up some pixels, but that will need to be done in a components library change.

**Before**:

image

**After**:

image

### DIFF
--- a/src/app/explore/explore.component.html
+++ b/src/app/explore/explore.component.html
@@ -88,51 +88,65 @@
         >
           @for (campaign of individualCampaigns; track campaign.id) {
             <div>
-              <!-- note the primary figure amount below in case of regular giving has to be a non-empty string to
+              <!-- note the secondary figure amount below in case of regular giving / shared-funds has to be a non-empty string to
                   make the component
                        reserve space for it to keep the layout consistent. Once we're happy with this layout
-                       consider editing component code to allow this to happen in a more intuitive way
+                       consider editing component code to allow this to happen in a more intuitive way.
 
-                       I think as of Feb 2025 the logic for campaign cards in general has got quite confusing, with too many
-                       differences between how the card appears on the metacampaign, charity, and explore pages. May be worth trying to rework subsantially so we can encapsulate
-                       and replace the tag below with a simple <biggive-campaign-card [campaign]="campaign">. Would require perhaps some changes at SF so we don't have to pass
-                       the metacmpaign, or processing the campaign data in FE before it gets to here. Could also consider moving the `campaign-card` component out of
-                       components library and into this repo for to make it easier to work on if we're not likely to use it anywhere else.
-
-                       Logic for primary figure amount is particularly confusing - we have to pass null here if the campaign is an emergency match fund campaign
-                       as it doesn't make sense to display any primary figure or secondary figure, but the only way to know in FE at present that it's an emergency campaign
-                       is by the fact that it uses match funds and *isn't* regular giving. If it is regular giving we pass a space to take up space in the layout without being visible.
+                       In practice the `|| campaign.isRegularGiving` should be redundant everywhere except on dev machines,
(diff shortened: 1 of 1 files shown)


src/app/explore/explore.component.html
```diff
@@ -88,51 +88,65 @@
...
+                    campaign.isRegularGiving ? '/regular-giving/' + campaign.id : '/donate/' + campaign.id
+                  "
+                  [donateButtonLabel]="campaign.isRegularGiving ? 'Donate monthly' : 'Donate now'"
+                  [moreInfoButtonUrl]="'/campaign/' + campaign.id"
+                  [organisationName]="campaign.charity.name"
+                  [primaryFigureLabel]="'Amount Raised'"
+                  [primaryFigureAmount]="
+                    campaign.amountRaised | currency: campaign.currencyCode : 'symbol' : currencyPipeDigitsInfo
+                  "
+                  [progressBarCounter]="undefined"
```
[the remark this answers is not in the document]
DomTBG: Other than the fact that the Amount Raised is not horizontally aligned yet (which as Barney mentioned above needs to be done in the components library), this change looks good! Doesn't look like I am able to approve this PR though...

src/app/explore/explore.component.html
```diff
@@ -88,51 +88,65 @@
...
-                secondary-figure-label="Amount Raised"
-                [secondaryFigureAmount]="
-                  campaign.amountRaised | currency: campaign.currencyCode : 'symbol' : currencyPipeDigitsInfo
-                "
-                [isFutureCampaign]="isInFuture(campaign)"
-                [isPastCampaign]="isInPast(campaign)"
-                [datetime]="getRelevantDateAsStr(campaign)"
-              >
-              </biggive-campaign-card>
+              @if (campaign.parentUsesSharedFunds || campaign.isRegularGiving) {
```
NoelLH, line 99: I think optionally we could simplify this because the fund setup for Regular Giving defines it as shared. Not a blocker.

```suggestion
              @if (campaign.parentUsesSharedFunds) {
```

src/app/explore/explore.component.html
```diff
@@ -88,51 +88,65 @@
...
+                  [donateButtonLabel]="campaign.isRegularGiving ? 'Donate monthly' : 'Donate now'"
+                  [moreInfoButtonUrl]="'/campaign/' + campaign.id"
+                  [organisationName]="campaign.charity.name"
+                  [primaryFigureLabel]="'Amount Raised'"
+                  [primaryFigureAmount]="
+                    campaign.amountRaised | currency: campaign.currencyCode : 'symbol' : currencyPipeDigitsInfo
+                  "
+                  [progressBarCounter]="undefined"
+                  secondary-figure-label=""
+                  [secondaryFigureAmount]="' '"
```
NoelLH, line 117:
```suggestion
                  secondary-figure-amount=" "
```

NoelLH (approved):
Minor code legibility ideas inline, no blockers, LGTM

src/app/explore/explore.component.html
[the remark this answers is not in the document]
bdsl: Thanks @DomTBG. You should be able to approve, can take a look on a call with you to see why you can't if you like, but really approving with words is as good as approving formally.

src/app/explore/explore.component.html
bdsl: Not sure - In general yes the second form without the property binding is simpler and I'd prefer it, but I think in this case where almost everything else has to be property-bound it makes for easier reading when we use property binding on this too as I had done.

src/app/explore/explore.component.html
NoelLH: Right, sorry I missed that in unified view as the delete was more prominent

PR #2108, opened 2025-09-02
DON-984: Publish link from robots.txt to sitemap

bdsl:
See https://developers.google.com/search/docs/crawling-indexing/robots/create-robots-txt

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -101,10 +101,12 @@ app.use(morgan('combined')); // Log requests to stdout in Apache-like format
  */
 app.get('/robots.txt', (_req: Request, res: Response) => {
   res.type('text/plain');
+  const sitemapLine = `Sitemap: ${environment.matchbotApiOrigin}/sitemap`;
+
   if (environment.production) {
-    res.send('User-agent: *\nAllow: /');
+    res.send('User-agent: *\nAllow: /\n\n' + sitemapLine);
   } else {
-    res.send('User-agent: *\nDisallow: /');
+    res.send('User-agent: *\nDisallow: /\n\n' + sitemapLine);
   }
 });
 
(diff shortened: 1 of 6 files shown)


src/server.ts
```diff
@@ -101,10 +101,12 @@ app.use(morgan('combined')); // Log requests to stdout in Apache-like format
...
 app.get('/robots.txt', (_req: Request, res: Response) => {
   res.type('text/plain');
+  const sitemapLine = `Sitemap: ${environment.matchbotApiOrigin}/sitemap`;
+
   if (environment.production) {
-    res.send('User-agent: *\nAllow: /');
+    res.send('User-agent: *\nAllow: /\n\n' + sitemapLine);
   } else {
-    res.send('User-agent: *\nDisallow: /');
+    res.send('User-agent: *\nDisallow: /\n\n' + sitemapLine);
```
NoelLH, line 109: Is there a particular advantage to serving the sitemap in envs where we want to discourage bots? It seems like it could slightly increase the chance of a confused or bad bot scanning a lot of pages.

NoelLH (commented):
Aside from maybe the `Disallow` case, LGTM

src/server.ts
bdsl: @NoelLH It's somewhat useful for the same reason we do everything else in those envs - to let us check things work before we push to prod or with different data than what we have in prod. E.g. in this case we can check that the robots.txt looks OK and that the sitemap URI is correct. On the other hand if this is broken in prod it's not a huge issue, and its not like we're going to actually review the sitemap with every change before we push to prod.

I can take it out, I think probably slightly worth keeping in, and worth keeping at least initially before we've seen this work in prod.

NoelLH: We know the path to the sitemap in MatchBot though, so all this lets us additionally check is Donate's _reference_ to the sitemap which I can't imagine we're going to change much

PR #2135, opened 2025-09-17
DON-1152: Fix ES5 build for older chrome versions

bdsl:
Implemented using junie, prompts:

> Our approach for serving a more-compatible but larger ES5 build for older browsers is
> coming along ok, but babel alone is not enough because e.g. Chrome 73 cannot run the
> require() etc. that are in the output. Could you propose a minimal Webpack configuration
> that will output something we can serve to such browsers in the browser-es5 folder?

---

> Please run `scripts/create-legacy-build.js` and fix the errors about `Requests that should resolve in the current directory need to start with './'.`.

> Output currently includes many lines like so:

> ```

> ERROR in reset-password.component-6S6KTTMX
> Module not found: Error: Can't resolve 'dist/donate-frontend/browser/reset-password.component-6S6KTTMX.js' in '/home/barney/projects/donate-frontend'
> Did you mean './dist/donate-frontend/browser/reset-password.component-6S6KTTMX.js'?
> Requests that should resolve in the current directory need to start with './'.
> Requests that start with a name are treated as module requests and resolve within module directories (node_modules).
> If changing the source code is not an option there is also a resolve options called 'preferRelative' which tries to resolve these kind of requests in the current directory too.

```

### DIFF
--- a/docs/es5-build.md
+++ b/docs/es5-build.md
@@ -1,0 +1,96 @@
+# ES5 Build for Legacy Browsers
+
+This document explains how we create a more compatible ES5 build for older browsers like Chrome 73 that cannot run modern JavaScript features or module syntax like `require()`.
+
+## Overview
+
+The ES5 build process uses a combination of Webpack and Babel to:
+
+1. Bundle JavaScript modules into browser-compatible code (removing `require()` statements)
+2. Transpile modern JavaScript syntax to ES5-compatible code
+3. Serve the resulting files from the `/d-es5/` path for legacy browsers
+
+## Files Involved
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: docs/es5-build.md [hunk 1]
```diff
@@ -0,0 +1,99 @@
...
+1. Create the ES5 build in `dist/donate-frontend/browser-es5`
+2. Bundle and transpile all JavaScript files
+3. Update HTML files to reference the ES5 assets
+
+### Testing the ES5 Build
+
+To verify the ES5 build process:
+
+```bash
+node scripts/test-es5-build.js
```
bdsl, line 50:
```suggestion
scripts/test-es5-build.js
```

PR #2138, opened 2025-09-18
DON-1157 – fix leaving 'Your donation' step via enter key

NoelLH:


src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -173,13 +173,7 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
                 Retry
               </button>
             } @else {
-              <button
-                type="button"
-                class="continue"
-                mat-raised-button
-                color="primary"
-                (click)="progressToNonAmountsStep()"
```
NoelLH, line 181: I know we moved to single-purpose button click events, but I think this bug is a case in point re. keeping track of all the various ways donors can actually navigate around the form.

It's now feeling less risky to me to move back towards a global change handler which always checks the state as it runs rather than trying to remember every possible keypress and button press that could make changes and handle those individually.

bdsl: I'm not sure I quite get yet how you mean moving back towards a global change handler - maybe would be easier to understand if I can separate out the bugfix from any refactoring here. 

Are you saying you'd want to remove `continueFromPaymentStep`, `interceptSubmitAndProceedInstead`, `progressFromStepGiftAid` and `progressFromStepReceiveUpdates` and replace them all with direct calls to `next` more like we had before the change two years ago at https://github.com/thebiggive/donate-frontend/blob/adec0b2104be7a72df1e609b890987bca458a6a0/src/app/donation-start/donation-start-form/donation-start-form.component.html ?

NoelLH:
> Changes as they are LGTM, I see how its fixing the bug. Added some questions and comments about any wider plan inline.

Thanks, I'll merge so we can test and think more about those tomorrow

PR #2140, opened 2025-09-19
BG2-2892: Output alt-text on campaign additional images if present

bdsl:
(We know it won't be present currently as we don't offer a way for charities to enter it, but hoping to implement that soon)

Screenshot From 2025-09-19 11-08-33

### DIFF
--- a/src/app/campaign-details/campaign-details.component.html
+++ b/src/app/campaign-details/campaign-details.component.html
@@ -199,10 +199,24 @@
 
               @if (campaign.additionalImageUris.length > 0) {
                 <div class="c-additional">
-                  <!-- I think we didn't collect alt-text from charities for these images, so can't show it here -->
-                  <!-- eslint-disable @angular-eslint/template/alt-text -->
                   @for (image of campaign.additionalImageUris; track $index) {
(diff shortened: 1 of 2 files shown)


src/app/campaign-details/campaign-details.component.html
```diff
@@ -199,10 +199,24 @@
...
-                  <!-- eslint-disable @angular-eslint/template/alt-text -->
                   @for (image of campaign.additionalImageUris; track $index) {
-                    <img [src]="image.uri | optimisedImage: 850 | async" class="c-additional__images" />
+                    <!--- at east for now, using typeof so we treat empty alt as distinct from missing alt. Ideally
+                          empty alt will mean someone has decided the image does not want alt-text, while missing alt
+                          means they haven't made that decision, probably because the image was uploaded at a time when
+                          we did not accept alt-text. Whether we'll actually build the data structure to allow
+                          distinguishing these cases remains to be decided.
+                          -->
+                    @if (typeof image.alt_text === "string") {
```
bdsl, line 209: I wasn't quite sure if the `@if` was necessary, I thought maybe angular would be clever enough to omit the alt attribute if we pass undefined, but it isn't, if we pass undefined then it converts it to the string `undefined` and outputs that, so we do need the check here.

PR #2146, opened 2025-09-19
DON-1157 – run step exit validation actions more consistently

NoelLH:


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1204,30 +1199,6 @@ export class DonationStartFormComponent
     );
   }
 
-  captchaIdentityReturn(captchaResponse: string | null) {
```
NoelLH, line 1207: This fn just happens to be unused already as far as I can see, not directly related to the main validation change

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -191,7 +192,12 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
         </div>
       </mat-step>
       @if (campaign.currencyCode === "GBP") {
-        <mat-step formGroupName="giftAid" [stepControl]="giftAidGroup" label="Gift Aid">
+        <mat-step
+          formGroupName="giftAid"
+          [stepControl]="giftAidGroup"
+          label="Gift Aid"
+          (interacted)="progressFromStepGiftAid()"
```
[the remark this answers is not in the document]
NoelLH: I think it's the former because we are using the `@Output` – from Jira earlier:

> I did a bit more local testing, focusing on the Gift Aid step, and I don’t think there are really any critical validation problems. The fact that the steps are all bound to form groups means that Angular is enforcing all the rules that we defined at form level.
>
> If you happen to press enter with an incomplete Gift Aid step the UX is certainly imperfect though – it seems like stepping forward is blocked but nothing tells you what the error is, or that there is one.
>
> One promising options seems to be to use the `interacted` `@Output` on each [MatStep](https://material.angular.dev/components/stepper/api#MatStep). We already have methods on the buttons per-step so maybe we can directly swap those out and use this instead.
>
> The event
>
>>   Emits when the user has attempted to move away from the step.
>
> So tweaking my experiment above slightly we could make it show the toast and/or update the page to show both Angular & custom errors in-page.

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -302,19 +308,16 @@ <h3 class="b-rt-0 b-m-0 b-bold"><span class="span-hr"></span>How does Big Give u
...
+            <button type="button" class="continue" mat-raised-button color="primary" (click)="next()">Continue</button>
           </div>
         </mat-step>
       }
-      <mat-step formGroupName="payment" [stepControl]="paymentGroup" label="Payment details">
+      <mat-step
+        formGroupName="payment"
+        [stepControl]="paymentGroup"
+        label="Payment details"
+        (interacted)="continueFromPaymentStep()"
```
bdsl, line 319: won't this stop people going backwards from the payment step if they haven't filled in payment details? Presumably we should still allow backwards movement in that case just not forwards.

bdsl: hmm so tested, on this branch and it does allow me progress backwards, without filling in the payment step, but it also shows this error toast - we shouldn't be showing any error message if the user just wants to progress backwards.

image

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1331,7 +1293,7 @@ export class DonationStartFormComponent
    * Callers must refuse to leave step 1 until this is passing because Stripe will be unusable if we haven't
    * done this initial work.
```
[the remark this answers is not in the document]
NoelLH: This might point to another logic tweak that's needed with the new approach. This doesn't have any non-stepper-Output callers any more so its return value being/seeming important is probably a mistake.

PR #2152, opened 2025-09-30
BG2-2972 slider refinement

NoelLH:
* [BG2-2972 – replace tip slider's containing box](https://github.com/thebiggive/donate-frontend/commit/dfeb273c0c08688a1af7be3d14b508bcac664564) – Make input wider, replace the box with an active thumb focus colour, and add an ARIA label directly on the input
* [BG2-2972 – update tip % (and therefore preview in £) as donors drag](https://github.com/thebiggive/donate-frontend/commit/79f8a1870cc43ab8d4f40724bbd332ed7ff13bf6)

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -281,7 +281,7 @@ export class DonationStartFormComponent implements AfterContentInit, OnDestroy, 
    * should always come after the last slider label update including when e.g. using keyboard escape, so this
    * should be safe.
    */
-  formatSliderLabel = (value: number): string => {
+  updateTipPercentAndFormatSliderLabel = (value: number): string => {
     this.setTipPercentage(value);
 
     return `${value}%`;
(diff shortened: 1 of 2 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -274,9 +274,18 @@ export class DonationStartFormComponent implements AfterContentInit, OnDestroy,
...
 
-  formatSliderLabel(value: number): string {
+  /**
+   * Because MatSliderThumb doesn't have any events to react to changes until the end of a drag, we've given
+   * this the side effect of updating the form's tip amount value in absolute currency, because that was the
+   * simplest way to give a live preview of the tip amount while dragging. The final thumb `valueChange()` event
+   * should always come after the last slider label update including when e.g. using keyboard escape, so this
+   * should be safe.
+   */
+  formatSliderLabel = (value: number): string => {
```
bdsl, line 284: maybe rename to something like updateTipPercentAndFormatAsPercent so it will also be clear when reading the html what's going on?

bdsl (dismissed):
Added a suggestion inline, not a blocker.

PR #2153, opened 2025-09-30
BG2-2962 – add a Matomo log to help track payment method outcomes

NoelLH:
And evaluate Pay By Bank impact overall and completion vs. initial interest

A sample real-time log entry from my local test:
Screenshot 2025-09-30 at 13 42 00

I'm not 100% sure but I think using Event Name like this is probably just as flexible as if each method had its own Event Action. It looks like Goal configs can use either so I am assuming other ways of splitting up traffic can too.


bdsl (approved):
LGTM

PR #2156, opened 2025-10-01
BG2-2972 – give tip slider a 3rd state & colour for when it's mid-drag

NoelLH:


bdsl (approved):
Code looks fine. The three colours do feel like a bit too maximalist in design now for me, especially since they seem to be three unrelated colours rather than any of them being variants of the same thing, but not a blocker.

This his how it appears on my machine when using a combination of keyboard and mouse input:

https://github.com/user-attachments/assets/9aab43fb-8f63-4d57-86dd-32a9a0d9c74a

NoelLH:
> Code looks fine. The three colours do feel like a bit too maximalist in design now for me, especially since they seem to be three unrelated colours rather than any of them being variants of the same thing, but not a blocker.
> 
> This his how it appears on my machine when using a combination of keyboard and mouse input:
> Screencast.From.2025-10-01.16-11-57.mp4

Thanks, will merge for now as we've not had any other input on the choice yet and in practice most donors will only see 2 of the 3 colours.

PR #2159, opened 2025-10-06
BG2-2971: Replace all links to mailjet signup form with links to new …

bdsl:
…form that creates SF comms preference

(and so will indirectly add records to mailjet)

Enabling this new form and linking to it instead of the mailchimp form:

image


DomTBG (approved):
LGTM

PR #2317, opened 2026-03-30
Bg2 3106: WIP - take payments using Ryft

bdsl:


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1017,6 +1021,37 @@ export class DonationStartFormComponent implements OnDestroy, OnInit, AfterViewI
     await this.payWithStripe();
   }
 
+  payWithRyft: () => Promise<void> = async () => {
```
bdsl: I think this was taken directly from the ryft docs - untested so far, and also I think we'll probably want to change this to something where we store the ryft session ID in matchbot and initiate the payment from there at this point, more like we do with Stripe.

NoelLH (approved):

PR #2323, opened 2026-04-07
DON-1194: Show total to pay at start of donation form

bdsl:
image

image

Not sure if there's any reason we'd really want the dialog text to be grey rather than black as it is here - because of defaults we're inheriting from the material design component - but that's the same as in our existing continue donation modal dialogs.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.scss
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.scss
@@ -62,7 +62,6 @@ div.mat-step-label.mat-step-label-selected {
   font-size: 17px !important;
   line-height: 20px;
   @media #{abstract.$breakpoint-lg} {
-    font-size: 20px !important;
     line-height: 27px;
   }
 }
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.scss
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.scss
@@ -31,7 +31,6 @@
   margin-bottom: -1em;
   text-align: center;
   .total {
-    font-size: 20px;
     font-weight: bold;
     color: abstract.$colour-primary;
   }
@@ -62,6 +61,7 @@ div.mat-step-label.mat-step-label-selected {
   font-size: 17px !important;
   line-height: 20px;
   @media #{abstract.$breakpoint-lg} {
+    font-size: 20px !important;
     line-height: 27px;
   }
 }
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.scss
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.scss
@@ -22,6 +22,21 @@
   width: 100%;
 }
 
+.donation-input.donation-input-main.tipSliderWrapper {
+  margin-bottom: 0;
+}
+
+.stepOneAmountSummary {
+  margin-top: 1em;
+  margin-bottom: -1em;
+  text-align: center;
+  .total {
+    font-size: 20px;
+    font-weight: bold;
+    color: abstract.$colour-primary;
+  }
+}
+
 .mat-stepper-horizontal,
 .mat-stepper-vertical {
   background: none;
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -155,12 +155,3 @@ div.mat-expansion-panel-body {
   font-size: 18px;
   font-weight: normal;
 }
-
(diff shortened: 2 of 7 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.scss
```diff
@@ -22,6 +22,21 @@
...
+.donation-input.donation-input-main.tipSliderWrapper {
+  margin-bottom: 0;
+}
+
+.stepOneAmountSummary {
+  margin-top: 1em;
+  margin-bottom: -1em;
+  text-align: center;
+  .total {
+    font-size: 20px;
```
NoelLH: Because this is fixed while other text size scales by breakpoint, it is very large relative to the other text on mobile; probably worth reducing it a bit.

At desktop, the whole new block is also as big as the tip messaging above whereas in the mockup it's a little smaller. That might be worth adjusting too, although at mobile I'm not sure if we would want to go below 14px.

That does bring up one final minor issue in this area – on mobile the line with 'Why tip?' doesn't fit without wrapping. Wondering if it could be worth the complexity of wrapping before the brackets on the smallest screens, but not a blocker.

src/app/donation-start/donation-start-form/donation-start-form.component.scss
```diff
@@ -22,6 +22,21 @@
...
+    font-weight: bold;
+    color: abstract.$colour-primary;
```
NoelLH, line 35:
```suggestion
    color: abstract.$colour-primary;
    margin: 0.8em 0;
    display: block;
```
IMO this looks too bunched up with no vertical space around it.

NoelLH (commented):
I can't think of a good reason for the modal text to be the lower contrast Material default grey. Happy for us to change that across the board if it's straightforward.

I've made some small visual suggestions inline. Otherwise all LGTM.

src/app/donation-start/donation-start-form/donation-start-form.component.scss
bdsl: Thanks, pushed b88810e3 about this. I've removed the font-size entirely from the total line - I don't think it's needed to emphasize it, given that it's bold and a different colour and now also has a paragraph break.

I've put a non-breaking space in "Why tip" so it will wrap before the brackets when needed.

I'm not sure it's worth adjusting to make the text slightly smaller - maybe if we had either utility classes or a relative sizing system so that we didn't have explicitly set multiple sizes for each element for different screen sizes.

src/app/donation-start/donation-start-form/donation-start-form.component.scss
bdsl: Right, I think this is the same issue that Leondré raised on slack, addressed by https://github.com/thebiggive/donate-frontend/pull/2323/changes/0f1c0fe66898e564b769dc4106a8a51086a988ea

src/app/donation-start/donation-start-form/donation-start-form.component.scss
```diff
@@ -47,7 +62,6 @@ div.mat-step-label.mat-step-label-selected {
   font-size: 17px !important;
   line-height: 20px;
   @media #{abstract.$breakpoint-lg} {
-    font-size: 20px !important;
```
NoelLH, line 50: I think this was removed in error instead of the new 20px above

NoelLH (commented):
I think the wrong 20px line was deleted.

Apart from that and prettier checks, LGTM

src/app/donation-start/donation-start-form/donation-start-form.component.scss
bdsl: Thanks, fixed in 9a3964d5

bdsl:
image

image

src/styles.scss
```diff
@@ -155,3 +155,12 @@ div.mat-expansion-panel-body {
...
 }
+
+:root {
+  @include mat.dialog-overrides(
+    (
+      supporting-text-color: abstract.$colour-black,
+      // overrides default grey
+    )
+  );
+}
```
NoelLH: Perhaps fits slightly better / easier to find in `_overwrite-mat.scss` where we do similar for the slider. Not a blocker.

NoelLH (dismissed):

leondre-bg:
LGTM!

PR #2329, opened 2026-04-14
Use much smaller metacampaign banners again

NoelLH:
via `optimisedImage` pipe.

e.g. webp in all modern browsers which reduces ER26 banner size by more than 10x.

Also retire the old hero component that no(?) metacampaigns use any more.


bdsl (approved):
LGTM

PR #2334, opened 2026-04-16
DON-1198: Refactor, extract function DonationStartFormComponent.donor…

bdsl:
…HasDonationFunds

This is used several times and about to be used once more, so make sense to extract to a function.

image

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -44,128 +44,122 @@
             {{ displayableAmountsStepErrors() }}
           </div>
         }
-        @if (creditPenceToUse === 0) {
+        @if (!donorHasDonationFunds()) {
           <div>
             <hr class="u-margin-auto-y-20" />
             <h2 class="b-rt-0 b-m-0 b-bold">Support Big Give</h2>
-            @if (creditPenceToUse === 0) {
-              <p>
-                Tips to Big Give (registered charity number 1136547) enable us to deliver our charitable mission and
-                help us provide match funds to our charity partners. We truly appreciate any tip you can give - no
-                matter the size.
-              </p>
-            }
-            @if (creditPenceToUse === 0) {
-              <div class="u-text-center">
-                @if (!showCustomTipInput) {
-                  <div class="donation-input donation-input-main tipSliderWrapper">
-                    <mat-slider
-                      id="tipSlider"
-                      [min]="minimumTipPercentage"
-                      [max]="maximumTipPercentage"
-                      step="1"
-                      discrete
-                      [displayWith]="formatSliderLabel"
-                      class="tip-slider"
-                      [class.slider-being-dragged]="isTipSliderBeingDragged"
-                      aria-label="Tip percentage"
-                    >
-                      <input
-                        matSliderThumb
-                        [value]="tipPercentage"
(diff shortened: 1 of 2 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -44,18 +44,15 @@ <h1 class="donationInfoWrapper">
             {{ displayableAmountsStepErrors() }}
           </div>
         }
-        @if (creditPenceToUse === 0) {
+        @if (! donorHasDonationFunds()) {
           <div>
             <hr class="u-margin-auto-y-20" />
             <h2 class="b-rt-0 b-m-0 b-bold">Support Big Give</h2>
-            @if (creditPenceToUse === 0) {
```
bdsl, line 51: not sure why this was here, think it must have been left over from old code, maybe it made sense before some things were moved around, but clearly redundant since it's the same condition as the surrounding `if`. A better static analyser would say it's redundant. Same applies to next one.

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -44,138 +44,134 @@ <h1 class="donationInfoWrapper">
...
           <div class="stepOneAmountSummary">
             <p>
               Donation to {{ campaign.charity.name }}: {{ donationAmount | exactCurrency: campaign.currencyCode }}
-              <br />
-              Tip Donation to support Big Give: {{ tipAmount() | exactCurrency: campaign.currencyCode }}
-              <a role="button" tabindex="0" (click)="showTipExplanation()">(Why&nbsp;tip?)</a>
+              @if (!donorHasDonationFunds) {
+                <br />
+                Tip Donation to support Big Give: {{ tipAmount() | exactCurrency: campaign.currencyCode }}
+                <a role="button" tabindex="0" (click)="showTipExplanation()">(Why&nbsp;tip?)</a>
```
shaundbs, line 173: Should these be buttons rather than links?

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -63,8 +63,8 @@ import { LiveAnnouncer } from '@angular/cdk/a11y';
 import { BiggiveFormFieldSelect, BiggiveTextInput } from '@biggive/components-angular';
 import { MatInput } from '@angular/material/input';
 import { MatHint } from '@angular/material/form-field';
-import { MatExpansionPanel, MatExpansionPanelHeader } from '@angular/material/expansion';
 import { MatIcon } from '@angular/material/icon';
+import { MatExpansionPanel, MatExpansionPanelHeader } from '@angular/material/expansion';
```
bdsl, line 67: moved this by mistake but the order here doesn't matter

src/app/donation-start/donation-start-form/donation-start-form.component.html
shaundbs: Good spot, thanks for making this change @bdsl

src/app/donation-start/donation-start-form/donation-start-form.component.html
bdsl: Maybe, but that's not being changed in this PR. Do you think they should be? Probably better to change separately if so.

shaundbs (approved):

PR #2339, opened 2026-04-21
BG2-3121: Style hide/show password buttons like links for brand consi…

bdsl:
…stency

image

image

### DIFF
--- a/src/app/account/delete-account/delete-account.component.html
+++ b/src/app/account/delete-account/delete-account.component.html
@@ -39,6 +39,7 @@
                 [formControl]="form.controls.password"
               />
               <biggive-button
+                aria-controls="password"
                 colourScheme="clear-primary"
                 (doButtonClick)="toggleShowPassword()"
                 [label]="showPassword() ? 'Hide' : 'Show'"
--- a/src/app/account/delete-account/delete-account.component.html
+++ b/src/app/account/delete-account/delete-account.component.html
@@ -38,9 +38,12 @@
                 autocomplete="password"
                 [formControl]="form.controls.password"
               />
-              <button type="button" (click)="toggleShowPassword()">
-                {{ showPassword() ? "Hide" : "Show" }}
-              </button>
+              <biggive-button
+                colourScheme="clear-primary"
+                (doButtonClick)="toggleShowPassword()"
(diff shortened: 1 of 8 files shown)


NoelLH (commented):
I think the basic component use looks good but we should consider accessibility more for this feature generally before it goes out.

I think both ideas in the "Making sure it’s accessible" section of https://technology.blog.gov.uk/2021/04/19/simple-things-are-complicated-making-a-show-password-option/ apply to us too. I'm pretty sure we can do both things – associating the buttons properly with what they affect and live state announcement – without altering our custom components' internals.

(I briefly looked at `aria-pressed` but because we are varying the copy I don't think it's appropriate to treat these like toggle buttons, so it doesn't fit.)

bdsl:
> I think the basic component use looks good but we should consider accessibility more for this feature generally before it goes out.
> 
> I think both ideas in the "Making sure it’s accessible" section of https://technology.blog.gov.uk/2021/04/19/simple-things-are-complicated-making-a-show-password-option/ apply to us too. I'm pretty sure we can do both things – associating the buttons properly with what they affect and live state announcement – without altering our custom components' internals.
> 
> (I briefly looked at `aria-pressed` but because we are varying the copy I don't think it's appropriate to treat these like toggle buttons, so it doesn't fit.)

Thanks @NoelLH - I have implemented both of those things.

NoelLH (approved):
Updated version LGTM.

Findings from accessibility testing on local:
* Firefox and Chrome with Apple VoiceOver on both have good keyboard usability and now announce the updated state as I would expect. They correctly report the status of the password field when it is secure and not read; and read its contents only upon focus when it is `text`. Typically that would only happen after the state change announcement, giving donors extra warning before getting to a point where their password may be read out loud.
* `aria-control` does not seem to set an association in Firefox's tree at all when shadow DOM is involved in this way, even though light DOM equivalents do it as I'd expect; but a connection does seem to show up in Chrome. And without another tool involved, it doesn't seem to change the behaviour in practice in either browser, so doesn't seem very high value to spend more time on unless we learn more about e.g. a concrete benefit with a screen reader that we are missing because of this. If we were to later want to change all password fields that have the new feature again for another reason anyway, this might be an extra point in favour of abstracting the whole pattern into a new composed 'Big Give component'.

PR #2344, opened 2026-04-28
DON-1211 Guarantee that trying to log a mysterious `error` can't let resolvers take out the rest of the page

NoelLH:


src/app/campaign-stats-resolver.ts
```diff
@@ -21,7 +21,7 @@ export class CampaignStatsResolver implements Resolve<FormattedCampaignStats> {
       catchError((error) => {
         logCampaignCalloutError(
           isPlatformBrowser(this.platformId),
-          `CampaignStatsResolver: ${error.message}`,
+          `CampaignStatsResolver: ${error.message ?? error ?? 'Unknown error'}`,
```
bdsl: should we use null-safe navigation, `error?.message` instead of `error.message` to be sure that can't cause a crash? I don't think `error` could ever a non-object but seems worth it if we want to make this as robust as possible. If we type error as `unknown` then I think it would force null-safe navigation.

bdsl (commented):
Added a comment inline that applies to all changed lines, otherwise LGTM.

src/app/campaign-stats-resolver.ts
NoelLH: Thanks, yep added the `?` in 264490772e0307bff897d1f80acf0a3defed3337.

I also missed a resolver before, fixed that in same commit.

If I try to say it's `unknown` then it looks like `error?.message` is also a type error, so leaving it implicit `any` for now.

bdsl: Yep, `error?.message` is still a type error but I don't think it can ever throw an error at runtime now.

bdsl (approved):
LGTM

PR #2351, opened 2026-05-07
BG2-3143: Fix null name (standard for orgs) being shown as 'null' ins…

bdsl:
…tead of ''

### DIFF
--- a/src/app/account/delete-account/delete-account.ts
+++ b/src/app/account/delete-account/delete-account.ts
@@ -50,7 +50,7 @@ export class DeleteAccount implements OnInit {
       }
 
       this.person = person;
-      this.fullName = this.normaliseName(person.first_name + ' ' + person.last_name);
+      this.fullName = this.normaliseName([person.first_name, person.last_name].filter(Boolean).join(' '));
     });
   }
 
--- a/src/app/donation.service.ts
+++ b/src/app/donation.service.ts
@@ -504,6 +504,7 @@ export class DonationService {
         return false;
 
       default:
+        // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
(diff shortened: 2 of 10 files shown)


src/app/account/delete-account/delete-account.ts
```diff
@@ -50,7 +50,7 @@ export class DeleteAccount implements OnInit {
       }
 
       this.person = person;
-      this.fullName = this.normaliseName(person.first_name + ' ' + person.last_name);
+      this.fullName = this.normaliseName([person.first_name, person.last_name].filter(Boolean).join(' '));
```
bdsl, line 53: the old version can produce results like `"null Our Giving Club"`. Eslint changes disallow concatenating null or undefined and other changes follow from that.

src/app/donation.service.ts
```diff
@@ -504,6 +504,7 @@ export class DonationService {
         return false;
 
       default:
+        // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
```
bdsl: eslint doesn't allow concatenating with `never`. IMHO since never is the bottom type and therefore a subtype of `string` I should be allowed to do anything I can do with a string with a never, but eslint disagrees so disabling the rule here.

shaundbs:
Nice cleanup and good use of stricter ESLint rules!

Only thing to add is that might there’s some duplicated formatting logic that could be extracted (non-blocking)

shaundbs (approved):

bdsl:
Thanks, yep the logic in explore component is a bit duplicated but I'm not seeing an easy way to improve it - if I extract a function it would have so many params that I think the call would be almost as long as the implementation. Will merge-as is, feel free to make an improvement after if you can see one.

PR #2361, opened 2026-05-28
DON-1212 exit step tuning

NoelLH:
* [Standardise form submit interceptions](https://github.com/thebiggive/donate-frontend/pull/2361/changes/8a4308633fc744c5ea788bdf33d5ae47c8faefcf)
Primarily, in the two important donate forms, ensure the interception logic is independent of the input method.
* [Limit when pressing enter advances the donor from step 1, amount & tip](https://github.com/thebiggive/donate-frontend/pull/2361/changes/1b9f3a60a7fe7a62e0d83bd77e697d8724501a1c)
And when it's not suitable to advance, scroll them to the tip information. Never advances from this step on touch devices, as explained in code comments.

(Also make various less-critical forms where we just want simple submission not be needlessly coupled to input method.)

Input interception change tested in desktop Firefox and simulated iOS 26.2 mobile Safari with software keyboard enabled and pressing its enter key – which previously caused iOS to try to submit the whole form and show validation errors. Both LGTM so far with the change.

The hidden submit button is an AI suggestion, not 100% certain it's necessary but if it doesn't cause screen reader regressions we can identify then it sort of makes sense that it might let us be more sure of the same behaviour across platforms.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -1,7 +1,7 @@
 @if (campaignOpenOnLoad) {
   <form class="c-donate-form" (ngSubmit)="interceptSubmitAndProceedInstead($event)" [formGroup]="donationForm">
     <!-- Hidden submit button to ensure mobile / soft keyboards consistently trigger form submission -->
-    <button type="submit" class="b-hidden" style="display: none" aria-hidden="true"></button>
+    <button type="submit" class="b-hidden" aria-hidden="true"></button>
 
     <mat-vertical-stepper id="stepper" linear #stepper (selectionChange)="stepChanged($event)" animationDuration="0">
       <mat-step
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -1,7 +1,7 @@
 @if (campaignOpenOnLoad) {
   <form class="c-donate-form" (ngSubmit)="interceptSubmitAndProceedInstead($event)" [formGroup]="donationForm">
     <!-- Hidden submit button to ensure mobile / soft keyboards consistently trigger form submission -->
-    <button type="submit" class="b-hidden" style="display: none;" aria-hidden="true"></button>
+    <button type="submit" class="b-hidden" style="display: none" aria-hidden="true"></button>
 
     <mat-vertical-stepper id="stepper" linear #stepper (selectionChange)="stepChanged($event)" animationDuration="0">
       <mat-step
@@ -50,7 +50,7 @@
         @if (!donorHasDonationFunds()) {
           <div>
             <hr class="u-margin-auto-y-20" />
-            <h2 class="b-rt-0 b-m-0 b-bold">Support Big Give</h2>
+            <h2 id="supportBigGiveHeading" tabindex="-1" class="b-rt-0 b-m-0 b-bold">Support Big Give</h2>
             <p>
               Tips to Big Give (registered charity number 1136547) enable us to deliver our charitable mission and help
               us provide match funds to our charity partners. We truly appreciate any tip you can give - no matter the
@@ -179,7 +179,7 @@
         }
 
         <div aria-live="polite">
-          <div class="u-text-center">
+          <div id="step1Continue" class="u-text-center">
             @if (donationCreateError) {
               <p class="error">
                 Sorry, we can't register your donation right now. Please try again in a moment or
@@ -685,13 +685,7 @@
             </table>
           }
           @if (!submitting && !runningFinalPreSubmitUpdate && !finalPreSubmitUpdateFailed) {
-            <button
-              type="button"
-              (click)="submit()"
-              class="continue c-donate-button"
-              mat-raised-button
-              color="primary"
-            >
+            <button type="button" (click)="submit()" class="continue c-donate-button" mat-raised-button color="primary">
               Donate now
             </button>
           }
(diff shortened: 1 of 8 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -1,5 +1,8 @@
 @if (campaignOpenOnLoad) {
-  <form class="c-donate-form" (keyup.enter)="interceptSubmitAndProceedInstead($event)" [formGroup]="donationForm">
+  <form class="c-donate-form" (ngSubmit)="interceptSubmitAndProceedInstead($event)" [formGroup]="donationForm">
+    <!-- Hidden submit button to ensure mobile / soft keyboards consistently trigger form submission -->
+    <button type="submit" class="b-hidden" style="display: none;" aria-hidden="true"></button>
```
bdsl: Not sure if it does any harm but seems redundant to have both `style="display: none;"` and `aria-hidden="true"`. MDN says "_aria-hidden="true" should not be added when:... The element or the element's ancestor is hidden with [display: none](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/display)_"

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -683,8 +686,8 @@ <h3>A small tip makes a big difference. You will help us:</h3>
           }
           @if (!submitting && !runningFinalPreSubmitUpdate && !finalPreSubmitUpdateFailed) {
             <button
+              type="button"
               (click)="submit()"
-              (keyup.enter)="submit()"
```
bdsl, line 687: @NoelLH I found your comment on github from when you added this line, you [said](https://github.com/thebiggive/donate-frontend/pull/1419):

> In macOS Firefox, without the key binding I was able to select the button like others and saw it in a visually focused state, but had to press space to activate it. I think supporting enter too is possibly helpful for occasional keyboard focus users.

I think now removing enter support may actually be worthwhile to make things more consistent.

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -278,7 +278,7 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy
 
   async interceptSubmitAndProceedInstead(event: Event) {
     event.preventDefault();
-    await this.next();
+    this.continue();
```
bdsl, line 281: shall we delete the now unused `RegularGivingComponent.next` (which is now unused and is just a one liner `this.stepper.next();`. I'm not sure why we had both `continue` and `next`.

src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1338,8 +1338,47 @@ export class DonationStartFormComponent implements OnDestroy, OnInit, AfterViewI
...
+   * When it's not appropriate to auto exit step 1, scroll down if applicable to move the tip information into view.
+   *
+   * Mobile is handled differently because of the high probability that a software keyboard is obscuring tip information
+   * in a way we can't detect. Directly pressing Continue or navigating by pressing step headers still work without
+   * this interception.
+   */
   async interceptSubmitAndProceedInstead(event: Event) {
     event.preventDefault();
+
+    const isTouchDevice = window.matchMedia('(pointer: coarse)').matches;
```
bdsl, line 1355: TIL of pointer: coarse 👍

src/app/donation-start/donation-start-form/donation-start-form.component.html
NoelLH: Yeah, I think given tools don't always perfectly implement the specs it's OK and probably safer to keep the redundant ARIA part here as it's pretty critical.

I now see that the class and inline styling were the same thing though, deleted the latter in deeca74ea8d70f478ea919f24af2d2211abaec56

src/app/regular-giving/regular-giving.component.ts
NoelLH: Yep, done in dc2ca3773e217a79c75a806d639b5754a51f8baf

src/app/donation-start/donation-start-form/donation-start-form.component.html
```diff
@@ -179,7 +179,7 @@ <h3>A small tip makes a big difference. You will help us:</h3>
         }
 
         <div aria-live="polite">
-          <div class="u-text-center">
+          <div id="step1Continue" class="u-text-center">
```
bdsl, line 182: not sure if it'd be worth setting the IDs from class properties so the compiler will check the reference to it matches something, and we can click through in the IDE to see usages.

e.g. `[id]="step1ContinueID"`

NoelLH: I think for me it's not quite worth the cognitive overhead when reading while there's only 1 use case.

bdsl:
Re "@NoelLH
[DON-1212 Delete extraneous inline style](https://github.com/thebiggive/donate-frontend/pull/2361/commits/deeca74ea8d70f478ea919f24af2d2211abaec56)" 

Isn't that going to make it show up visually while staying hidden from screen readers? I thought we wanted to hide it for everyone which `display: none;` alone should do.

bdsl:
@NoelLH I've reviewed, added comments inline, all else LGTM.

NoelLH:
> Re "@NoelLH [DON-1212 Delete extraneous inline style](https://github.com/thebiggive/donate-frontend/pull/2361/commits/deeca74ea8d70f478ea919f24af2d2211abaec56)"
> 
> Isn't that going to make it show up visually while staying hidden from screen readers? I thought we wanted to hide it for everyone which `display: none;` alone should do.

The class has `display: none`.

bdsl (approved):
LGTM!

PR #2362, opened 2026-05-29
DON-1204 Add Leaflet maps w/ initial UK outline + project highlight support

NoelLH:
Initial map loaded with leaflet v2 alpha, now with ONS Geography overlays for showing project impact at all supported region sizes.

Screenshot 2026-06-05 at 18 52 26

Code's now in `campaign-info` so it is technically reused by the mobile & desktop versions of this one page. We'll probably want to abstract further into a more flexible `MapComponent` when we built a search with map, but I think we could merge this meanwhile.

With screen readers this should now read a list of impact areas like so (visuals are from temporary WIP but the grey box shows what VoiceOver read out to me. In Safari this happened automatically after going through menu items, when the page loaded):

Screenshot 2026-06-05 at 17 52 39

(Some aspects of latest commits had some Copilot + Gemini input but not a majority of most so haven't kept track in detail.)

---

(previous early demo):

Screenshot 2026-06-02 at 20 13 51


NoelLH:
Blank map from first commit WIP (superseded):

Screenshot 2026-05-29 at 12 57 42

src/app/campaign.model.ts
```diff
@@ -51,10 +51,11 @@ export type Campaign = {
     twitter?: string;
     website: string;
   };
-  countries: string[];
+  countries: string[]; // TODO remove
   impactReporting: string | null;
   impactSummary: string | null;
   isMatched: boolean;
+  locations?: Array<{ countryName: string; regionCode: string }>; // TODO make always defined once other apps ready
```
bdsl:
```suggestion
  locations?: Array<{ countryName: string|null; regionCode: string|null }>; // TODO make always defined once other apps ready
```

But also consider replacing with a union so we assume each location is exactly one of a countryName or a regionCode

src/app/campaign-details/campaign-details.component.html
```diff
@@ -46,7 +48,7 @@
             [imageAlt]="campaign.banner.alt_text"
             [slug]="campaign.isMatched ? 'Match Funded' : null"
             [charityName]="campaign.charity.name"
-            [charityLocation]="campaign.countries.length > 1 ? 'Multiple locations' : campaign.countries[0] || null"
+            [charityLocation]="null"
```
NoelLH, line 49: Takes out the country name or 'Multiple locations' just below the charity name.

Next step will be to make it optional in components, then delete.

src/app/campaign.model.ts
NoelLH: Made a union in 823764b95ec7fc9948973de3782de0d02827b205

src/app/campaign-info/campaign-info.component.html
```diff
@@ -57,5 +57,27 @@ <h5>Beneficiaries</h5>
...
+      <h5>Impact areas</h5>
+      <hr />
+      <ul>
+        @for (country of countries; track country) {
+          <li>
+            {{ country }}
+          </li>
+        }
+      </ul>
+      @if (hasRegions) {
```
bdsl, line 71: I know we expect this to be false for everything in the short term, but I think it might be worth introducing an explicit feature flag as well as belt-and-braces to make sure the map doesn't get shown before we're ready to show it.

bdsl (approved):
Added one new comment inline, otherwise all LGTM.

src/app/campaign-info/campaign-info.component.html
NoelLH: As discussed, think this particular feature's safe enough to not flag.

PR #2363, opened 2026-06-02
BG2-3310: Use a three column grid for four homepage highlight cards.

bdsl:
Use up to three columns for any other number of cards.

And to allow some testing this will detect the query string `?repeatLastCard=true` and add an extra copy of the last card to the page.

image


NoelLH (approved):
LGTM.

I had a bit more of a look at the general `biggive-grid` SCSS and production's current behaviour and actually only a very narrow range of sizes look slightly suboptimal. I think nothing is terrible and the most important breakpoints look good, so no real need to spend time tuning that.

PR #2367, opened 2026-06-03
DON-1192: Clarify that we don't need to explicitly handle 3DS for Ryft

bdsl:
I tested using the `4938730000000001` card and although it didn't actually work in my test on staging or my first test on local it did work on a third test on local with no substantive changes, so it looks like we don't need to explicitly handle 3DS.

Possibly the Ryft 3DS authentication infra and/or my current home connection is a bit unreliable affecting those tests.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -1036,12 +1036,6 @@ export class DonationStartFormComponent implements OnDestroy, OnInit, AfterViewI
       throw new Error('donation  is not initialized');
     }
 
-    // @todo BG2-3106 - call matchbot to check it's OK to take payment for this donation
-    // (e.g. matching hasn't expired) and await the decision before the next line, as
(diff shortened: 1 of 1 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -1036,12 +1036,6 @@ export class DonationStartFormComponent implements OnDestroy, OnInit, AfterViewI
       throw new Error('donation  is not initialized');
     }
 
-    // @todo BG2-3106 - call matchbot to check it's OK to take payment for this donation
-    // (e.g. matching hasn't expired) and await the decision before the next line, as
-    // it isn't possible to initiate the payment from server side.
-
-    // Also make sure the amount on the payment session is updated based on the selected tip.
-
```
bdsl, line 1044: as noted below the capture is handled in matchbot so we don't need to ask matchbot's permission before we attempt payment.

NoelLH (approved):

PR #2370, opened 2026-06-09
DON-1215 Make smallest base font size 16px

NoelLH:
Sort of an experiment for now but so far it seems to work.


src/app/campaign-details/campaign-details.component.html
```diff
@@ -181,7 +181,7 @@
       <div class="c-tabs">
         <mat-tab-group class="c-tab-group" animationDuration="0ms" [color]="'primary'">
           @if (campaign.problem?.trim() || campaign.solution?.trim()) {
-            <mat-tab label="Campaign overview">
+            <mat-tab label="Overview">
```
NoelLH, line 184: Long labels here are in danger of obscuring the existence of their neighbours on mobile. I also just don't think they need to be so verbose.

bdsl (approved):
LGTM

PR #2381, opened 2026-06-17
DON-1192: Allow donors who have saved funds to donate via ryft (not u…

bdsl:
…sing those funds)

Funds are stored at Stripe so only usuable for charities on Stripe, but previously we assumed any stored funds could be used on any campaign as long as it had the right currency, which led to breakage when someone with funds would try to donate to a campaign on Ryft.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.ts
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.ts
@@ -170,7 +170,8 @@ export class DonationStartFormComponent implements OnDestroy, OnInit, AfterViewI
 
   friendlyCaptchaSiteKey = environment.friendlyCaptchaSiteKey;
 
-  creditPenceToUse = 0; // Set non-zero if logged in and Customer has a credit balance to spend. Caps donation amount too in that case.
+  donorsCreditPence = 0; // Set non-zero if logged in and Customer has a credit balance in the right currency.
+  creditPenceToUse = 0; // Set non-zero if donorsCreditPence is non-zero and the campaign is one that can accept donor credit (i.e. on stripe)
   currencySymbol!: string;
 
   donationForm!: FormGroup;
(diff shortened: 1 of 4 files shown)


src/app/donation-start/donation-start-form/donation-start-form.component.ts
```diff
@@ -170,7 +170,8 @@ export class DonationStartFormComponent implements OnDestroy, OnInit, AfterViewI
 
   friendlyCaptchaSiteKey = environment.friendlyCaptchaSiteKey;
 
-  creditPenceToUse = 0; // Set non-zero if logged in and Customer has a credit balance to spend. Caps donation amount too in that case.
+  donorsCreditPence = 0; // Set non-zero if logged in and Customer has a credit balance in the right currency.
+  creditPenceToUse = 0; // Set non-zero if donorsCreditPence is non-zero and the campaign is one that can accept donor credit (i.e. on stripe)
```
NoelLH, line 174: Maybe still worth mentioning near the property's first appearance that it can limit the donation amount?

src/app/donation-start/donation-start-login/donation-start-login.component.html
```diff
@@ -11,6 +11,15 @@
           {{ creditPenceToUse / 100 | exactCurrency: campaign.currencyCode }}
         </p>
       }
+      @if (donorsCreditPence > 0 && creditPenceToUse === 0 && this.campaign.charity.psp !== "stripe") {
```
NoelLH, line 14: Makes sense on its own terms. May overstate the case and appear confusing until we use `noPsps` more helpfully (alerting donors to a Stripe config problem sooner on the donate page), as discussed yesterday.

NoelLH (approved):
One comment suggestion, no blockers. LGTM

src/app/donation-start/donation-start-login/donation-start-login.component.html
bdsl: I'm not sure how you mean overstating the case? This isn't checking for noPsps being set on the donation apge, it's just checking for the charity not having Stripe as their PSP. Isn't that different to the case we were looking at yesterday?

NoelLH: @bdsl We know from the code we checked yesterday that `psp` is null when the Stripe account ID is missing, but that doesn't actually always mean it's a non-Stripe charity.

bdsl: Right. Small PR to fix: https://github.com/thebiggive/donate-frontend/pull/2383/changes

PR #2382, opened 2026-06-18
DON-1192 Fix Ryft CSP errors seen in Staging

NoelLH:
Also don't mark non-Stripe donations 'powered by Stripe'

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -129,6 +129,10 @@ import { SADMDADomainVerificationFile } from './stripe-apple-developer-merchanti
       ],
       'default-src': [`'none'`],
       'font-src': [`'self'`, 'data:'],
+      'form-action': [
+        `'self'`,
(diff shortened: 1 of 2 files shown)


src/server.ts
```diff
@@ -129,6 +129,10 @@ import { SADMDADomainVerificationFile } from './stripe-apple-developer-merchanti
       ],
       'default-src': [`'none'`],
       'font-src': [`'self'`, 'data:'],
+      'form-action': [
```
shaundbs, line 132: Have we tested a full Ryft payment flow in browser to confirm the CSP additions are sufficient? Ryft may also need other allowances depending on what the SDK loads?

NoelLH: I think this makes sense to try with Staging next, per Jira & Slack discussions.

PR #2383, opened 2026-06-18
DON-1192: Only show copy that assumes PSP is non-stripe if PSP is truthy

bdsl:


NoelLH (approved):
I think that if we did the more general fix of handling `noPsps` much earlier we wouldn't really need this, but safe enough to do both

NoelLH:
@bdsl Shall we either close or merge this, depending on whether you agree https://github.com/thebiggive/donate-frontend/pull/2385 reduces the practical risk enough that we don't really need it?

bdsl:
@NoelLH yep, I think https://github.com/thebiggive/donate-frontend/pull/2385 is enough, thanks. I'll close.

PR #2385, opened 2026-06-18
Don't accept donation amounts in case there is no usable PSP for a ch…

bdsl:
…arity

Letting the donor get to the end and then giving an error message is a lot more confusing for us ot diagnose than blocking the donation from the start in case we don't have the PSP details set up for the charity.

Applies to saving us time dealing with incomplete test data as much as it should save donors time trying to donate to potential campaigns in prod that are not fully set up.

image

Copy is mine, we might want to try and make it a bit friendly if we can while still keeping it concise and clear.

### DIFF
--- a/src/app/donation-start/donation-start-form/donation-start-form.component.html
+++ b/src/app/donation-start/donation-start-form/donation-start-form.component.html
@@ -38,6 +38,12 @@
             />
           </biggive-text-input>
         </div>
+        @if (noPsps) {
+          <p class="error">
+            Sorry, there is a technical issue with this campaign. We cannot take donations right now. Please try again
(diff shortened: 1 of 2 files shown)


bdsl:
also wondering if the copy should change to clarify that it's about the charity campaign not the metacampaign. But we might not want to suggest the donor should donate to a different charity, when they can wait for (and/or ask for) the one they wanted to donate to to fix the PSP config.

NoelLH (approved):
LGTM. IMO the copy's a sensible level of ambiguous and this would rarely happen for long in Production.

NoelLH:
@bdsl shall we merge?

PR #2386, opened 2026-06-23
BG2-3324 CSP: Allow frame-src and form-action that Ryft 3DS pages require

NoelLH:


bdsl:
I know this is draft, but LGTM. I think we could only allow any https host on the donation start page, where the Ryft 3DS iframe would appear, and still keep the old restrictions on other pages, but it's probably not that useful since the donation start page is likely to be a main target for potential attacks anyway.

NoelLH:
> I know this is draft, but LGTM. I think we could only allow any https host on the donation start page, where the Ryft 3DS iframe would appear, and still keep the old restrictions on other pages, but it's probably not that useful since the donation start page is likely to be a main target for potential attacks anyway.

Right, and adds complexity and potential for mistakes if we start making these protections conditional. I'm also not sure how it would work in theory to lock down one page you end up on depending on the page you started on, seems like the app kind of has to agree across the board for any later navigation to work reliably.

bdsl (approved):

PR #2388, opened 2026-06-26
DON-1196 Remove old campaign compatibility error printing

NoelLH:


bdsl (approved):
LGTM - as noted on matchbot ticket the compatibility is a historical requirement, not a current one. Looks fine to merge this and the matchbot change in either order.

PR #2394, opened 2026-06-30
DON-1205: Update big give components to for filter-grid with location…

bdsl:
… search button


NoelLH (commented):
Needs the lockfile updated too?

src/app/explore/explore.component.html
```diff
@@ -63,13 +63,6 @@
         @if (this.clientCountryCode === "GB" || !this.clientCountryCode) {
           <!-- We only support searching by geolocation coordinates within the UK.
         Dev environments don't have this header so we also need to support it there. -->
-          <biggive-button
-            (doButtonClick)="searchByLocation()"
-            [disabled]="fetchingLocation"
-            [label]="fetchingLocation ? 'Fetching location info....' : 'Set location from device'"
-            size="medium"
-            colour-scheme="primary"
-          ></biggive-button>
```
bdsl, line 72: not needed now that we use the button in the campaign-filter-grid-component for this instead.

NoelLH (approved):

PR #2397, opened 2026-07-02
So 78 port filter grid

bdsl:
Porting the biggive-campaign-filter grid from Stencil to Angular, so that it can exist in this repo for a fast development cycle as we work on location based search, and because we don't anticipate it being used outside of Angular.

image


src/app/explore/campaign-card-filter-grid/campaign-card-filter-grid.component.html
```diff
@@ -0,0 +1,148 @@
...
+          [prompt]="null"
+          select-style="underlined"
+          [placeholder]="sortByPlaceholderText"
+          [selectedLabel]="selectedSortByOption"
+          [selectedValue]="getSelectedValue() || null"
+          [selectionChanged]="sortBySelectionChanged"
+          id="sort-by"
+        >
+          <!--
+        commented all nodes inside biggive-form-field-select out because there doesn't seem to be anything called
```
shaundbs: If no longer being used then these can be removed?

bdsl:
For ref the component here is a port from the one at https://github.com/thebiggive/components/tree/main/src/components/biggive-campaign-card-filter-grid . Once this is merged that can be deleted.

src/app/explore/campaign-card-filter-grid/campaign-card-filter-grid.component.html
bdsl: Yes I think so, will do. I don't think they were ever used. Wanted to leave it in at first just in case it had some purpose that I couldn't see, but since I now see that it seems to work fine without them I will delete in a separate commit.

shaundbs: LGTM in that case, feel free to merge after the deletion commit - thanks!

bdsl: Done in 11194bf3

PR #2398, opened 2026-07-06
MAT-488: Request a short donation reservation when creating donation and then extend it every minute if possible.

bdsl:
Allows matchbot to release the funds if we don't extend the donation every five minutes.

image


NoelLH (dismissed):
Apart from lint error, LGTM

PR #2406, opened 2026-07-09
Reduce overly verbose amount validation messages

NoelLH:
Discussed a couple of weeks ago with Dom & Vikram, had uncommitted since


bdsl (approved):
LGTM

PR #2408, opened 2026-07-22
DON-1219 Fix incorrect `banner` property assumption breaking Regular Giving campaigns

NoelLH:
And very possibly also some old other campaigns

### DIFF
--- a/src/app/account/edit-home-address/edit-home-address.component.ts
+++ b/src/app/account/edit-home-address/edit-home-address.component.ts
@@ -13,7 +13,6 @@ import {
 } from '@biggive/components-angular';
 import { flags } from '../../featureFlags';
 import { FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
-import { MatCheckbox } from '@angular/material/checkbox';
 import { firstValueFrom } from 'rxjs';
 import { countryOptions } from '../../countries';
 
@@ -26,7 +25,6 @@ import { countryOptions } from '../../countries';
     BiggivePageSection,
     BiggiveHeading,
     BiggiveButton,
-    MatCheckbox,
     ReactiveFormsModule,
     BiggiveFormFieldSelect,
     BiggiveTextInput,
--- a/src/app/campaign-details/campaign-details.component.html
+++ b/src/app/campaign-details/campaign-details.component.html
@@ -38,17 +38,19 @@
     <div class="b-primary-column">
       <div class="banner-container">
         <!-- 30.6 is approx. 352/1152 as %, the typical app campaign & standalone campaign banner ratio hardcoded in Salesforce -->
-        <biggive-branded-image
-          space-below="4"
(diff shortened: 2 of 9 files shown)


src/app/account/edit-home-address/edit-home-address.component.ts
```diff
@@ -26,7 +25,6 @@ import { countryOptions } from '../../countries';
     BiggivePageSection,
     BiggiveHeading,
     BiggiveButton,
-    MatCheckbox,
```
NoelLH, line 29: Fixed a couple of unused imports that angular build warnings alerted me to while I was here

src/app/campaign-details/campaign-details.component.html
```diff
@@ -38,17 +38,19 @@
...
+            [imageUrl]="campaign.banner.uri | optimisedImage: 2000 | async"
+            [logoUrl]="campaign.charity.logoUri ? (campaign.charity.logoUri | optimisedImage: 240 | async) : null"
+            [imageAlt]="campaign.banner.alt_text"
+            [slug]="campaign.isMatched ? 'Match Funded' : null"
+            [charityName]="campaign.charity.name"
+            [charityLocation]="null"
+            [imageHeightPercent]="30.6"
+          >
+          </biggive-branded-image>
+        }
```
bdsl: shouldn't we show what we can here even if the banner is missing? Looks like there's still quite a bit of useful info, i.e. the charity name, "Match Funded"  message, charity logo?

If we're taking it out just temporarily to silence the alarms maybe we need a ticket or something to know to fix it? Or it might be quick enough to just put a placeholder for the uri and alt text or even use null-safe navigation if the component will accept null Url and alt text?

bdsl (commented):
Reviewed, added comment inline, otherwise all LGTM.

src/app/campaign-details/campaign-details.component.html
NoelLH: Hm, I assumed because of the component's name that the most appropriate thing was to skip it.

Looking more closely at the spec it also seems that e.g `imageAlt` is unconditionally required which seems odd if we are saying images in general are optional. What do you think we should do with that, pass a blank string?

bdsl: Looking at the component implementation it should be fine to use null-safe navigation for both the uri and the alt - if the uri is null or undefined then the alt won't be used anyway: https://github.com/thebiggive/components/blob/666e65e0881d844da094e151c1c4dd87633e3232/src/components/biggive-branded-image/biggive-branded-image.tsx#L65

bdsl (dismissed):
LGTM!

src/app/campaign-details/campaign-details.component.html
NoelLH: Right, made this closer to how `donation-start-container` handles its optimised `bannerUri` which may also be null, except still using the `OptimisedImagePipe` here, in 3e6b7478a29e0a44e6f4a621139b2673f068b9ce.

PR #2411, opened 2026-07-23
Use `polyfill-corejs3` for babel ES5 / tier 2 legacy builds

NoelLH:
`usage-global` sounds like it should reinstate the removed-from-babel option we were using until today. https://github.com/babel/babel-polyfills/blob/HEAD/docs/usage.md#method


bdsl (commented):
LGTM

PR #2413, opened 2026-07-28
BG2-3353 Use OpenStreetMap tiles as map base layer instead of DIY approach

NoelLH:
The layer reordering as it stands also changes the copyright order, which I wanted to do anyway.

Screenshot 2026-07-28 at 13 39 59

### DIFF
--- a/src/app/campaign-info/campaign-info.component.ts
+++ b/src/app/campaign-info/campaign-info.component.ts
@@ -13,7 +13,7 @@ import { isPlatformBrowser } from '@angular/common';
 import { PLATFORM_ID, AfterViewInit, ViewChild, ElementRef, OnDestroy } from '@angular/core';
 import { HttpClient } from '@angular/common/http';
 import { firstValueFrom } from 'rxjs';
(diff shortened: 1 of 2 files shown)


src/app/campaign-info/campaign-info.component.ts
```diff
@@ -145,46 +145,22 @@ export class CampaignInfoComponent implements OnInit, AfterViewInit, OnDestroy {
...
-    const englandRegionsData = await firstValueFrom(this.http.get('../../assets/map/englandRegions.geojson'));
-    new GeoJSON(englandRegionsData, { style: () => constantLayerStyle }).addTo(this.map);
-
     const matchedRegions: string[] = [];
 
     // Build a layer with just project-relevant locations and a list of their names
     const highlightAreas = await this.getHighlightedFeatures(regionCodes);
 
+    new TileLayer('https://tile.openstreetmap.org/{z}/{x}/{y}.png', {
+      maxZoom: 13,
```
NoelLH, line 154: Slightly arbitrary, from what I can remember of the early experiments I don't think we'll ever want this much detail. Can try a few campaigns and see.

PR #2415, opened 2026-07-30
DON-1195: Show campaign summary near top of regular giving signup form

bdsl:
This doesn't make clear that the summary is written by the charity not us, but I think we can trust our charities not to write anything harmful here.

Also wondering if we should make this page more like the ad-hoc donation campaign detail page (since this page is sharing the function of that one as well as showing the form) by making the charity name blue and perhaps removing the "you are supporting" text that may be redundant.

Also note that we still don't show the campaign name, since regular giving campaigns have all been to provide unrestricted funds to the charity so the campaign names didn't seem to be useful.

The campaign summary in this screenshot is "We can Save Matchbot 1"

image

### DIFF
--- a/src/app/app.routes.ts
+++ b/src/app/app.routes.ts
@@ -198,7 +198,10 @@ export const routes: (Route & {
     title: undefined, // set from inside component
     pathMatch: 'full',
     component: RegularGivingComponent,
-    canActivate: [requireLogin],
+    canActivate: [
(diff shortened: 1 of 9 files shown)


NoelLH (commented):
I think tidying up the 'you are supporting' bit, aligning with individual donation page updates, makes sense. Changes so far LGTM.

shaundbs (approved):

PR #2416, opened 2026-07-30
Don-1195 regular giving: Allow viewing form without logging in.

bdsl:
In the environments where this is enabled the donor can now progress through the form as far as the payment step, which will broken as stripe isn't loaded and they can't progress further. 

But I think this is fine to merge and we know that we now have to add elements into the form to allow and require the donor to login/signup before they reach the payment step.

image

### DIFF
--- a/src/app/regular-giving/regular-giving.component.ts
+++ b/src/app/regular-giving/regular-giving.component.ts
@@ -1,4 +1,4 @@
-import { AfterViewInit, Component, ElementRef, OnDestroy, OnInit, ViewChild, inject } from '@angular/core';
+import { AfterViewInit, Component, ElementRef, inject, OnDestroy, OnInit, ViewChild } from '@angular/core';
 import { ActivatedRoute, Router, RouterLink } from '@angular/router';
 import { Campaign, formattedCampaignSummary } from '../campaign.model';
 import {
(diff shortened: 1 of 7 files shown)


src/app/regular-giving/regular-giving.component.ts
```diff
@@ -650,7 +661,11 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy
   private validatePaymentInformationStep(): boolean {
     this.paymentInfoErrorMessage = undefined;
 
-    if (!this.stripePaymentMethodReady && !this.donorAccount.regularGivingPaymentMethod) {
+    if (!this.donorAccount) {
+      // likely this branch can never happen as we will check donor has an account before they see the payment information
+      // step
+      this.paymentInfoErrorMessage = 'Please login or create your donor account';
```
bdsl, line 667: @NoelLH In our discussion this morning think you were saying this validation should be one step earlier, (e.g. in the `validateAmountStep` function) to stop the donor filling in the gift aid information before they have an account to follow HMRC requirements. But I didn't get what requirements you mean - having an account seems like an internal concept to Big Give, we define what it means for someone to have an account with us so I don't know why it would matter to HMRC.

But maybe it doesn't matter and we prefer the process of them making sure they're logged in in a step we'll add before gift aid anyway.

Image

NoelLH: I was thinking the declaration may be specific to who is doing it. As it doesn't contain their name maybe not, providing we can ensure it is never shown to organisations.

NoelLH: We didn't really talk about whether we _should_ move it, if we could.

I don't think the complexity is enough higher/lower than auth to justify a change personally.

bdsl: Thanks - I think it's a change either way, if we add login as a new step in the stepper, but probably does make most sense to put that as step two, so when I do I'll add a validation rule that you can't progress past there without being logged in.

NoelLH (approved):

PR #2418, opened 2026-07-31
DON-1195: Send registration email from regular giving signup form.

bdsl:
Code mostly copied from https://github.com/thebiggive/donate-frontend/blob/c79d3bd2260a3f23d9cd0b5fa2acc06868a1202c/src/app/account/register/register.component.ts and https://github.com/thebiggive/donate-frontend/blob/8a4308633fc744c5ea788bdf33d5ae47c8faefcf/src/app/account/register/register.component.html . 

Probably not worth trying to share the code with those files since its copied in several small pieces.

### DIFF
--- a/src/app/regular-giving/regular-giving.component.ts
+++ b/src/app/regular-giving/regular-giving.component.ts
@@ -207,7 +207,7 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy 
   private friendlyCaptchaWidget!: WidgetInstance;
   private platformId = inject(PLATFORM_ID);
 
-  protected processing = false;
+  protected processingTempPasswordRequest = false;
 
   ngOnInit() {
     this.donor = this.route.snapshot.data['donor'];
@@ -612,6 +612,7 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy 
         return;
       }
 
+      this.processingTempPasswordRequest = true;
       try {
         // @todo-DON-1195: CHeck the friendlyCaptchaSolution is provided, don't just assume its truthy - show the donor an error message if its missing e.g. because they clicked send email too quickly.
         // @todo-DON-1195: Request an email with different copy from the idenity service that's specific to the fact that they're in the process of setting up a regular giving mandate, and refers to "temporary password"
@@ -629,7 +630,7 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy 
       } finally {
         this.friendlyCaptchaWidget?.reset();
         await this.friendlyCaptchaWidget?.start();
-        this.processing = false;
+        this.processingTempPasswordRequest = false;
       }
     }
 
--- a/src/app/regular-giving/regular-giving.component.ts
+++ b/src/app/regular-giving/regular-giving.component.ts
@@ -620,6 +620,7 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy 
         /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
       } catch (error: any) {
         this.extractErrorMessage(error);
+        return;
       } finally {
         this.friendlyCaptchaWidget?.reset();
         await this.friendlyCaptchaWidget?.start();
--- a/src/app/regular-giving/regular-giving.component.ts
+++ b/src/app/regular-giving/regular-giving.component.ts
@@ -306,7 +306,13 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy 
       doneCallback: (solution) => {
         this.friendlyCaptchaSolution = solution;
       },
-      errorCallback: () => {},
+      errorCallback: (error: unknown) => {
+        // not sure if this ever really happens, but will show an error message in case it does.
+        console.error(error);
+        this.toast.showError(
+          'Sorry, something went wrong with the CAPTCHA - please try again or contact Big Give support.',
+        );
+      },
     });
     await this.friendlyCaptchaWidget.start();
   }
--- a/src/app/regular-giving/regular-giving.component.ts
+++ b/src/app/regular-giving/regular-giving.component.ts
@@ -1,4 +1,4 @@
-import { AfterViewInit, Component, ElementRef, inject, OnDestroy, OnInit, ViewChild } from '@angular/core';
+import { AfterViewInit, Component, ElementRef, inject, OnDestroy, OnInit, PLATFORM_ID, ViewChild } from '@angular/core';
 import { ActivatedRoute, Router, RouterLink } from '@angular/router';
 import { Campaign, formattedCampaignSummary } from '../campaign.model';
 import {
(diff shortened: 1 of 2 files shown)


src/app/regular-giving/regular-giving.component.ts
```diff
@@ -572,6 +604,29 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy
...
+      } catch (error: any) {
+        this.extractErrorMessage(error);
+      } finally {
+        this.friendlyCaptchaWidget?.reset();
+        await this.friendlyCaptchaWidget?.start();
+        this.processing = false;
+      }
+    }
+
     this.stepper.selected = this.stepper.steps.get(stepIndex);
```
shaundbs, line 643: Here will the donor not be moved to the next step despite the registration request failing, wouldn't we want to stop that flow and display an error?

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -572,6 +604,29 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy
...
+      // this is the Send email button so let's send an email.
+
+      try {
+        // @todo-DON-1195: CHeck the friendlyCaptchaSolution is provided, don't just assume its truthy - show the donor an error message if its missing e.g. because they clicked send email too quickly.
+        // @todo-DON-1195: Request an email with different copy from the idenity service that's specific to the fact that they're in the process of setting up a regular giving mandate, and refers to "temporary password"
+        // @todo-DON-1195: instead of a verification code (once we've adjust the login function to accept a verification code typed instead of a password).
+        // @todo-DON-1195: work out how/where we're going to be collecting the donor's first and last name, which we should only need to ask for if its a new account. May be a challenge to the idea of using the same input box to accept either
+        // @todo-DON-1195: a password for an existing account or a verification code aka temporary password for a new account.
+        await this.identityService.requestEmailAuthToken(this.mandateForm.controls.emailAddress.value!, {
+          captcha_code: this.friendlyCaptchaSolution!,
```
shaundbs: Shouldn't there be a guard here before calling the identity service since it could still be undefined? 

e.g. 
`if (!this.friendlyCaptchaSolution) {
    this.emailErrorMessage = 'Please complete the CAPTCHA before continuing.';
    return;
}`

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -193,6 +199,15 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy
...
   protected formattedCampaignSummary!: string;
+  private sanitizer = inject(DomSanitizer);
+
+  @ViewChild('frccaptcha', { static: false })
+  protected friendlyCaptcha!: ElementRef<HTMLElement>;
+  private friendlyCaptchaSolution: string | undefined;
+  private friendlyCaptchaWidget!: WidgetInstance;
+  private platformId = inject(PLATFORM_ID);
+
+  protected processing = false;
```
shaundbs: Is this var ever set to true? Without that can't users click repeatedly and send multiple registration emails?

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -277,6 +292,23 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy
...
+    if (environment.environmentId === 'regression') {
+      this.friendlyCaptchaSolution = 'dummy-captcha-code';
+      return;
+    }
+
+    this.friendlyCaptchaWidget = new WidgetInstance(this.friendlyCaptcha.nativeElement, {
+      doneCallback: (solution) => {
+        this.friendlyCaptchaSolution = solution;
+      },
+      errorCallback: () => {},
```
shaundbs: Not sure that we want CAPTCHA errors to be silently ignored, wouldn't we want something visible?

shaundbs (changes requested):

src/app/regular-giving/regular-giving.component.ts
bdsl: hmm not it's not. It's also not currently read anywhere. It's there because I pasted code from `register.component.ts` , and its used in the associated `register.component.html` to show a spinner instead of a button while awaiting the response from the server. I'm not 100 sure if we actually want to do the same here, or the stepper design maybe means we should updated more optimistically. I don't think there is anything currently that would stop users repeatedly clicking past step 1 to send multiple registration emails.

Do you have a view on if we should delay progress to step two until the promise of `this.identityService.requestEmailAuthToken` resolves?

src/app/regular-giving/regular-giving.component.ts
bdsl: I'm not sure - I think the captcha widget itself can display errors without our help, and in terms of our application I'm not sure we need to treat an error any differently to the case where the user was too quick and the `doneCallback` just hasn't been triggered yet. We could display a toast error message in the errorCallback though. Not sure how likely any error is.

This code is copied verbatim from register.component.ts where it's been for several months afaik without complaint. What would you put in the errorCallback?

src/app/regular-giving/regular-giving.component.ts
bdsl: @shaundbs yes there should be, hence the todo message on line 611. I was trying to get the basic case working before adding in all the guards etc. But I can add that now.

src/app/regular-giving/regular-giving.component.ts
bdsl: Yes good point, fixed in 7d186c1e

src/app/regular-giving/regular-giving.component.ts
bdsl: Done in b18c2087

src/app/regular-giving/regular-giving.component.ts
bdsl: Fixed in 9a789e59

src/app/regular-giving/regular-giving.component.ts
shaundbs: Sure, fair point - I’m mainly concerned that an empty errorCallback makes a CAPTCHA failure invisible to us and potentially leaves the user unable to continue without any explanation.

I don’t think we need anything complicated here, maybe just a generic message like 'The security check could not be completed. Please refresh the page and try again.'  would be enough. If the widget already displays its own error state, then I’m happy for us not to duplicate it but it would be good to confirm that behaviour rather than silently swallowing the callback.

bdsl: I've added both a visible error display and an error log in the errorCallback in commit ~cc846a7a~ 759a2883 .   

I looked at https://developer.friendlycaptcha.com/docs/v2/guides/upgrading-from-v1/javascript-api#2-change-how-you-create-widgets - it's a bit hard to see what we should be doing because it looks like we're using the JS v1 api and the docs are al about the v2 api now apart from this page showing how to upgrade. It doesn't really give any details of if/when we should expect the error call back to run. I guess most obviously when the friendlyCaptcha servers are down.

shaundbs: LGTM

shaundbs (approved):

PR #2424, opened 2026-08-05
DON-1195: Request regular-giving specific auth email during regular g…

bdsl:
…iving registration

### DIFF
--- a/src/app/regular-giving/regular-giving.component.html
+++ b/src/app/regular-giving/regular-giving.component.html
@@ -216,7 +216,7 @@
                 />
               </biggive-text-input>
               <biggive-text-input spaceBelow="4">
-                <label slot="label" for="password">Set new password *</label>
+                <label slot="label" for="password">New password *</label>
                 <div slot="input" style="display: flex">
                   <div class="visually-hidden" aria-live="polite">
                     {{ showPassword() ? "Your password is shown" : "Your password is hidden" }}
--- a/src/app/regular-giving/regular-giving.component.html
+++ b/src/app/regular-giving/regular-giving.component.html
@@ -200,10 +200,10 @@
               </button>
             </mat-step>
             <mat-step label="About You">
-                <biggive-text-input>
-                  <label slot="label" for="firstName">First Name *</label>
-                  <input slot="input" formControlName="firstName" id="firstName" autocomplete="given-name" matInput />
-                </biggive-text-input>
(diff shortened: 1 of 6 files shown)


src/app/regular-giving/regular-giving.component.html
```diff
@@ -71,22 +71,20 @@
                   </label>
                   <input maxlength="10" formControlName="donationAmount" id="donationAmount" matInput slot="input" />
                 </biggive-text-input>
-                @if (enableCondensedRegularGivingSignup) {
```
bdsl, line 74: as noted in commit message I'm removing the feature flag because the flag-off version was clearly broken some time ago, it doesn't seem worthwhile to fix it, and so having the flag at all is just misleading.

If we wanted to have the flag I probably should have made a second copy of the whole component directory.

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -129,6 +140,11 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy
     homePostcode: new FormControl<string | null>(null),
     unmatched: new FormControl(false), // If ticked, indicates that the donor is willing to donate without match funding.
     aged18OrOver: new FormControl(over18DefaultValue, [Validators.requiredTrue]),
+    password: new FormControl('', [
+      Validators.required,
+      Validators.minLength(6), // temp password is a number between 0 and 1,000,000.
```
NoelLH: Rule and comment here seem to disagree, or don't explain the padding clearly.

src/app/regular-giving/regular-giving.component.ts
```diff
@@ -858,4 +885,94 @@ export class RegularGivingComponent implements OnInit, AfterViewInit, OnDestroy
...
+
+    const emailAddress = this.mandateForm.controls.emailAddress.value;
+    if (!emailAddress) {
+      this.toast.showError('Email address missing - cannot continue');
+      return;
+    }
+
+    const password = this.mandateForm.controls.password.value;
+    if (!password) {
+      this.toast.showError('password address missing - cannot continue');
```
NoelLH:
```suggestion
      this.toast.showError('password missing - cannot continue');
```

NoelLH (commented):
Couple of things inline, and there's currently a lint error. Otherwise LGTM.

PR #2427, opened 2026-08-12
DON-1026 Add Matomo tracking for Regular Giving sign up

NoelLH:
Copies from one-time donate journey guided by Copilot / Gemini session.

The dummy Donation feels slightly messy but I don't have an immediate better idea.


bdsl (commented):
LGTM - will merge after I've un-reverted the revert at https://github.com/thebiggive/donate-frontend/pull/2428 . 

Although it's true that we don't directly make a donation within the mandate form, we do create three donations immediately as the mandate is created, so in principle we could send date from one or more of them to matamo when the donor hits the thanks page.

PR #2440, opened 2026-08-20
BG2-3354: WIP: Display map of UK when UK location filter is enabled

bdsl:
Before enabling this in prod we need to adjust the layout so it goes to the side of the campaign cards on desktop, and more importantly overlay useful info on the map, like how many campaigns are in each region of the UK.

I think it's not part of this ticket but we may also want to enable clicking a region to filter results to just that region and simulteanously zoom in the map before releasing.

image


src/app/explore/campaign-card-filter-grid/campaign-card-filter-grid.component.scss
```diff
@@ -211,3 +211,30 @@ fa-icon {
...
+}
+
+.campaign-map {
+  height: 500px;
+  margin-right: 23px;
+  margin-left: 9px;
+  @media #{abstract.$breakpoint-lg} {
+    height: 800px;
+  }
+  //width: 100%;
```
NoelLH:
```suggestion
```

NoelLH (approved):

src/app/explore/campaign-card-filter-grid/campaign-card-filter-grid.component.scss
```diff
@@ -211,3 +211,30 @@ fa-icon {
...
+.grid-map-container {
+  display: flex;
+  gap: 20px;
+  flex-direction: column-reverse;
+  @media #{abstract.$breakpoint-lg} {
+    flex-direction: row;
+  }
+}
+
+.grid-map-container div {
```
shaundbs: I think this wants to target only the immediate grid/map children rather than every descendant div? As it is this will also apply flex-basis/flex-grow to divs inside the campaign grid and Leaflet map.

Maybe something like:
```
.grid-map-container {
    display: flex;
    gap: 20px;
    flex-direction: column-reverse;

    @media #{abstract.$breakpoint-lg} {
        flex-direction: row;
    }

    > .campaign-grid,
    > .campaign-map {
        @media #{abstract.$breakpoint-lg} {
            flex-basis: 0;
            flex-grow: 1;
        }
    }
}
```

bdsl: Thanks. I think in practice it can't target the individual cards because those are in a different Angular component and the CSS is scoped. But I agree it should be just the direct children even if we end up adding indirect child divs inside this component. 

I've used the child combinator in 63e40974 , I think that should be enough.

PR #2442, opened 2026-08-26
Bg2 3359 - Metacampaigns bugfix

shaundbs:
Adds cdn-wp.thebiggive.org.uk to the img-src CSP allowlist so these banners can load and display correctly.


DomTBG (approved):
LGTM

